PR #1, opened 2022-12-23
Merge Importers with Filesystem

ericdacal:
- feat: Implementation of an abstract base class Importer

### DIFF
--- a/Source/Engine.vcxproj
+++ b/Source/Engine.vcxproj
@@ -334,6 +334,7 @@
     <ClInclude Include="debugdraw.h" />
     <ClInclude Include="debug_draw.hpp" />
     <ClInclude Include="EngineLog.h" />
+    <ClInclude Include="FileSystem\Importers\Importer.h" />
     <ClInclude Include="FileSystem\ModuleFileSystem.h" />
     <ClInclude Include="Globals.h" />
     <ClInclude Include="Modules\Module.h" />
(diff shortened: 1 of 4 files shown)


gerard-queralt:
I have a few questions, if you could tell me why you did it the way you did:

- Why do ModuleFileSystem::Load and Save return an unsigned int that is always 0?
- What is the `uint64_t` returned by Importer::Save?
- Why do the methods in Importer use output parameters instead of return types?

And sorry to be picky, but a couple things:

- The parameter `file_path` in the methods of ModuleFileSystem should be in camelCase
- If you could, organize your classes with filters in Visual Studio's solution explorer following the way you organized it in the project structure (so a FileSystem filter with an Importers filter inside)

gerard-queralt:
Also, I noticed you included a file named FileSystem.h that is empty

ericdacal:
Corrections:
- I refactor the filePath variable to CamelCase.
- I remove the unnused FileSystem.h
Answers:
- Importer::Save have to return IUD identificator of the resource
- FileSystem is only a basic version now, we have to improve then before. 
- Return 0 if the function doesn't have error and 1 if there is a error. But we need to create a catch for the error.
- That is base in the sliders of the teacher.

gerard-queralt (approved):
Great, thank you very much for the answers and for your work!

AuxPort66 (approved):

PR #4, opened 2022-12-26
Merge refactored GameObject Class into SceneManagement

sncarbo:
Changes in constructor, destructor, Update and CreateComponent

### DIFF
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -6,21 +6,34 @@
 
 #include <assert.h>
 
-GameObject::GameObject()
+GameObject::GameObject(const char* name, GameObject* parent)
 {
+	this->name = name;
+	this->parent = parent;
+
+	if (this->parent != nullptr)
+		this->parent->children.push_back(this);
+
 	Component* transform = CreateComponent(ComponentType::TRANSFORM);
 	components.push_back(transform);
+
+	active = true;
 }
 
 GameObject::~GameObject()
 {
+	components.clear();
+	children.clear();
 }
 
 void GameObject::Update()
 {
-	for (int i = 0; i < components.size(); i++)
+	if (active)
 	{
-		components[i]->Update();
+		for (int i = 0; i < components.size(); i++)
+		{
+			components[i]->Update();
+		}
 	}
 }
 
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -6,26 +6,21 @@
 
 #include <assert.h>
 
-GameObject::GameObject(const char* name, GameObject* parent)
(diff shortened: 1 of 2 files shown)


markmeliaa (changes requested):
Some comments about your work,
- First of all, you refactored the CreateComponent function, good change, well done, but you forgot to add a break at the end of each switch case, and thus when the code will enter inside a case, instead of exiting directly from the switch it will check all the other cases before exiting (before that there was a return in each case block and that served as exiting the switch case, but now, as there are no returns, a break should be placed instead).

- Also, in that same function, you are pushing back the new created component to the GameObject components, but then, when creating a transform in the GameObject constructor, you are pushing it back again; I'd say to delete the push back inside the CreateComponent function because that function already returns the newly created component, and we can push it back outside the function (just like what was done with the transform).

Now, regarding the new GameObject constructor you added:
- The GameObject parent passed as a parameter should be const, as it is not modified inside the body of the constructor.
- name and parent should be initialized the following way: 
a doing it this way instead of this->name = name and this->parent = parent is better because you do not assign first a default value (0) to them and then change its value, instead you will be assigning them a value directly.

- parent would never be nullptr, the point of having parents and childrens is to have a tree structure that connects every GameObject in the current Scene, and if any GameObject has an empty parent it means it would not be connected to anything and thus will be not included in the scene hierarchy (the parent should either be another GameObject or, if itself is a parent GameObject, the root GameObject of the Scene will be its parent).

- also, the active = true is not needed because active is set to true by default in the header file.

Finally, I've also seen that you added in the destructor the deletion of both the vector components and the vector children, but I think that just calling clear on the vector is not enough to deallocate its memory and we could have memory leaks, check out this article: [https://stackoverflow.com/questions/10464992/c-delete-vector-objects-free-memory](url).

That's all, I've also seen that you updated the Update of the GameObject to only update its components when the GameObject is active, good job, well seen!

sncarbo:
I made some of the changes you suggested, thank you for noticing them. However, I consider that two of the changes are not necessary:

- In the GameObject Constructor, the "parent" parameter cannot be const because it is not possible to assign a const element to a non-const element.
- In my opinion, the "push_back" of the new component should be done inside the "CreateComponent" method because, in order to guarantee a low coupling, it is necessary that the functionalities inherent to a class are done in the same class. Inside the GameObject Constructor the "CreateComponent" method is not called, but the new pointer is created directly with "new ComponentTransform()" and then the "push_back" is done, so it is not added twice.

Anyway, thank you very much for your comments!

markmeliaa (approved):

markmeliaa:
Alright then, if we push_back the component from inside its function, should the function still be returning Component*? or should we change it to void? I do not see the point (for now) to return the component we created if we already assigned it to its parent.

sncarbo:
I think we will need that pointer in the future in order to access to or modify attributes through methods in the different modules.

totisama (approved):

PR #8, opened 2022-12-29
Merge branch 'scenemanagement' into 'link-hierarchy'

sncarbo:
Finally, the GameObject Constructor has been implemented in this way ("GameObject(const char* name, GameObject* parent = nullptr)") to be able to encapsulate two ways of calling it.

In case of needing a root GameObject this is the correct way to call the constructor: GameObject("root-name").

Otherwise, it will be called like this: GameObject("node-name", parent).

### DIFF
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -6,6 +6,12 @@
 
 #include <assert.h>
 
+GameObject::GameObject(const char* name) : name(name)
+{
+	Component* transform = new ComponentTransform(true, this);
+	components.push_back(transform);
+}
+
 GameObject::GameObject(const char* name, GameObject* parent) : name(name), parent(parent)
 {
 	if(this->parent != nullptr)
@@ -46,13 +52,13 @@ void GameObject::SetParent(GameObject* parent)
 
 void GameObject::AddChild(GameObject* child)
 {
-	if (!IsADescendant(child))
+	if (!IsAChild(child))
 		children.push_back(child);
 }
 
 void GameObject::RemoveChild(GameObject* child)
 {
-	if (IsADescendant(child))
+	if (IsAChild(child))
 	{
 		bool loop = true;
 
@@ -103,15 +109,15 @@ Component* GameObject::CreateComponent(ComponentType type)
 	return newComponent;
 }
 
-bool GameObject::IsADescendant(const GameObject* descendant)
+bool GameObject::IsAChild(const GameObject* child)
 {
-	bool isChild = false;
+	bool isAChild = false;
 
-	for (GameObject* child : children)
+	for (GameObject* gameObject : children)
 	{
-		if (child == descendant || child->IsADescendant(descendant))
-			isChild = true;
+		if (gameObject == child)
+			isAChild = true;
 	}
 
-	return isChild;
+	return isAChild;
 }
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -59,14 +59,12 @@ void GameObject::RemoveChild(GameObject* child)
 {
 	if (IsAChild(child))
 	{
-		bool loop = true;
-
-		for (std::vector<GameObject*>::const_iterator it = children.begin(); it != children.end() && loop; ++it)
+		for (std::vector<GameObject*>::const_iterator it = children.begin(); it != children.end(); ++it)
 		{
 			if (*it == child)
 			{
 				children.erase(it);
-				loop = false;
+				return;
 			}
 		}
 	}
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -107,9 +107,9 @@ bool GameObject::IsADescendant(const GameObject* descendant)
 {
 	bool isChild = false;
 
-	for (std::vector<GameObject*>::const_iterator it = children.begin(); it != children.end() && !isChild; ++it)
+	for (GameObject* child : children)
 	{
-		if (*it == descendant || (*it)->IsADescendant(descendant))
+		if (child == descendant || child->IsADescendant(descendant))
 			isChild = true;
 	}
 
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -42,12 +42,11 @@ void GameObject::Update()
 void GameObject::SetParent(GameObject* parent)
 {
 	if (this->parent != nullptr)
+	{
 		this->parent->RemoveChild(this);
-
-	this->parent = parent;
-
-	if (this->parent != nullptr)
+		this->parent = parent;
 		this->parent->AddChild(this);
+	}	
 }
 
 void GameObject::AddChild(GameObject* child)
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -8,7 +8,8 @@
 
 GameObject::GameObject(const char* name, GameObject* parent) : name(name), parent(parent)
 {
-	this->parent->children.push_back(this);
+	if(this->parent != nullptr)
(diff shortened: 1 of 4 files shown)


markmeliaa (changes requested):
Some changes to GameObject.cpp:
- GameObject constructor -> avoid using constructors with default types, prefer in-class initializers (instead of GameObject(const char* name, GameObject* parent = nullptr) as a constructor for the root (parent is already set to nullptr by default in GameObject.h), use GameObject(const char* name) for when no parent is needed). 
It's nice that you have polymorphism in mind, but with constructors, C++ standard guidelines are against using default parameters: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c48-prefer-in-class-initializers-to-member-initializers-in-constructors-for-constant-initializers.
- Also, when creating this new constructor GameObject(const char* name), know that, in C++, constructors that have one and just one non-default argument should be 'explicit' to avoid that the compiler uses them as typecasters (note that this would have also applied for GameObject(const char* name, GameObject* parent = nullptr) because it alo has one and just one non-default argument), more in-depth explanation here https://www.geeksforgeeks.org/use-of-explicit-keyword-in-cpp/, or here https://www.sjbrown.co.uk/posts/always-use-explicit/.
- Nice change of for statements to range-based for statements in the Update method.
- SetParent function -> I am guessing that this function is for when a GameObject is moved around in the hierarchy (if not please correct me and ignore the rest). In that case, the first if condition will always be true, as any of the GameObjects that can be moved around (the root should not be moved around) always have a parent, as it either would be another GameObject or the root itself, so when moving a GameObject around it should always be removed from its parent.
Same with the second if condition, when setting the GameObject as a child, its new parent would never be nullptr, because only the root has a nullptr parent, and, as stated, we will never move the root around.
- RemoveChild function -> this function is correct, but I will make some changes later to increase its readability. Know that, instead of creating a boolean for a for loop and change its value when the required object is met, you can always use a break when the required object is found and avoid creating an unnecessary boolean: https://stackoverflow.com/questions/3922599/is-it-a-bad-practice-to-use-break-in-a-for-loop.
- IsADescendant function -> in the if inside the loop the first condition is okay, the second condition (child->IsADescendant(descendant)) is also checking if any children of the parent have the 'descendant' as children isn't it? Why did you include that? It will lead to bugs when moving a GameObjects from the child of a parent to the parent itself (this is confusing, please if you do not understand, ask me through WhatsApp and I will explain this better).

Some changes to GameObject.h:
- Include the new GameObject constructor as stated in the prior section.

Some changes to ModuleScene.h:
- Why did you change the forward declaration of the GameObject class and instead included GameObject.h? As I am seeing, there are only pointers in the header file and GameObject.h can be included in ModuleScene.cpp instead and we will avoid including header files in other header files.

That is all, any questions please feel free to reach me through WhatsApp or Discord, thanks for your work! :)

sncarbo:
Regarding your instructions:

- I have created the two GameObject Constructors following the rule "C.48: Prefer in-class initializers to member initializers in constructors for constant initializers" and use "explicit" in the constructor with one parameter, as you have indicated.
- In the SetParent() method, as you have pointed out, the root should not be moved around, so an if-statement that guarantees it is necessary.
- It is desirable to minimize the use of "break" within loops, as indicated by the Statement Rules of the C++ Standard Guidelines (https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es77-minimize-the-use-of-break-and-continue-in-loops). Currently, the body of the loop is trivial, however, in the future it may not be the case, and it is easy to overlook it. That is why I have decided not to use "break" in the RemoveChild() method.
- IsADescendant() method was implemented because I considered that it was necessary for GameObjects to not to be a descendant of itself. I just checked Unity and I think it might be possible for a GameObject to be a descendant of itself, so I've changed the IsADescendant() method to IsAChild(), with the appropriate implementation.
- I changed the forward declaration of GameObject to an "include" because in line number 14 ("root = new GameObject("Root");") incomplete type is not allowed.

Thank you for your feedback!

markmeliaa:
Thank you for your changes,
- Now, the condition if(this->parent != nullptr) in the GameObject constructor is not necessary because, as stated, every GameObject that uses this constructor will always have parent, and the root will use a different constructor.
- In the SetParent method I do not understand why you say that this method prevents the root by being moved around, if SetParent is called in the root and is dragged to another GameObject, neither RemoveChild nor AddChild will be called, that is true, but the parent of the root will change regardless and the structure will be broken. To avoid this, what I thought is making the root not draggable (I myself will do this later), so it will be static in the hierarchy and thus both if statements of the SetParent method will have no meaning.
- About the break, thanks for pointing out the C++ standard guidelines about this, I did not check them before writting the last changes, but I strongly suggest about a break instead of a boolean is better, not only because it will be less verbose and will increase the readability of the function, but also because, as you said, now this loop is trivial, and as it is inside a function and a funcion ideally only does one task, I doubt that we will add later such funcionality to the loop to make it not trivial (and in that case, the break could always be changed to a variable, but for now I find it very verbose.
- And for the last point, as you said the "incomplete type not allowed" triggers in the ModuleScene.cpp, not in ModuleScene.h, so a forward declaration should be the correct thing to do in the .h and then in the .cpp the #include "GameObject/GameObject.h" will avoid the error if I am not mistaken.

markmeliaa:
Another thing about the commits, when making a change to a script that improves performance, you should write perf: instead of fix: at the beginning of the commit (for example, in your second commit in this branch), and when renaming a function (like in your last commit), refactor: should be written instead of fix:. Only write fix: in a commit when fixing a bug existent for the user, not a fix for a build script, as stated in the programming rules (https://github.com/Pre-SuperAwesomeEngine/Engine/wiki/Programming-rules#github).

sncarbo:
Regarding your last feedback,

- I consider that checking that the parent pointer is not nullptr in the GameObject Constructor doesn't hurt hugely in terms of performance and prevents errors in case the Constructor is used to create a root pointer.
- Thank you for your recommendation about the "break", but I think that it is not possible to know if the loop will be modified in the future, so I think that said break is not necessary. I believe that the method is readable as is.
- I just modified the SetParent() method so that in no case can root be moved around.
- Regarding your comment about the "include" in the .cpp, I refer you to the Source Files rules in the C++ Standard Guidelines (SF.5: A .cpp file must include the header file(s) that defines its interface, https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#sf5-a-cpp-file-must-include-the-header-files-that-defines-its-interface). I understand that sometimes it is strictly necessary because the header could contain a lot of "includes", but in this context I understand that it is not.

Thank you again!

markmeliaa:
Okay, I accept your changes.

For the last point, I think that we are in different pages. What I am saying is that forward declaring should be used in "ModuleScene.h", and instead of #include "GameObject/GameObject.h", use a forward declaration of the GameObject class, so, as Carlos told us when revising the first assignment, we avoid including header files in other header files. 
Now, to solve up the error "incomplete type is not allowed" that the ModuleScene.cpp will trigger, the #include "GameObject/GameObject.h" will be added to the ModuleScene.cpp.
What you refered me only says that every .cpp file must include its .h interface file, and that is correct, but it is not what I am trying to say to you.

That is all, ty! :)

markmeliaa:
Wait, in the RemoveChild(), as indicated by the Statement Rules of the C++ Standard Guidelines (https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es77-minimize-the-use-of-break-and-continue-in-loops) why not, instead of defining a boolean, use a return when the desired child is met? The Standard Guidelines state that it is correct to exit the loop inside a function using a return statement (in this case just a return because it is a void function).

sncarbo:
I personally don't like using return except at the end of a method, because it breaks the sequentiality of the method, but I understand that it is possible to include it. I have just updated the code with the respective changes.

markmeliaa (approved):

diegusmejia117 (approved):
Nice code Santi, i am personally learning  a lot from your implementation.

sncarbo:
@diegusmejia117 Thank you very much Diego. I really appreciate your comment.

PR #25, opened 2023-01-03
Merge GameObjects-UniqueID into scenemanagement

markmeliaa:


markmeliaa:
UniqueID.h I think that must be included in the GameObject.h file because if not, I can't use the UID declaration in the header. If you have any other solution that avoids including a header file in another header file lmk, if not, we have to include it.

totisama (approved):

sncarbo (commented):
I think in this case it is unavoidable to write the include in the header.

By the way, I think you should remove #include "../../FileSystem/UniqueID.h" from GameObject.cpp then.

markmeliaa:
My bad, well seen, thanks!

sncarbo (approved):

totisama (approved):

PR #27, opened 2023-01-04
feat: Add a new function to start importation in Importer base class

ericdacal:
Add new base function to start importion in Importer class


gerard-queralt (changes requested):
I think `Start` should also have an output parameter `U* ourMaterial`, just like `Import`.

ericdacal:
I add the output parameter and change the name for Resource to be more clear.

gerard-queralt (changes requested):
`Application.cpp` seems to have some temporary code

ericdacal:
Removed temporary code from Application

gerard-queralt (approved):
Thank you very much!

AuxPort66 (approved):

PR #30, opened 2023-01-05
Merge InspectorWindow into scenemanagement

markmeliaa:


sncarbo (approved):

sncarbo (commented):
I think there should be an if-statement inside the GameObject Update to update its components only if the GameObject is active. I don't know to what extent this is the case, but I imagine that if the GameObject is not active its components are not active either.

diegusmejia117 (approved):
I also believe that we might need to deactivate the gameobjects components, so that we dont use those resourrces unless needed.

totisama (approved):

markmeliaa:
@sncarbo That was how it was done before, but I changed it because, with the addition of the "enabled" bool in the GameObject class, now UpdateGameObjectAndDescendants() from ModuleScene first checks if a GameObject is enabled, and if it is not, neither them nor their children will call their Update method.

The only situation in which the Update method of a GameObject will be called through UpdateGameObjectAndDescendants() is when it itself is enabled and all their parents were enabled too (because if any of their parents were not enabled, the recursive funcion would've stopped with them), and thus, when Update is called, active is always true (because neither of their parents were disabled). Doing (if (active)) before updating the components of a GameObjects was an unnecessary check that was always true.

This may be a bit confusing, please if you need a more exhaustive explanation lmk, I will approve the pull request once you agree.

PR #33, opened 2023-01-05
feature: File system dialogs

elbuho12Jesus:
Feature: new WindowImporter.cpp , WindowImporter.h class for the dialog (ImGuiFileDialog) and a library for handling icons for the buttons (FontIcons).
file_system_dialog


Piterga12 (changes requested):
In the WindowMainMenu.cpp line 61 you need to add {} to the if

gerard-queralt (changes requested):
- try to keep the 120 limit in `WindowImporter.cpp`
- there is an empty `WindowImporter.cpp` in `Source`
- some files seem to be missing in `ImGuiFileDialog`, since the project can't find them and does not compile

PR #38, opened 2023-01-07
Merge some fixes into scenemanagement

markmeliaa:


markmeliaa:
The 
if (gameObject != root)
		gameObject->Update();
added to the ModuleScene.cpp is because, in the future, each component should draw itself in the Inspector window through its Update() method, and this will prevent the root from drawing its transform and other bugs that may happen if the root updates.

totisama (approved):
good changes

sncarbo (approved):

PR #54, opened 2023-01-10
Scenemanagement editor control window

diegusmejia117:
Created new EditorControlWindows with buttons  Play, Pause and SkipForward (all with calls to functions in ModuleScene).


sncarbo (approved):

totisama (approved):

markmeliaa:
Why is there no Stop button and there is a SkipForwardScene?

PR #55, opened 2023-01-10
Merge display-models-hierarchy into scenemanagement

sncarbo:


markmeliaa (changes requested):
- Why did you declared the Init method abstract in Component.h? For now, only the ComponentMeshRenderer uses it, I think that it should not be abstract but rather be created as a base class in Component and if any Component needs to init anything, add an override Init to those classes.

- In ModuleRender.cpp, the function you created called DrawGameObject() calls to every component the GameObject has (being them Transforms, Meshes, Lights, Cameras), it converts them to MeshRendererComponent and then draws them to the scene, this is wrong.
Instead of that I would do the following:
- Create the Draw() method in the Component.h/Component.cpp file, so every component has a Draw() method, but do not make it abstract like Update, because not every component will draw itself (for example, now the Transform is not drawn through this. MeshRendererComponent will override this Draw() method.
- In the DrawGameObject() function, call the Draw() method for every component, no need to call the components meshRenderers and also no need to convert them to MeshRenderComponent*, if they are MeshRendererComponent they will call the Draw() overrided by MeshRendererComponent, if not, the base Draw() created in the Component.h/Component.cpp.

If I am mistaken in anything or I interpreted something wrong, please let me know.

sncarbo:
@markmeliaa , in relation to your requests:

- The Draw method should only be in the ComponentMeshRenderer class as it is a functionality exclusive to it. I don't see the point of another class using it.

- Regarding the ModuleRender Update loop, it does it well. Although its static type is Component*, the ComponentMeshRenderer* downcasting allows the Draw method to be called only by objects whose dynamic type is ComponentMeshRenderer*. This is not the optimal way to do it (it loops through all the Components unnecessarily), since you could create a GetComponents method of GameObject that returns only components of a certain type (like, for example, ComponentMeshRenderer) and loop through them. I can do the latter if you prefer. 

- Regarding the Init method, I do see the need for it to be virtual, as the Constructor is used for the most trivial initialisations, but in many cases initialising everything in the Constructor, when dealing with derived classes, can lead to errors. I recommend you to visit the following link: https://stackoverflow.com/questions/3786853/why-use-an-initialization-method-instead-of-a-constructor

markmeliaa:
- The Draw method is important for some of the components that we will have in the future, such as the colliders for example (or the material, I think it must be drawn too), take a look in Unity, the MeshRenderer is not the only component that renders something into the scene, that is why I think it should be stated in the Component.h/Component.cpp but without making it virtual, because not all the components will be drawn into the scene.

- If what you want to do with the DrawGameObject function is only drawing the MeshRendererComponent and no other component, I recommend you that you use the GetComponent method of the GameObject class, that returns you the Component of the type you introduced as a parameter of the method (in this case, it will search the GameObject for the MeshRenderer component and retrieve it).

- And regarding the Init method, I do see its usefulness, but, as they also say in the link you sent me, it is not necessary that it is always present in every Components (only in the ones that, as you say, could bring to errors or exceptions due to timing) that is why I said that it is useful but it should not be abstract in Component.h/Component.cpp (for example, look at how the Modules given to us by the teachers in the first assignment were made, the Init method was never virtual, because, although is true that it may be necessary in certain situations, it is not always, so their base declaration in the Module interface is not abstract).

Also please, resolve the conflicts that are marked by this pull request, this can't be merged into scenemanagement if those conflicts exist.

sncarbo:
Although the different Components have rendering code, I don't think it is appropriate to call them all Draw, as each one can have a very specific functionality and the name can lead to confusion. They should have a name related to the rendering functionality they implement. I also think it is desirable for Init to be abstract, as I believe that in the future most derived classes will use it. In order to speed up the process, as other groups depend on this work, I have made the requested changes.

markmeliaa (approved):
Thanks :) good job, it was a pretty hard task!

totisama (approved):

sncarbo:
@markmeliaa Thank you Mark, I appreciate it!

PR #57, opened 2023-01-10
feat[EditorControl]

diegusmejia117:
Modified the control buttons to be only 2 with Play/Stop function being used on the Play button.  Also stremlined the functions so they do not need any kind of variable for now.


sncarbo (approved):

totisama (approved):

markmeliaa:
Thank you for your work Diego, very appreciated :)

PR #73, opened 2023-01-13
Material and Texture full flow

AuxPort66:
- Added Material between Model and Texture flow (Importer/Resource)
- Added Material placeholder between Model.ccp and Mesh.cpp
- Normalized extensions because some assets have the extension on UpperCase and others on LowerCase.
- On some models the file.data get from assimp was the full path of the texture and not only the name (Expected output). Now it's normalized.
- Imported changes from [pbrmaterial-materialMap](https://github.com/Pre-SuperAwesomeEngine/Engine/tree/pbrmaterial-materialMap) to add Materials. And moved to the correct functions on Resource
- Texture now save on binary


PaablooCH (changes requested):
- Source/DataModels/3DModels/Mesh.h there is a include model.h but I don't see any use of model class in that .h 
- Source/DataModels/3DModels/Model.h in my opinion I think normal is not a component of material it should be inside of model or mesh, but I will talk it with miki
- Source/DataModels/Resources/ResourceModel.h nummaterials doesn't follow camelCase also there is a schema in the slides that tells us one model has only one material so having a variable that informs you how many materials exists when it can be only one is redundant maybe it could be a bool or just deleted and check if material its null or not, same to the vector materialsUIDs.

AuxPort66:
- #include model.h on the placeholder class mesh.cpp is there because model.h have the placeholder material struct, used on the call NewDraw of mesh.cpp on a parameter.
- In my opinion Materials should have all the Maps (NormalMap, SpecularMap, OcclusionMap, etc) We can talk it about but other engines (Blender, Unity, etc) put all that references to textures inside the material.
- CamelCase numMaterial solved. With the assimp import of a model (inside ModelImporter) some models have more than 1 material. And I need to save the size of that vector for binary read/write purposes.

PaablooCH (approved):

csandra24 (dismissed):

gerard-queralt (approved):

ericdacal (approved):

PR #77, opened 2023-01-13
Filesystem window file browser

Piterga12:
The File Browser properly works and i have merged the WindowImporter into WindowFileBrowser, so that the button for WindowImporter is in WindowFileBrowser, so WindowImporter.cpp is not needed anymore and has been deleted (as mentioned now is inside WindowFileBrowser)


gerard-queralt (commented):
- Aren't WindowFileBrowser's blocks of code between lines `30` and `44` and lines `53` and `61` equivalent to each other? Are both necessary, or can we use one instance of file dialog?
- Should the file browser be in the Main Menu or as a regular window in ModuleEditor? I think it's better to add it to the ModuleEditor, but I could be wrong

gerard-queralt (changes requested):
Also please remove the commented code since we don't intend to use it

gerard-queralt (changes requested):
Thank you! Checking again, I think there's still some repeated code. There are two calls to `OpenDialog` and to `Display`, some using `ImguiFileDialog::Instance` and others using the member `fileDialog`. Since in the creator you set all the styles for the member, my guess is that all calls to `ImguiFileDialog::Instance` should be replaced by calls to the member. Also, the `if` block run when the action is OK was updated in the `filesystem` branch to call the import method, please check there and copy it in your code.

gerard-queralt (approved):
Thank you for your work!

ericdacal (approved):

PR #78, opened 2023-01-13
Request Resource returns weak_ptr

gerard-queralt:
`ModuleResources::RequestResource` returns weak pointer instead of shared. Also made minimal changes to allow the project to compile.
Important to bear in mind that a `weak_ptr` can't be assigned to `nullptr`; instead, use the default constructor as shown in `ModuleResources::RequestResource`


gerard-queralt:
@markmeliaa I made minor changes to the `ComponentMeshRenderer` so it has a `weak_ptr` instead of a shared one. Make sure I didn't break anything, and if I didn't, you can use it as reference for future components.

markmeliaa:
As far as I can see, everything seems right, thank you Gerard for that change 👍

markmeliaa (approved):

gerard-queralt:
I just realized that calling `weak_ptr::lock` without a check could thorw a NPE. I will add that check, so don't merge just yet, sorry!

gerard-queralt:
After checking the documentation and doing some tests, it seems that calling `weak_ptr::lock` on an empty pointer throws no error; instead, it just returns another empty pointer. Calling that one does raise a NPE, but it would happen regardless if we are using shared or weak pointers, so no new errors should be added after merging this branch. All clear!

markmeliaa:
Great, thanks for looking into it!

markmeliaa (approved):

AuxPort66 (approved):

PR #79, opened 2023-01-13
Divide ModuleScene in ModuleScene and Scene

markmeliaa:
(Hay tantos cambios porque he tenido que cambiar todas las llamadas que se hacía a ModuleScene por 
ModuleScene->GetLoadedScene() para que no pete todo)
(También los meta, pero bueno, esos ya van a estar siempre)


totisama (approved):

sncarbo (changes requested):
I saw two things I wanted to comment on:

- In ModuleScene::Save() the iterator-based loop should be replaced by a range-for loop, as the latter is more efficient. It can be as follows:

for(std::vector<Scene*>& savedScene : savedScenes)
{
    if(savedScene->GetUID() == lastSavedScene->GetUID())
    {
        savedScene = loadedScene;
    }
}

- In ModuleScene::Load(), if you do "loadedScene = savedScenes[0]", wouldn't it always load the first saved scene? If there were more saved scenes it wouldn't give you the option to choose another one. I don't know if I'm wrong about the latter, tell me if I am.

markmeliaa:
Both Save and Load are now just placeholders, they will be modified once I talk with filesystem about how to link those functions with their funciontality, so their contents will be completely erased it was just so the basic funcionality could be tried out.

sncarbo (approved):

sncarbo:
Okay then @markmeliaa

PR #85, opened 2023-01-14
Lights - GameObject linkage

MacXxs:
Now Lights use the transform components from gameObjects to draw themselves and its aim.
Added two functions to the ComponentTransform class to get the Forward vector from the global and local matrices


markmeliaa (approved):

PaablooCH (changes requested):
- Source/DataModels/3DModels/Mesh.cpp when you do things like p1.position = float4 it will be best to set those variables through seter. I doubt if things like pointLights.size() should be .GetSize() idk.
- Source/DataModels/Components/ComponentSpotLight.cpp in ComponentSpotLight idk if it's github but can you put color variable under radius variable?
- Source/DataModels/Components/ComponentSpotLight.h same here.

MacXxs:
> * Source/DataModels/3DModels/Mesh.cpp when you do things like p1.position = float4 it will be best to set those variables through seter. I doubt if things like pointLights.size() should be .GetSize() idk.
> * Source/DataModels/Components/ComponentSpotLight.cpp in ComponentSpotLight idk if it's github but can you put color variable under radius variable?
> * Source/DataModels/Components/ComponentSpotLight.h same here.

I answer point by point in the same order respectively:
- pl is an struct, not an class object that's why i initiate them like this. Furthermore, they are here just to test if lights bind correctly with the shaders, on the final engine this values will be initialized by using the already existing getters from the different componentLight classes. On the next merge, this class will be removed so any changes done here are just for merging purposes later during development of the engine.
- Yes i continued the line under the radius variable for verbosity purposes (not exeeding the 120 limit and it's preferred to arrange the variables on the same tab space)
- Same as previous point.

PaablooCH (approved):
Sure, thanks

PR #86, opened 2023-01-14
feature: Camera Window options

Dedado3000:


xavi1111 (approved):
Needed for seeing enourmous objects. With this there is a better range of camera distances.

OriolSors (approved):

PR #92, opened 2023-01-14
Creation of new quadtree subnodes + refactoring of subdivide method

xavi1111:


Dedado3000 (approved):
We need to define a full debug  quadtree before a push

PR #98, opened 2023-01-15
Remove component[GameObject] and Create Light with Parameters[GameObject]

UriMurcia:


MacXxs (changes requested):
The code is good! Only a few comments:

We need to have the CreateComponent functions as generic as possible (by order of the teachers), so instead of creating a function "CreateComponentLight(LightType::POINT, color, intensity);"
we should use the default CreateComponentLigjht(LightType::POINT) and then use the setters for the light components.

I don't know if it is done un purpose, but the radius isn't maintained when you change a light type from point to spot or viceversa, is there any reason why it's reseted to default?

On the RemoveComponent function, why do you cicle through the components vector instead of doing components.delete(compToDelete) directly? (previously checking that compToDelete is not nullptr)

The rest is perfect!! the only issues I have are the ones i commented! Thanks and great work!

UriMurcia:
> reviously checking that compToDelete is not nullptr

Do not vectors need to iterate through them to know which element is being deleted? Is there a method to delete an element by just passing the pointer of the element that wants to be deleted? I'll check that is not nullptr though!

And for setting the new light with the previous parameters I should do something like this if I want to set the parameters:
image
Do you know any other idea to set the parameters? I think there is not an easier way yet, should we create one?

Radius maintained with the new code!

MacXxs (approved):

PR #101, opened 2023-01-15
feat: meshes deleted when deleting model asset

gerard-queralt:


Piterga12 (commented):
In ModuleResources.cpp line 238 there's an error saying that skyboxImporter is undefined (idk if that's supposed to be like this for now)
Then again I've made the same fix as in the other pull request

gerard-queralt:
I just realized my changes weren't uploaded properly. I'll do it again, sorry!

gerard-queralt:
Everything should be ok now!

AuxPort66 (approved):

PR #104, opened 2023-01-16
fix: Fixed the path for saving and loading the scene

Piterga12:


gerard-queralt (approved):
Thank you for debugging!

AuxPort66 (approved):

PR #112, opened 2023-01-17
Refactor: raw pointers replaced with smart pointers

gerard-queralt:
:GerardPatPat:


gerard-queralt:
A common mistake I've done while doing this refactor is making the return type of methods a non-const reference. I think I have fixed all such cases, but I'd like if you could pay special attention to that while reviewing this request.

markmeliaa:
When launching the engine, the root in the hierarchy shows nothing as their children, although they appear rendered on the scene, idk why that happens.

PR #115, opened 2023-01-18
Material and Mesh imGUI importer

AuxPort66:
- Add Mesh imGUI, remove and import file browser
- Add Material imGUI functionality
- Add material delete and import file browser
- It was necessary to add some basic functionality in the file system flow (Update .bin when Resource changes and special case for Material because we also changed the .mat of Assets/Materials)


Source/DataModels/Windows/EditorWindows/ImporterWindows/WindowTextureInput.cpp
```diff
@@ -0,0 +1,35 @@
...
+		switch (textureType)
+		{
+		case TextureType::DIFFUSE:
+			materialAsShared->SetDiffuseUID(uidTexture);
+			break;
+		case TextureType::NORMAL:
+			materialAsShared->SetNormalUID(uidTexture);
+			break;
+		case TextureType::OCCLUSION:
+			break;
```
markmeliaa, line 27: Why the occlusion is not set? Is this alright?

Source/DataModels/Components/ComponentMaterial.cpp
```diff
@@ -40,80 +41,280 @@ void ComponentMaterial::Draw()
...
+				}
+			}
+		}
+		//bool hasNormal;
+		//std::shared_ptr<ResourceTexture> textureAsShared = textureNormal.lock();
+		//textureAsShared ? hasNormal = true : hasNormal = false;
+
+		//ImGui::Checkbox("Normal slider", &hasNormal);
+		ImGui::SliderFloat("Normal", &normalStrength, 0.0f, 1.0f, "%.1f", ImGuiSliderFlags_AlwaysClamp);
+		ImGui::Separator();
```
markmeliaa, line 313: this ImGUI::Separator should not be here

Source/DataModels/Windows/EditorWindows/ImporterWindows/WindowTextureInput.cpp
AuxPort66: We haven't processed the occlusion yet, that's why it's empty

Source/DataModels/Components/ComponentMaterial.cpp
```diff
@@ -40,80 +41,280 @@ void ComponentMaterial::Draw()
...
+
+		ImGui::Separator();
```
markmeliaa, line 315: And this ImGui separator should be outside the }

Source/DataModels/Windows/EditorWindows/ImporterWindows/WindowTextureInput.cpp
markmeliaa: oke!

markmeliaa:
Everything is good, just a couple separators for aesthetic purposes and that's all.

PR #131, opened 2023-01-24
refactor: created ComponentWindow hierarchy and removed Component::Display

gerard-queralt:
The refactor that no one asked for


gerard-queralt:
Oh I also deleted ModuleTexture, forgot about that. Turns out nothing depended on it

markmeliaa (approved):

markmeliaa (commented):
Wait, one question that I never asked, why this refactor?

gerard-queralt:
The point was to removed the dependency on ImGui from the components, to better organize the code and have our classes follow the [single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).
I think all references to ImGui should be kept in the `ModuleEditor` and the `Window`s hierarchy, and the components shouldn't need to know how to render themselves in our graphical interface (which, in the case of `ComponentTransform`, was more than half of the code in the source file). Also - and this is a bit more picky - the former `Display` methods in the components could be called outside of an ImGui context, leading to crashes. Sure, we wouldn't do that, but this refactor removes that possibility entirely, making the application more robust.
Of course, I did this on my own and I welcome the possibility of you disagreeing with me; as I said, no one asked for this refactor. But I hope to have shown why I think it improves our code, even if it's just picky and nerdy design stuff :P

markmeliaa:
Okay, I will try its functionality later and if everything works great I'll approve it, sorry it's just that I never knew why this refactor was happening so I needed an explanation.

gerard-queralt:
Oh no need to apologize, it's important we are all in the same page; quite the opposite, sorry for not explaining myself better. And as I said, now that I have explained why I did it, feel free to question my reasoning if you think such a refactor is unnecessary.

gerard-queralt:
Also, if this refactor is not approved, we'll need to remove the `ModuleTexture` anyway in some other pull request. It's not a big deal, there's no dependencies as I said, but it's something to keep in mind. Sorry for mixing both of those tasks, I should have removed the dependency in a different branch.

markmeliaa:
No worries!

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentMeshRenderer.cpp
```diff
@@ -0,0 +1,91 @@
...
+}
+
+void WindowComponentMeshRenderer::DrawWindowContents()
+{
+	this->DrawEnableAndDeleteComponent();
+
+	std::shared_ptr<ComponentMeshRenderer> asSharedMeshRenderer =
+		std::dynamic_pointer_cast<ComponentMeshRenderer>(this->component.lock());
+
+	if (asSharedMeshRenderer)
```
markmeliaa, line 25: When a new ComponentMesh is created there is no option to load a new mesh, we need that option back.

markmeliaa:
Everything else is correct, so once those changes are done I will approve this without any issue.

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentMeshRenderer.cpp
markmeliaa, line 25: Also, when the MeshComponent with a mesh is drawn, the whole path and the Remove Mesh button are a bit too lenghty to draw them next to the Remove Comp. button, if it is not an issue, I would rather have it like in the Material Component: the enable/disable button and the remove button of the component in one line and the rest of the component in the next line (if you do not understand it please feel free to ask me).

gerard-queralt: Hopefully I understood you correctly. I made it so the path and remove buttons are only moved to the next line when a Mesh is set, but it would be very easy to have it in both cases. In fact, maybe the `ImGui::SameLine` called in `ComponentWindow::DrawEnableAndDeleteComponent` could be removed entirely, depending on how it looks with future components.

gerard-queralt:
Thank you very much for all your feedback, Mark. I think I fixed everything you found, let me know if I missed anything or you spot something else!

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentMeshRenderer.cpp
markmeliaa: Yep, that was right what I was saying, and as you say, we could discuss about that ImGui::SameLine in the future when we know how other components will look like.

markmeliaa (commented):
Thanks to you for performing the changes so quickly, everything looks fine to me now!

markmeliaa (approved):

AuxPort66 (approved):

PR #137, opened 2023-02-07
Program corrections

DADSILVER:
With this change is allowed the use of multiples programs shaders.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -437,3 +437,5 @@ FodyWeavers.xsd
 
 !**/lib/**
 imgui.ini
+/Game/Lib/Shaders/1f.glsl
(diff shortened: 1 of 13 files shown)


gerard-queralt (changes requested):
First of all, I want to say the `Program` class looks very clean, and I think it's a great idea to encapsulate all that logic in a single place. Great job!
The only thing I see (which might not even be you who did it) is that there are some places where a `std::string` is passed by copy, meaning a copy of the original will be created, instead of passing it by const reference. I think it's limited to `Program` and `ModuleProgram`, if I didn't miss any such cases. It'd be great if you could change it. Again, thank you for your work!

gerard-queralt (approved):
Thank you very much for your work!

PR #138, opened 2023-02-08
refactor: shared pointers removed from trivial places, replaced by unique and raw pointers

gerard-queralt:
I have removed all of the uses of shared pointers in those classes where it was trivial to do so. If I didn't miss any of those cases, the only places in the engine where they remain are the Quadtree class, the ModuleProgram and Program classes, and the ModuleScene, Scene, GameObject and Component classes, pending a more supervised refactor.

### DIFF
--- a/Source/Application.cpp
+++ b/Source/Application.cpp
@@ -17,18 +17,16 @@ constexpr int FRAMES_BUFFER = 50;
 Application::Application()
 {
 	// Order matters: they will Init/start/update in this order
-	modules.push_back(window = std::make_shared<ModuleWindow>());
-	modules.push_back(editor = std::make_shared<ModuleEditor>());
-	modules.push_back(input = std::make_shared<ModuleInput>());
-	modules.push_back(program = std::make_shared<ModuleProgram>()); 
-	modules.push_back(fileSystem = std::make_shared<ModuleFileSystem>());
-	modules.push_back(resources = std::make_shared<ModuleResources>());
-	modules.push_back(engineCamera = std::make_shared<ModuleEngineCamera>());
-	modules.push_back(scene = std::make_shared<ModuleScene>());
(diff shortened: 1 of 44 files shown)


sncarbo:
@gerard-queralt In the method "void WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject)", gameObject could be const? I don't know if it will be altered inside the method.

gerard-queralt:
Currently no, since `shared_from_this` can't be invoked on a `const` pointer (or at least I didn't find a way to do it) and `GameObject::GetGameObjectsInside`, called in the method, is not `const`, so it can't be called on a `const` object. However, all calls to `shared_from_this` must be removed once we are done with Scene and company, and maybe the method in GameObject can be made `const`; then the parameter could be `const` with no issue, so we can keep it in mind when we start working with Scene.
There might be other instances in my changes where the pointer could be `const` and it's not, so I appreciate if you could have a look. I re-checked it and didn't find any such case, but we all know it's hard to review our own work :P

sncarbo:
@gerard-queralt So, if it's ok with you, I'll change it to `const` and delete `shared_from_this` once I do the SceneManagement refactor.

markmeliaa (approved):
Nice and hard work Gerard, thanks for this. If you need any help removing smart ptrs from the Scene stuff, lmk. 🫡

gerard-queralt:
@sncarbo  I don't think there would be any issue, so feel free to do so. But as I said you'd also have to change `GameObject::GetGameObjectsInside` to be const, and that might not be possible since the name doesn't sound very const to me, but I didn't look at the method so I don't know ^^

gerard-queralt:
Thank you Mark! All help is appreciated, since it will be the most delicate part of this refactor, so it'd be great if we could work together on it. We can talk about it in class :)

PR #143, opened 2023-02-12
Engine Refactor about the use of Pointers

sncarbo:
Refactor performed with the purpose of discarding unnecessary Smart Pointers and substitute them with Raw or Unique Pointers.

### DIFF
--- a/Source/DataModels/Components/ComponentBoundingBoxes.cpp
+++ b/Source/DataModels/Components/ComponentBoundingBoxes.cpp
@@ -9,7 +9,7 @@
 #include "Scene/Scene.h"
 #include "FileSystem/Json.h"
 
-ComponentBoundingBoxes::ComponentBoundingBoxes(bool active, const std::shared_ptr<GameObject>& owner)
+ComponentBoundingBoxes::ComponentBoundingBoxes(bool active, GameObject* owner)
 	: Component(ComponentType::BOUNDINGBOX, active, owner, false)
 {
 	localAABB = { {0 ,0, 0}, {0, 0, 0} };
@@ -20,8 +20,8 @@ ComponentBoundingBoxes::ComponentBoundingBoxes(bool active, const std::shared_pt
 
 void ComponentBoundingBoxes::CalculateBoundingBoxes() 
 {
-	std::shared_ptr<ComponentTransform> transform =
-		std::static_pointer_cast<ComponentTransform>((this)->GetOwner().lock()->GetComponent(ComponentType::TRANSFORM));
+	ComponentTransform* transform =
+		static_cast<ComponentTransform*>((this)->GetOwner()->GetComponent(ComponentType::TRANSFORM));
 	objectOBB = OBB(localAABB);
 	objectOBB.Transform(transform->GetGlobalMatrix());
 	encapsuledAABB = objectOBB.MinimalEnclosingAABB();
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -409,7 +409,7 @@ bool GameObject::RemoveComponent(const Component* component)
 	return false;
 }
 
-Component* GameObject::GetComponent(ComponentType type)
+Component* GameObject::GetComponent(ComponentType type) const
 {
 	for (std::vector<Component*>::const_iterator it = components.begin(); it != components.end(); ++it)
 	{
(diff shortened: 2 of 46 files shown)


Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -180,53 +179,64 @@ void GameObject::InitNewEmptyGameObject()
...
+	if (!IsAChild(child.get()))
 	{
-		children.push_back(child);
 		child->active = (this->IsActive() && this->IsEnabled());
+		children.push_back(std::move(child));
 	}
 }
 
-void GameObject::RemoveChild(const std::shared_ptr<GameObject>& child)
+std::unique_ptr<GameObject> GameObject::RemoveChild(const GameObject* child)
```
markmeliaa, line 210: Why `RemoveChild` now returns a `unique_ptr`?

markmeliaa (approved):
I see no mistakes in anything related to Components, Gameobjects, Scene or Windows, well done!

Source/DataModels/Components/ComponentBoundingBoxes.cpp
```diff
@@ -20,8 +20,8 @@ ComponentBoundingBoxes::ComponentBoundingBoxes(bool active, const std::shared_pt
 
 void ComponentBoundingBoxes::CalculateBoundingBoxes() 
 {
-	std::shared_ptr<ComponentTransform> transform =
-		std::static_pointer_cast<ComponentTransform>((this)->GetOwner().lock()->GetComponent(ComponentType::TRANSFORM));
+	ComponentTransform* transform =
+		static_cast<ComponentTransform*>((this)->GetOwner()->GetComponent(ComponentType::TRANSFORM));
```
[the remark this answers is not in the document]
sncarbo: The first one tries to perform a const cast first, if it fails it tries a static cast, if it fails it tries a combination of static and const cast, if it fails it tries a reinterpret cast, and, finally, it tries a combination of reinterpret and const cast. In general, const casts are bad practice because it forces the const to be removed.

The latter always makes a static cast.

Source/DataModels/GameObject/GameObject.cpp
sncarbo: We decided to introduce this change because the `void GameObject::SetParent(GameObject* newParent)` method needs a pointer to the children that have been removed from one GameObject in order to add it to the other.

sncarbo:
@markmeliaa Thank you very much for your feedback Mark! It's important to discuss the changes, I really appreciate it.

Source/DataModels/Components/ComponentBoundingBoxes.cpp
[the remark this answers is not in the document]
gerard-queralt: To add to what Santi said, if I'm not mistaken `static_cast` performs no run-time checks, which makes it less safe, but also more efficient if you know for sure the cast can be made. At least that's what I gathered reading around. Here are some of the references I used: [cppreference](https://en.cppreference.com/w/cpp/language/static_cast), [Wikipedia](https://en.wikipedia.org/wiki/Static_cast), [StackOverflow](https://stackoverflow.com/questions/28002/regular-cast-vs-static-cast-vs-dynamic-cast).

Source/DataModels/GameObject/GameObject.cpp
gerard-queralt: To be more specific: when the method `SetParent` is run, the following sequence was executed (more or less):
```
parent->RemoveChild(this);
parent = newParent;
parent->AddChild(std::unique_ptr<GameObject>(this));
```
The issue with the above code is that, after the first line is executed, `this` is deleted by their former parent, since the unique pointer no longer has references. Therefore the method continues executing on deleted memory, resulting in undefined behaviour and ending in a crash.
By making the parent return the unique pointer to the child they just removed, we can make the transfer of ownership. Now the code above reads:
```
std::unique_ptr<GameObject> ptrToThis = parent->RemoveChild(this);
parent = newParent;
parent->AddChild(std::move(ptrToThis ));
```
Of course, no one is forced to store the returned pointer, so the method can be used like it was before if this transfer of ownership is not needed. Hope I explained myself properly!

Source/DataModels/Components/ComponentBoundingBoxes.cpp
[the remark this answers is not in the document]
markmeliaa: Thank you both for your answer :)

Source/DataModels/GameObject/GameObject.cpp
markmeliaa: Ooooooh, now I get it, I was really lost about why returning the `unique_ptr` but it totally makes sense, thanks again!

Dedado3000 (approved):
I don't see any problems at Quadtree changes, good job

gerard-queralt:
Thank you both for your reviews!

PR #144, opened 2023-02-14
fix: can load scene

gerard-queralt:
Fixed Quadtree, ComponentTransform and removed unnecessary unload in destructor of ComponentMaterial (that for some reason caused a crash, probably because they are deleted after ModuleResources)

### DIFF
--- a/Source/DataStructures/Quadtree.cpp
+++ b/Source/DataStructures/Quadtree.cpp
@@ -266,7 +266,7 @@ void Quadtree::RedistributeGameObjects(const GameObject* gameObject)
 	// GameObject redistribution part
 	gameObjects.push_back(gameObject);
 
-	for (std::list<const GameObject*>::iterator it = gameObjects.begin(); it != gameObjects.end();)
+	for (std::list<const GameObject*>::iterator it = std::begin(gameObjects); it != std::end(gameObjects);)
 	{
 		if (*it)
 		{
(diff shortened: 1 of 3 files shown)


markmeliaa (approved):
This works fine, no crashes happen now when a new scene is loaded.

Source/DataStructures/Quadtree.cpp
```diff
@@ -266,7 +266,7 @@ void Quadtree::RedistributeGameObjects(const GameObject* gameObject)
 	// GameObject redistribution part
 	gameObjects.push_back(gameObject);
 
-	for (std::list<const GameObject*>::iterator it = gameObjects.begin(); it != gameObjects.end();)
+	for (std::list<const GameObject*>::iterator it = std::begin(gameObjects); it != std::end(gameObjects);)
```
PaablooCH, line 269: hi, one question, why do you use the iterator instead doing for(Gameobject * obj : gameobjects) ?

PaablooCH:
Task linked: [CU-865bq6duq Change flag from GAME to ENGINE](https://app.clickup.com/t/865bq6duq)

Source/DataStructures/Quadtree.cpp
gerard-queralt: If I'm not mistaken, it's because inside that loop some of the elements are removed from the vector, and the only way to do that while looping is by using iterators. But @Dedado3000 or @OriolSors can probably correct me on that if I'm wrong

OriolSors: Yes, exactly. You cannot remove an element from a container while iterating it in a for-each loop.

PaablooCH: okey perf, thanks

PaablooCH (approved):

PR #145, opened 2023-02-14
chore: ticket #865bqcbz1

Dedado3000:
removed commented code

### DIFF
--- a/Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
+++ b/Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
@@ -88,6 +88,49 @@ void WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject)
                 App->scene->GetLoadedScene()->CreateCameraGameObject("Basic Camera", gameObject);
         }
 
+        //TODO: make this work
+        /*
+        if (gameObject != App->scene->GetLoadedScene()->GetRoot()) // The root can't be neither deleted nor moved up/down
+        {
+            std::vector<std::weak_ptr<GameObject> > parentsChildren = gameObject->GetParent().lock()->GetChildren();
+
+
+            if (ImGui::MenuItem("Move Up"))
+            {
+                if (parentsChildren.size() > 1 && parentsChildren[0].lock() != gameObject)
+                {
+                    for (int i = 0; i < parentsChildren.size(); ++i)
+                    {
+                        std::shared_ptr<GameObject> asShared = parentsChildren[i].lock();
+                        if (asShared && asShared == gameObject)
+                        {
+                            std::iter_swap(parentsChildren[i - 1].lock(), parentsChildren[i].lock());
+                            App->scene->SetSelectedGameObject(parentsChildren[i - 1].lock());
+                            break;
+                        }
+                    }
+                }
+            }
+
+            if (ImGui::MenuItem("Move Down"))
+            {
+                if (parentsChildren.size() > 1 && parentsChildren[parentsChildren.size() - 1].lock() != gameObject)
+                {
+                    for (int i = 0; i < parentsChildren.size(); ++i)
+                    {
+                        std::shared_ptr<GameObject> asShared = parentsChildren[i].lock();
+                        if (asShared && asShared == gameObject)
+                        {
+                            std::iter_swap(parentsChildren[i].lock(), parentsChildren[i + 1].lock());
+                            App->scene->SetSelectedGameObject(parentsChildren[i + 1].lock());
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+        */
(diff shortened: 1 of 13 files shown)


PaablooCH:
Task linked: [CU-865bqcbz1 Remove all commented-out code](https://app.clickup.com/t/865bqcbz1)

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
```diff
@@ -88,48 +88,6 @@ void WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject)
                 App->scene->GetLoadedScene()->CreateCameraGameObject("Basic Camera", gameObject);
         }
 
-        /*
```
markmeliaa, line 91: Can you restaurate this long commented section please? This is something that, if uncommented and used in the engine, gives a compiler error, but once it is fixed it sould work fine, so this is not an unnecessary coment but rather a "lets fix this later" one because there was no time to fix it for the first assignment.

gerard-queralt: OK Mark, I'll undo this change and maybe try and make it work

markmeliaa (approved):

gerard-queralt:
Well, I have some good news, some bad news, and some slightly good news. The good news is that that block of code now compiles! The bad news is that it does nothing. The slightly good news is that I'm pretty sure I know why: when calling `std::iter_swap`, we are only swapping on the vector created (and later destroyed) in the method, without propagating those changes to the actual vector in the GameObject. I can think of a few solutions for this:

- Give WindowHierarchy access to the vector of children using `friend class` annotation. This is the quickest solution, but I don't really like the idea of accessing the unique pointers of another class
- Store the vector of children in `WindowHierarchy` and only change it when the selected GameObject changes. I was about to say that this would waste considerable memory, but writing it I realize that the changes wouldn't propagate to the GameObjects anyway, making this option useless :)
- Create a method in GameObject to encapsulate the functionality of `std::iter_swap`. The method would recieve two objects (maybe raw pointers to GameObjects, maybe iterators) and perform the swap on the actual vector. I think this is the best solution, though some issues may come up because the inner workings of `std::iter_swap` may not merge well with unique pointers, so it should be done with some care

Let me know what you think! I'll upload the version with the compiling code so we still remove all commented-out code in this request, and maybe create a new task to make this work

markmeliaa:
Let me take a look to the code real quick, I was performing some changes in scenemanagement so we can bring its corrections asap to develop.

markmeliaa:
> Well, I have some good news, some bad news, and some slightly good news. The good news is that that block of code now compiles! The bad news is that it does nothing. The slightly good news is that I'm pretty sure I know why: when calling `std::iter_swap`, we are only swapping on the vector created (and later destroyed) in the method, without propagating those changes to the actual vector in the GameObject. I can think of a few solutions for this:
> 
> * Give WindowHierarchy access to the vector of children using `friend class` annotation. This is the quickest solution, but I don't really like the idea of accessing the unique pointers of another class
> * Store the vector of children in `WindowHierarchy` and only change it when the selected GameObject changes. I was about to say that this would waste considerable memory, but writing it I realize that the changes wouldn't propagate to the GameObjects anyway, making this option useless :)
> * Create a method in GameObject to encapsulate the functionality of `std::iter_swap`. The method would recieve two objects (maybe raw pointers to GameObjects, maybe iterators) and perform the swap on the actual vector. I think this is the best solution, though some issues may come up because the inner workings of `std::iter_swap` may not merge well with unique pointers, so it should be done with some care
> 
> Let me know what you think! I'll upload the version with the compiling code so we still remove all commented-out code in this request, and maybe create a new task to make this work

I also see the last option as the best one, let me see if I could make it work now and, if not, we can take a deeper look into it when you are available.

gerard-queralt:
> Let me take a look to the code real quick, I was performing some changes in scenemanagement so we can bring its corrections asap to develop.
> 
> Its weird that it does not do anything now, it used to swap both desired gameobjects but without changing its transform, which broke the engine because it altered the quadtree behaviour iirc.

If I had to guess, the previous version of `GameObject::GetChildren` returned a non-const reference to the children vector, so the swap was performed on the actual member of the GameObject instead of a copy, which is what's happening now

markmeliaa:
Created a [new branch](https://github.com/Horizons-Games/Axolotl-Engine/tree/moveup-movedown-gameobjects) to fix this, so we can merge and close this pull request.

markmeliaa (approved):

xavi1111 (approved):
Good practice to remove commented code. Having comented code defeats the pourpose of version control so really good change. Maybe the changes to WindowHierarchy shouldn't have been done in this branch but nontheless great job.

PR #146, opened 2023-02-14
refactor: ModuleProgram uses unique pointers

gerard-queralt:
Change ModuleProgram so its `Program` members are unique pointers instead of shared, and are accessed as a raw pointer from the outside. Also made some small changes, like removing `char*` in favour of `std::string` and loading the shader source files through `ModuleFileSystem`.


PaablooCH:
Task linked: [CU-865bpj35x Refactor: unique + raw pointers in ModuleProgram and Program](https://app.clickup.com/t/865bpj35x)

AuxPort66 (approved):

AlexisDieu (approved):
I tested and checked all i need , everything seems working for me !

PR #147, opened 2023-02-14
feat: GameObjects can now be moved up and down in the hierarchy

markmeliaa:
In order to approve this pull request, first #145 should have been approved and merged.

### DIFF
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -487,7 +487,7 @@ void GameObject::MoveUpChild(GameObject* childToMove)
 		if ((*it).get() == childToMove)
 		{
 			std::iter_swap(it - 1, it);
-			App->scene->SetSelectedGameObject((it - 1)->get());
+			App->scene->SetSelectedGameObject((*(it - 1)).get());
 			break;
 		}
 	}
@@ -502,7 +502,7 @@ void GameObject::MoveDownChild(GameObject* childToMove)
 		if ((*it).get() == childToMove)
 		{
 			std::iter_swap(it, it + 1);
-			App->scene->SetSelectedGameObject((it + 1)->get());
+			App->scene->SetSelectedGameObject((*(it + 1)).get());
 			break;
 		}
 	}
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -484,7 +484,7 @@ void GameObject::MoveUpChild(GameObject* childToMove)
 		it != std::end(children);
 		++it)
 	{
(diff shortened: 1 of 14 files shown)


gerard-queralt:
Nice catch on that last commit, definitely looks better now! Though I don't feel like I should be one of the reviewers of this request, since I've helped with it. I'll tag some more people ;)

markmeliaa:
> Nice catch on that last commit, definitely looks better now! 

Thanks! Maybe it looks a bit picky but it felt weird having two different ways of getting the pointer of the iterator in the same function.
> Though I don't feel like I should be one of the reviewers of this request, since I've helped with it. I'll tag some more people ;)

I agree, but as there is a task that we found and solved quickly I did not know who to tag, thanks for doing it yourself!

gerard-queralt:
> Thanks! Maybe it looks a bit picky but it felt weird having two different ways of getting the pointer of the iterator in the same function.

Don't worry, I'm not one to call you picky, you know that :P

Dedado3000 (approved):
Is nice to have the moving objectrs back in the hierarchy

markmeliaa:
@PaablooCH

Source/Modules/ModuleRender.h
```diff
@@ -57,6 +56,7 @@ class ModuleRender : public Module
 	const std::vector<std::string> modelTypes = { "FBX" };
 
 	//should this be here?
+	//Note (David Martin): Nope, but at 14/2 still alive
```
PaablooCH, line 59: I know its not yours but pls don't speak in the code

Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -477,3 +477,33 @@ std::list<GameObject*> GameObject::GetGameObjectsInside()
 	}
 	return familyObjects;
 }
+
+void GameObject::MoveUpChild(GameObject* childToMove)
+{
+	for (std::vector<std::unique_ptr<GameObject>>::iterator it = std::begin(children);
```
PaablooCH, line 483: with these moves and other things like delete and create, I think we should use a list instead of a vector for the gameobjects storage

PaablooCH (approved):

Source/DataModels/GameObject/GameObject.cpp
markmeliaa: We can think about that, right now I think that the move up/move down functionality is as used as we need to change gameobject children to a list.

What I mean by this is that the action of adding objects to the children vector will be performed far more times in average than the move up/move down action, so, for now and afaik, we should leave it as a vector due to it being more efficient at managing additions at the end.

Source/Modules/ModuleRender.h
markmeliaa: Yup, I think the same, code comments shouldn't be used to have conversations.

PR #148, opened 2023-02-14
refactor: Scenemanagement corrections

markmeliaa:
Scene management corrections are finally here!

Also, added a couple things that were missing from the last time this pr was open and deleted usless saved scenes that we had in the engine.


Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -193,6 +193,11 @@ void GameObject::SetParent(GameObject* newParent)
 	parent = newParent;
 	parent->AddChild(std::move(pointerToThis));
 
+	// Update the transform respect its parent when moved around
+	ComponentTransform* childTransform = static_cast<ComponentTransform*>
+		(this->GetComponent(ComponentType::TRANSFORM));
+	childTransform->UpdateTransformMatrices();
```
gerard-queralt, line 199: Is this call to `UpdateTransformMatrices` necessary? If I didn't misread the code, this method is called whenever theses matrices are requested, so as an outsider it feels like calling it here is redundant

gerard-queralt: The same thing happens in `Scene.cpp`, which probably means I'm wrong and this is useful, but no harm in asking :P

gerard-queralt:
If I recall correctly, most of your feedback ~~that wasn't my fault~~ was about not using Quaternions outside of transform and try to minimize computations of global matrices, right? You seem to have tackled all those issues, great job! Thank you very much for your effort!

Source/DataModels/GameObject/GameObject.cpp
markmeliaa: What `UpdateTransformMatrices` does is updating the transform of a gameobject and the transform of their children only when needed (another thing that they told us in the feedback is that updating that every frame is excessive), so we added UpdateTransformMatrices() to the places where the transform can be modified:

- When the transform is moved through its ImGui (I just changed this in a commit because I used the wrong function)
- When the transform is created (a gameobject could be created as a child of another gameobject and if that gameobject is in any position, that new transform of the new gameobject should be updated to that position)
- When a gameobject is moved using the drag and drop in the hierarchy.

This are the only cases that come to my mind, and they are there instead of just in the getters because the little axis transform is drawn every frame, so the transform should be updated not only when required by a getter, but at the time they are changed.

Regardless of that, I do not see that having them in the getters is bad (????), yes it may be redundant, but we would be sure that when the pos, rot, sca is retrieved in any other part of the code it would always be updated? I want to know your take on this one.

gerard-queralt: Oh okay, I see what you mean, it makes total sense! I thought they might only be needed when the getters of those methods were called, but I see how that's not the case. Thank you very much for your in-depth explanation!!
Regarding the redundancy, I see how that could be an issue. In general I agree with you that redundancy is good as it makes our code more robust, but as much as it pains me to say it performance should come first. I don't know how expensive those calculations are, though, so it might not hurt, but if they really are redundant and the matrices are updated when they need to be, it's good to keep them in mind; it's a good candidate for optimization down the line.

markmeliaa: `UpdateTransformMatrices` is a recursive function, and it goes as deep as how many children a gameobject has, so in some cases it could, in fact, hinder a little the performance of the engine.

What I've done is comment the `UpdateTransformMatrices` of the getters instead of deleting them so if we have any problem in the future we could just uncomment those and see what to do and where to relocate the calls to `UpdateTransformMatrices`, but for now it works fine without them (and maybe if no problems arise in the future we can delete the completely (?)).

gerard-queralt: Sounds great, thank you! Let me try to run the engine in my machine and I'll approve the pull reques :)

markmeliaa:
Done! It was trivial 😮‍💨

gerard-queralt (dismissed):
Sorry for the delay. Everything seems to run fine, thank you again for your work!

markmeliaa:
Closing this because something about the transforms is not working correctly when a new scene is loaded, I'll take a look at it.

markmeliaa:
Fixed, opening this again!

gerard-queralt (approved):
Thank you again!

gerard-queralt (approved):

Source/DataModels/Components/ComponentTransform.cpp
```diff
@@ -93,63 +66,92 @@ void ComponentTransform::LoadOptions(Json& meta)
...
+	CalculateMatrices();
 }
 
-void ComponentTransform::CalculateLocalMatrix()
+const float3& ComponentTransform::GetGlobalPosition() const
 {
-	localMatrix = float4x4::FromTRS(pos, rot, sca);
+	//UpdateTransformMatrices();
+
+	return globalPos;
```
PaablooCH, line 76: have you considered making this function inline?

Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -236,6 +236,11 @@ void GameObject::SetParent(GameObject* newParent)
 	parent = newParent;
 	parent->AddChild(std::move(pointerToThis));
 
+	// Update the transform respect its parent when moved around
+	ComponentTransform* childTransform = static_cast<ComponentTransform*>
+		(this->GetComponent(ComponentType::TRANSFORM));
```
PaablooCH, line 241: this `this` is redundant

PaablooCH (changes requested):
Good job, just a few things

Source/DataModels/GameObject/GameObject.cpp
markmeliaa: q si pesao

PaablooCH: estos ataques hacia mi persona no hacen falta, btw no esta corregido

PaablooCH (dismissed):
finally well done

Source/DataModels/GameObject/GameObject.cpp
markmeliaa: ya está, contento?

PaablooCH (dismissed):
🇺🇸 : thank you marc
🇪🇸 : ya era hora macho

PaablooCH (approved):
🇬🇧 : nice changes with the warnings, seems all work fine
🇪🇸 : otra review más te busco y te encuentro

markmeliaa:
🇬🇧 : thanks for your words Pablo, and thank you for reviewing this multiple times
🇪🇸 : eres un puto pesao, dejame en paz, buscate una vida

PaablooCH (approved):
🇬🇧 : 😄
🇪🇸 : te avisé, estoy de camino a tu casa

markmeliaa:
🇬🇧 :  Thx for your patience.
🇪🇸 : Soy un pesado lo siento uwu

gerard-queralt (approved):
I feel like I'm going to put you in danger if I don't merge this...

PR #149, opened 2023-02-14
feat: improved on implementation of ENGINE mode

gerard-queralt:
Changed flag from GAME to ENGINE, as explained in #865bq6duq. Also removed flag from Globals and instead converted it into a Preprocessor definition, which led to changes in build configurations. Instead of the old Debug and Release, now there is DebugEngine, ReleaseEngine, DebugGame and ReleaseGame

### DIFF
--- a/Source/Engine.sln
+++ b/Source/Engine.sln
@@ -7,14 +7,20 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Engine", "Engine.vcxproj", 
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
-		Debug|x64 = Debug|x64
-		Release|x64 = Release|x64
+		DebugEngine|x64 = DebugEngine|x64
+		DebugGame|x64 = DebugGame|x64
+		ReleaseEngine|x64 = ReleaseEngine|x64
+		ReleaseGame|x64 = ReleaseGame|x64
 	EndGlobalSection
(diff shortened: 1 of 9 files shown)


PaablooCH:
Task linked: [CU-865bq6duq Change flag from GAME to ENGINE](https://app.clickup.com/t/865bq6duq)

gerard-queralt:
In case anyone is curious, as I said the flag is now specified in Visual Studio's Preprocessor, meaning that it's set "automatically". In case you need to change it or remove it (hopefully telling me beforehand), here is how you can access it:
b6fc46a0941beb6fa43f7f7a915e5ed9

gerard-queralt:
Also, I know the Solution Configuration window looks very small after this change. Here's a [link](https://stackoverflow.com/questions/32377568/can-i-change-the-width-of-the-solution-configuration-combobox-in-the-ide) showing how to change it. Sorry for the trouble!

markmeliaa (approved):
Nice! Thank you for your work Gerard, changes like these are the ones that will polish even more our engine por the possible future users.

sncarbo (approved):
Good work. It's a good idea to start designing and implementing ´ENGINE´ mode with preprocessor directives. 

Keep an eye on it later because these directives should be used carefully (there are some differences between ´#if´ and ´#ifdef´ that need to be discussed), but I think it's a good job.

gerard-queralt:
Thank you both for your feedback! As agreed with @AuxPort66, I'm going to close this pull request and only merge onto `develop` once ENGINE mode is fully implemented.

PR #150, opened 2023-02-16
perf: Constructors/destructors in .cpp files and constructors use initializer lists

markmeliaa:
ClickUp Tasks: https://app.clickup.com/t/865bph4e4 and https://app.clickup.com/t/865bph4g0

I've touched several files, so the leads of the 2nd assignment should look at their scripts to make sure that I didn't destroy anything (I've checked and everything compiles and works well, but I could have missed anything)


PaablooCH:
Task linked: [CU-865bph4e4 All constructors use initializer lists](https://app.clickup.com/t/865bph4e4)

PaablooCH:
Task linked: [CU-865bph4g0 Constructors and destructors in source (.cpp) files](https://app.clickup.com/t/865bph4g0)

PaablooCH:
Task linked: [CU-865bq6dm9 Move Lights structs outside of Scene](https://app.clickup.com/t/865bq6dm9)

Source/Modules/ModuleEngineCamera.cpp
```diff
@@ -383,7 +383,7 @@ void ModuleEngineCamera::Focus(const OBB &obb)
 	float radius = boundingSphere.r;
 	if (boundingSphere.r < 1.f) radius = 1.f;
 	float fov = frustum.HorizontalFov();
-	float camDistance = radius / sin(fov / 2.0);
+	float camDistance = radius / (float)(sin(fov / 2.0));
```
gerard-queralt, line 386: Why is this cast needed? Aren't all variables involved `float`s already?

gerard-queralt:
And another sidenote: while reviewing your request I noticed the destructors in some Components are not virtual/overriden. Since this is not an issue with your branch but rather technical debt that was in `develop`, I've created the ticket #865brqufy to fix it there and wherever else it may happen.

Source/Modules/ModuleEngineCamera.cpp
markmeliaa: For some reason, the `sin()` function called there returns a `double`, not a `float`, which raised a warning.

image

And meanwhile I was responding I discovered why it was returning a `double` instead of a `float`, the `2.0` was missing the `f` at the end. I changed it so we avoid both the warning and the cast.

markmeliaa:
> Amazing job Mark!! Thank you for your effort. As a sidenote, did you removed all the `= default` for any particular reason? Afaik, it's equivalent to just giving the constructor/destructor an empty body, so in terms of results it should be the exact same.

It is, but as one of the ClickUp tasks was to have the constructors/destructors in the .cpp files I decided that, in order to be consistent with the rest of the project, delete those `= default` and create its respective constructor/destructor in the .cpp file, so now all of them are in the same place.

Source/Modules/ModuleEngineCamera.cpp
gerard-queralt: Ohh, makes sense, good catch!

gerard-queralt:
I see your point! It also has the added benefit of making it easier to add logic to those methods if it's ever needed, since you already set up the code blocks. Thank you for your answer!

gerard-queralt (approved):
I'm approving the merge, but I'll wait for the leads of the second assignment to review their functionalities before actually merging, just in case. Thanks again Mark!

PaablooCH (approved):

markmeliaa:
>

markmeliaa:
> I think, before merging, this publication should be taken into account: https://isocpp.org/blog/2023/02/stdinitializer-list-in-cpp-caveats-and-improvements-bartomiej-filipek
> 
> `std::initializer_list`s are not optimal in every option.

`std::initalizer_list` is not what it is being used here, but instead initializer lists in the constructors [check this](https://en.cppreference.com/w/cpp/language/constructor). It is not the same ;)

markmeliaa:
@MacXxs @AuxPort66 @Dedado3000 if you could take a look at this please. You don't have to check every file, just the ones that belonged to your group during the 2nd assignment, to make sure everything is right.

MacXxs (approved):
Everything seems ok, it's only a refactor of the code so i see no changes in functionality (at least for everything regarding to my group). I can't test it right now 'cause i'm at work but if it compiles for you and you've test it i give my OK. If not, i can test it tonight and accept the pull request then, if everything works.

markmeliaa:
> Everything seems ok, it's only a refactor of the code so i see no changes in functionality (at least for everything regarding to my group). I can't test it right now 'cause i'm at work but if it compiles for you and you've test it i give my OK. If not, i can test it tonight and accept the pull request then, if everything works.

Thanks Miki, everything works correctly so no need to check it if you can't, it is just to have an extra pair of eyes to check the scripts.

Dedado3000 (approved):
Quadtree and camera refactor is good

gerard-queralt:
I was about to ask her to have a look, but I suppose we could do that. After all, there's not really any changes to the logic. Let me skim through it again just in case and I'll merge it. Thank you again for your work and sorry for all the merges you've had to do :P

gerard-queralt (approved):

gerard-queralt (approved):

PR #152, opened 2023-02-20
feat: Mouse picking

PaablooCH:
Now the mouse picking uses the quadtree


DADSILVER (approved):

gerard-queralt:
It appears you added/updated some assets and binaries. Are those changes intended?

markmeliaa:
> It appears you added/updated some assets and binaries. Are those changes intended?

Yes, it was because we modifed some meshes in order to check if the raycast was behaving correctly, do we need to delete those?

gerard-queralt:
> Yes, it was because we modifed some meshes in order to check if the raycast was behaving correctly, do we need to delete those?

I don't think there's any need, just wanted to make sure it was intended

Source/DataStructures/Quadtree.cpp
```diff
@@ -506,4 +508,47 @@ std::list<const GameObject*> Quadtree::GetAllGameObjects(const GameObject* gameO
 		familyObjects.insert(familyObjects.end(), objectsChildren.begin(), objectsChildren.end());
 	}
 	return familyObjects;
+}
+
+void Quadtree::CheckRaycastIntersection(std::map<float, const GameObject*>& hitGameObjects, const LineSegment& ray)
+{
+	std::queue<const Quadtree*> quadtreeQueue;
+	quadtreeQueue.push(this);
+
```
OriolSors, line 517: Perfect non-recursive implementation!

Source/Modules/ModuleRender.cpp
```diff
@@ -203,6 +201,8 @@ update_status ModuleRender::Update()
 		skybox->Draw();
 	}
 
+	gameObjectsToDraw.clear();
+
```
OriolSors, line 205: Do we need the cleaning in the Update instead of PreUpdate?

OriolSors (approved):
As we discussed during the last weeks, all seems to be agreed in the short term. I find more than good the approach of using the Quadtree to speed up the raycasting but taking into account the current selected object too. It is a functional and efficient solution but (maybe) not the cleanest one... We need to clarify and stablish all together what will the cases and limits of the Quadtree be, and then polish this (only if needed). But for now, perfect job all of you!!!

PD: If you can, answer my question commented. I don't know if this change was made during our discussion about using the RenderList instead of Quadtree GO, or if it is needed for some reason I miss right now. Thanks in advance!!

Source/DataStructures/Quadtree.cpp
markmeliaa: Thanks!

Source/Modules/ModuleRender.cpp
markmeliaa: Oh, it was my bad, fixed it!

OriolSors: Fine!! It was because of trying to raycast to the RenderList right?

markmeliaa: Yes! I was about to answer your other comment, it was left there when we tried to work with the RenderList, but forgot to changed it back later.

OriolSors: Perfect!!

markmeliaa:
> As we discussed during the last weeks, all seems to be agreed in the short term. I find more than good the approach of using the Quadtree to speed up the raycasting but taking into account the current selected object too. It is a functional and efficient solution but (maybe) not the cleanest one... We need to clarify and stablish all together what will the cases and limits of the Quadtree be, and then polish this (only if needed). But for now, perfect job all of you!!!

I agree that maybe a better solution could be possible, but yea, for now, this will do the work :)

markmeliaa:
Ready to be merged! @gerard-queralt 🫡

PR #153, opened 2023-02-20
style: adapt the code following programming rules

PaablooCH:


PaablooCH:
Task linked: [CU-865bph92u Coding style: correct cases, 120 column limit, etc.](https://app.clickup.com/t/865bph92u)

Source/DataModels/Resources/Resource.h
```diff
@@ -58,20 +61,23 @@ class Resource
...
+	return changed;
 }
 
 inline void Resource::SetChanged(bool changed)
 {
 	this->changed = changed;
 }
 
-inline Resource::Resource(UID resourceUID, const std::string& fileName, const std::string& assetsPath, const std::string& libraryPath):
+inline Resource::Resource(	UID resourceUID,
```
markmeliaa, line 77: I'd improve the indentation here, it looks wacky.

markmeliaa (approved):
Great Pablo, thanks for your work!

Source/DataModels/Resources/Resource.h
markmeliaa: Oh yea this is a constructor, you're right!

gerard-queralt (approved):
Thank you for your work, Pablo!

I'm not sure about removing all uses of `this`, as I think that it makes some methods easier to understand, but we can discuss this all together later and update the programming rules accordingly.

PR #154, opened 2023-02-20
Coding-style-correct-cases-120-column-limit-etc

PaablooCH:
revised all the files and adapted to programming rules


PaablooCH:
Task linked: [CU-865bph92u Coding style: correct cases, 120 column limit, etc.](https://app.clickup.com/t/865bph92u)

PR #155, opened 2023-02-20
Refactor camera

Dedado3000:


markmeliaa (approved):
Nice changes! You can merge this if you want, as it does not go directly to develop everything is fine.

Source/Modules/ModuleEditor.cpp
```diff
@@ -4,7 +4,7 @@
 #include "ModuleEditor.h"
 #include "ModuleWindow.h"
 #include "ModuleRender.h"
-#include "ModuleEngineCamera.h"
+#include "ModuleCamera.h"
```
UriMurcia, line 7: Is this include necessary?

UriMurcia (commented):
The camera in DebugGame works weird when is very close to the objects and we use the Right Button (Free look). I'm not sure if it happened before or not :(

Keep up the good work! :)

PaablooCH:
Task linked: [CU-865bt5vjd Refactor Camera to accept diferent types](https://app.clickup.com/t/865bt5vjd)

Source/Modules/ModuleCamera.cpp
```diff
@@ -0,0 +1,57 @@
+#pragma warning (disable: 26495)
+
+#include "Application.h"
```
DADSILVER, line 3: Review includes

DADSILVER (commented):
It a good refactor in general . I will point out a few things.

Source/DataModels/Camera/Camera.cpp
```diff
@@ -0,0 +1,347 @@
...
+	float width = windowScene->GetAvailableRegion().x;
+	float height = windowScene->GetAvailableRegion().y;
+
+	float normalizedX = -1.0f + 2.0f * mousePositionInScene.x / width;
+	float normalizedY = 1.0f - 2.0f * mousePositionInScene.y / height;
+
+	return frustum->UnProjectLineSegment(normalizedX, normalizedY);
+}
+
+void Camera::CalculateHittedGameObjects(const LineSegment& ray)
```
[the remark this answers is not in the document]
PaablooCH: this will be solved when the mouse picking's pull request will be accepted

Source/DataModels/Camera/Camera.h
```diff
@@ -133,18 +118,67 @@ class ModuleEngineCamera : public Module
...
+{
+	return this->type;
+}
+
+inline Frustum* Camera::GetFrustum()
+{
+	return this->frustum.get();
+}
+
+const std::string GetNameByCameraType(CameraType type)
```
PaablooCH, line 138: In a switch with returns inside of it, you dont need to use the break statement

Source/DataModels/Camera/Camera.h
```diff
@@ -133,18 +118,67 @@ class ModuleEngineCamera : public Module
...
+	case CameraType::C_GAMEOBJECT:
+		return "Camera_GameObject";
+		break;
+	default:
+		assert(false && "Wrong camera type introduced");
+		return "";
+	}
+}
+
+const CameraType GetCameraTypeByName(const std::string& typeName)
```
PaablooCH, line 157: Why this function and the previous one are in the .h?

Source/DataModels/Camera/Camera.h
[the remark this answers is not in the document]
PaablooCH: Well, the use of inline is just a hint to the compiler. It can choose to ignore it or, indeed, to inline functions not marked inline. So, probably if the compiler sees that make it inline is a good option it will do it.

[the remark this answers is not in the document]
gerard-queralt: I think this is more like a serializer for the enum

OriolSors (approved):
Very good refactor that provides a scalable solution for all type of cameras we want.
Thanks!!! @Dedado3000

PR #163, opened 2023-02-27
feat: docked windows on start-up

MarcGonzalezMoratona:


gerard-queralt (changes requested):
Thank you Marc, looks great! Unfortunately we cannot merge this as is, since the way it's currently set up we override the user's saved layout on each start-up, which we don't want.
I've been looking at the ImGui documentation to fix this in an idiomatical way using the flag `ImGuiCond_FirstUseEver`, which works for setting size or position, but I didn't manage to make it work with the method `SetWindowDock` which does take an `ImGuiCond` as an argument.
I also tried to check for the existance of the `imgui.ini` file using ModuleFileSystem, but it seems like it takes a few seconds to create it, and during that time the layout is completely messed up.
In my opinion the best solution would be to use the `ImGuiCond_FirstUseEver` flag, but if it's indeed not possible with docking, I would try combining both of our approaches: check if there's a previous configuration with ModuleFileSystem, but keep your `first_time` flag in order to avoid those seconds of the layout being broken. Something like:
```
if (first_time && !App->fileSystem->Exists("imgui.ini")) <- important to do it in this order, for lazy evaluation
```
This should make it so we use the user's saved layout if there is one, and otherwise set the default one.
Hopefully I explained myself properly!

gerard-queralt (changes requested):
Also, try to have variables in camelCase to follow programming style. Things like `first_time` should be `firstTime`. Sorry to be picky :P

gerard-queralt (dismissed):
Thanks Marc, everything looks great! One thing we can keep in mind is that, should we need to disable other windows by default like we did with Configuration, it might be a good idea to store that boolean in each class, but that's not needed for now.
Something I noticed when reviewing your branch is that we override the user's previous configuration regarding which windows are enabled and which are not, but I realize this issue already happens in `develop`, so it's not a new error. I'll create a different ticket in ClickUp and we'll tackle this from there ;)

AuxPort66 (approved):

gerard-queralt (approved):

PR #164, opened 2023-02-27
fix[Quadtree]: task #865bubdbk

Dedado3000:
Refactored Quadtree Remove and SmartRemove in order to fix some recursive problems

### DIFF
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -426,14 +426,11 @@ bool GameObject::RemoveComponent(const Component* component)
 
 Component* GameObject::GetComponent(ComponentType type) const
 {
-	if (!components.empty())
+	for (std::vector<std::unique_ptr<Component>>::const_iterator it = components.begin(); it != components.end(); ++it)
 	{
-		for (std::vector<std::unique_ptr<Component>>::const_iterator it = components.begin(); it != components.end(); ++it)
+		if ((*it)->GetType() == type)
 		{
-			if ((*it)->GetType() == type)
-			{
-				return (*it).get();
-			}
+			return (*it).get();
 		}
 	}
 	
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -426,14 +426,17 @@ bool GameObject::RemoveComponent(const Component* component)
 
 Component* GameObject::GetComponent(ComponentType type) const
 {
-	for (std::vector<std::unique_ptr<Component>>::const_iterator it = components.begin(); it != components.end(); ++it)
+	if (!components.empty())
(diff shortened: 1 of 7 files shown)


PaablooCH:
Task linked: [CU-865bubdbk Quadtree problems](https://app.clickup.com/t/865bubdbk)

Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -426,14 +426,17 @@ bool GameObject::RemoveComponent(const Component* component)
 
 Component* GameObject::GetComponent(ComponentType type) const
 {
-	for (std::vector<std::unique_ptr<Component>>::const_iterator it = components.begin(); it != components.end(); ++it)
+	if (!components.empty())
```
gerard-queralt, line 429: As I said in the ticket, this could not have been the cause of the bug, since iterator loops don't require a previous check, like all other loops. You can try it yourself: removing the `if` clause still allows for GameObject to be deleted from the hierarchy. I think it's better if we remove this `if`.

gerard-queralt (changes requested):

Source/DataModels/GameObject/GameObject.cpp
gerard-queralt: I've detected another bug while testing, but it's a different one from the one reported by the ticket. I'll create an issue.

Source/DataStructures/Quadtree.cpp
```diff
@@ -234,16 +205,20 @@ void Quadtree::OptimizeParentObjects()
...
 {
+	bool childrensDeleted = false;
 	std::list<const GameObject*> familyObjects = {};
 	GetFamilyObjects(familyObjects);
 	if (familyObjects.size() <= quadrantCapacity)
 	{
 		gameObjects.clear();
 		gameObjects.splice(gameObjects.end(), familyObjects);
+		//this causes a crash, need to find why
 		ResetChildren();
```
OriolSors, line 218: The splice method or the ResetChildren?

Dedado3000: Fuck, that was a comment of gerard that i forget to remove

OriolSors: Okey jajajajaja

Source/DataStructures/Quadtree.cpp
```diff
@@ -234,16 +205,20 @@ void Quadtree::OptimizeParentObjects()
 	}
```
OriolSors, line 205: We no longer need the OptimizeParentObjects() function.

OriolSors (changes requested):
If you agree we can delete the function I commented @Dedado3000

gerard-queralt (approved):
Everything looks good to me, resolve Oriol's request and we can merge.

PR #168, opened 2023-02-28
fix[WindowsHierarchy]: task #865buw83b

Dedado3000:


PaablooCH:
Task linked: [CU-865buw83b Bug: Destruction of a Father of selected GameObject](https://app.clickup.com/t/865buw83b)

markmeliaa:
Another better solution found.

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
```diff
@@ -112,7 +112,8 @@ void WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject)
         {
             if (ImGui::MenuItem("Delete"))
             {
-                if (gameObject == App->scene->GetSelectedGameObject())
```
gerard-queralt, line 115: Wouldn't it be more efficient to also leave this check, even though it's redundant? This way we would avoid traversing the vector when we are deleting the selected game object, I think

gerard-queralt:
Nevermind then :P

markmeliaa:
> Nevermind then :P

Sorry! We are working on it right now.

Issue #169, opened 2023-02-28
Problems deleting gameobjects with children


PaablooCH:
ClickUp task -> #865bubdex
Problem searching Camera Component inside a normal gameobject.

PaablooCH:
There are some problems deleting gameobjects:

- When we delete a father, in my case will be "Drawer", the father is deleted while its children still remaining inside sceneGameObjects
before:
![image](https://user-images.githubusercontent.com/62140600/222162922-3a997db5-8283-459f-833d-e626c8e22353.png)
after (drawer's child still inside):
![image](https://user-images.githubusercontent.com/62140600/222162706-da8ae953-21ef-4cbd-bb05-f87cc7f9ff84.png)

- If we delete a father while any of its children is selected, the selectedGameobject points to an empty space:
![image](https://user-images.githubusercontent.com/62140600/222165062-92bdcf15-28e5-4cba-9138-98644e98fd55.png)

- When a scene is loaded, the sceneGameObjects and sceneCameras vectors are filled with the same gameobjects, so if we delete correctly the father and its children from the first one, sceneCamera still has pointers pointing to an empty space.
![image](https://user-images.githubusercontent.com/62140600/222165925-88e2d8b6-2124-49cc-bd78-c8154007a25b.png)

Issue #167, opened 2023-02-28
Selected Game Object Destruction of his parent


Dedado3000:
When a gameObject is deleted in the hierarchy didn't check if the selected gameObject is a child of it

markmeliaa:
Do you mean that when you select a child and then delete its parent without selecting it, the engine breaks?

Dedado3000:
Yes, you can see the soluution that  @DADSILVER and me did at pull request https://github.com/Horizons-Games/Axolotl-Engine/pull/168

markmeliaa:
Nice, good job!

markmeliaa:
Now it is really fixed 🫡

Issue #166, opened 2023-02-28
GameObjects/Components deleted in the middle of the frame


gerard-queralt:
When deleting GameObjects (and possibly Components too) through the graphical interface, occasionally the program crashes during the call to `GameObject::Draw`. Upon inspecting the variables, it's pretty obvious that the code is running on "garbage" memory, which makes me think that the deletition is happening while the GameObject is drawing itself.
![image](https://user-images.githubusercontent.com/79362519/221723642-0f19fae6-fb99-48da-b478-158f6c964e7c.png)
Unfortunately, as far as I know, there's no way to detect whether memory is garbage or not (can't use `nullptr`). A solution that comes to mind is to push objects to delete into a queue/vector/list, and delete them in a `PostUpdate` call, after the frame is done.

gerard-queralt:
Found while reviewing #164

gerard-queralt:
Opened a ticket in ClickUp: #865buqzqp

sncarbo:
I suspect the problem is that `Scene::DestroyGameObject` deletes the `GameObject` from `sceneGameObjects` but not the children of the `GameObject`, so it tries to draw rubbish. I think that simply adding in the if-statement `*it == gameObject || gameObject->IsADescendant(*it)` would be enough.

sncarbo:
Good evening @gerard-queralt ! I have been doing some research about the issue. I have found that sceneGameObjects of the Scene class not only contains the imported models, but the meshes of this one are also treated as GameObjects and are children of the model. When a GameObject is deleted, its children are not deleted, so they are left with a garbage value. This makes me question two things:

1. Should the children of a model (meshes) be in sceneGameObjects?
2. Should they be deleted from sceneGameObjects when their parent is deleted? And if so, how?

I hope my comment was helpful.

sncarbo:
I think I solved it. I used the range element version of `std::vector::erase` so that I could remove the children contained in `sceneGameObjects` as well.

gerard-queralt:
[...]

PR #170, opened 2023-02-28
fix: problems deleting gameobjects with children 

PaablooCH:
When creating the scene all gameobjects go to vector sceneCamera provoking that when we search for objects inside camera we take those objects as cameras.

#172 must be approved in order to try this changes out correctly.

### DIFF
--- a/Source/DataModels/Scene/Scene.cpp
+++ b/Source/DataModels/Scene/Scene.cpp
@@ -104,18 +104,8 @@ GameObject* Scene::CreateCameraGameObject(const char* name, GameObject* parent)
 
 void Scene::DestroyGameObject(GameObject* gameObject)
 {
+	RemoveFatherAndChildren(gameObject);
 	gameObject->GetParent()->RemoveChild(gameObject);
-	RemoveCamera(gameObject);
-	for (std::vector<GameObject*>::const_iterator it = sceneGameObjects.begin();
-		it != sceneGameObjects.end();
-		++it)
-	{
-		if (*it == gameObject)
-		{
-			sceneGameObjects.erase(it);
-			return;
-		}
-	}
 }
 
 void Scene::ConvertModelIntoGameObject(const char* model)
@@ -176,7 +166,17 @@ GameObject* Scene::SearchGameObjectByID(UID gameObjectID) const
 	return nullptr;
 }
 
-void Scene::RemoveCamera(const GameObject* cameraGameObject)
+void Scene::RemoveFatherAndChildren(const GameObject* father)
+{
+	for (GameObject* child : father->GetChildren())
+	{
+		RemoveFatherAndChildren(child);
+	}
+	RemoveFromCamera(father);
+	RemoveFromScene(father);
+}
+
+void Scene::RemoveFromCamera(const GameObject* cameraGameObject)
 {
 	for (std::vector<GameObject*>::iterator it = sceneCameras.begin();
 		it != sceneCameras.end(); ++it)
@@ -184,6 +184,19 @@ void Scene::RemoveCamera(const GameObject* cameraGameObject)
 		if (cameraGameObject == *it)
 		{
 			sceneCameras.erase(it);
+			return;
+		}
+	}
+}
+
+void Scene::RemoveFromScene(const GameObject* gameObject)
+{
+	for (std::vector<GameObject*>::const_iterator it = sceneGameObjects.begin();
+		it != sceneGameObjects.end(); ++it)
+	{
+		if (*it == gameObject)
+		{
+			sceneGameObjects.erase(it);
 			return;
 		}
 	}
--- a/Source/DataModels/Scene/Scene.cpp
+++ b/Source/DataModels/Scene/Scene.cpp
@@ -172,29 +172,25 @@ void Scene::RemoveFatherAndChildren(const GameObject* father)
 	{
 		RemoveFatherAndChildren(child);
 	}
-	RemoveFromCamera(father);
-	RemoveFromScene(father);
-}
 
-void Scene::RemoveFromCamera(const GameObject* cameraGameObject)
-{
-	for (std::vector<GameObject*>::iterator it = sceneCameras.begin();
-		it != sceneCameras.end(); ++it)
+	Component* component = father->GetComponent(ComponentType::CAMERA);
+	if (component)
 	{
-		if (cameraGameObject == *it)
+		for (std::vector<GameObject*>::iterator it = sceneCameras.begin();
+			it != sceneCameras.end(); ++it)
 		{
-			sceneCameras.erase(it);
-			return;
+			if (father == *it)
+			{
+				sceneCameras.erase(it);
+				return;
+			}
 		}
 	}
-}
 
-void Scene::RemoveFromScene(const GameObject* gameObject)
-{
 	for (std::vector<GameObject*>::const_iterator it = sceneGameObjects.begin();
 		it != sceneGameObjects.end(); ++it)
 	{
-		if (*it == gameObject)
+		if (*it == father)
 		{
 			sceneGameObjects.erase(it);
 			return;
--- a/Source/DataModels/GameObject/GameObject.h
+++ b/Source/DataModels/GameObject/GameObject.h
@@ -34,10 +34,13 @@ public:
 	UID GetUID() const;
 	const char* GetName() const;
 	GameObject* GetParent() const;
+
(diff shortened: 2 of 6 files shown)


PaablooCH:
Task linked: [CU-865bubdex Problem with GetComponent](https://app.clickup.com/t/865bubdex)

gerard-queralt (dismissed):
Nice catch, this will certainly help improve the performance of the Engine. I have to say, though, that this does not solve the bug you found. The only reason it doesn't happen anymore is because there no longer are any cameras in the scene, so the loop where the crash occurs is never accessed. If you do a quick modification to `WindowInspector` to be able to add a Camera Component to an object, you can easily still trigger the bug: just add one camera to a parent and its child and then remove the parent. This looks like it's related to either #166 or #167, or both.

gerard-queralt:
This is the addition I did, in case you want to try it in you machine:
image
For easy copy-pase:
```
if (ImGui::MenuItem("REMOVE: Add camera"))
{
	App->scene->GetSelectedGameObject()->CreateComponent(ComponentType::CAMERA);
	auto debug = App->scene->GetLoadedScene()->GetSceneCameras();
	debug.push_back(App->scene->GetSelectedGameObject());
	App->scene->GetLoadedScene()->SetSceneCameras(debug);
}
```

gerard-queralt:
This has nothing to do with memory leaks, this is an issue of the timing of GameObjects and Components deletition, which ends up causing code to run on garbage memory. Also, please report whatever bug you found so we can all be up to date. And this MR is perfectly fine as is, you fixed an error with scene deserialization, so no need to close it.

PaablooCH:
I insist, the problem with the memory leak affects also this scope what you are talking (GameObjects and Components deletition) i dont want to push something that solves half of the problem, so this pr will be closed until i found a solution that could be solf the whole problem. Thanks you for your comprehension
P.S.: later on i will write about the memory leak problem

PaablooCH:
See issue #169 for what is solved

Source/DataModels/Scene/Scene.cpp
```diff
@@ -176,7 +166,17 @@ GameObject* Scene::SearchGameObjectByID(UID gameObjectID) const
...
 }
 
-void Scene::RemoveCamera(const GameObject* cameraGameObject)
+void Scene::RemoveFatherAndChildren(const GameObject* father)
+{
+	for (GameObject* child : father->GetChildren())
+	{
+		RemoveFatherAndChildren(child);
+	}
+	RemoveFromCamera(father);
```
markmeliaa, line 175: When a GameObject is deleted, you don't chech if it's a camera or not before removing it from cameras, why that?

markmeliaa (changes requested):
Tried deleting a parent when I have a child selected, and the engine still breaks:
image

It is trying to call the BoundingBox of a deleted GameObject:
image

I don't remember if this is what this pull request was trying to fix but I think that it enters in the scope of `problems deleting gameobjects with children` so I guess it should be addressed here too.

Also, when deleting a parent gameobjects, their children are still not correctly deleted, check out, after deleting anything from `mini_lights` scene:
image

And this is the result when I delete the wall (with Wall_0 and Wall_1 as children):
image

The children of Wall remain like this in the vector:
image

So the functions created in this pull request that remove the children from the parent are not working somehow...

I see the other issues addressed in #169 correctly resolved.

markmeliaa:
Wait, nevermind the first error that I reported, forgot that we fixed it in #172.

gerard-queralt:
If this were the cause of this issue it would crash everytime we delete something, as it stands it seems to crash randomly on develop. We certainly need to clear the various collections in the engine when deleting, but again, I seriously doubt this is what's causing that. Regardless, I think a better way of doing this would be to exploit the deleters of smart pointers. What you did here would need to be replicated in the Quadtree, and also do a similar thing for Components and Resources. Hopefully you see why I don't consider this the best approach. I talked a bit about this in #166

Source/DataModels/Scene/Scene.cpp
markmeliaa: There isn't a flag per se, but if a GameObject is a camera, it has a `CameraComponent` attached to it, so that is how you could differentiate them. Just search if the GameObject has a `ComponentCamera` before removing, if it does, delete that GameObject from the camera vector too, if it does not, do nothing.

PaablooCH: Good catch, but while i was implementing it i saw that if we delete the component Camera we no longer can track by this solution, because the camera will remain inside cameraScene but we can not ensure that the gameobject is a camera

markmeliaa (approved):

PaablooCH:
> > Good catch, but while i was implementing it i saw that if we delete the component Camera we no longer can track by this solution, because the camera will remain inside cameraScene but we can not ensure that the gameobject is a camera
> 
> Wait, `ComponentCamera` should not have the option to be removed, its `canBeRemoved` variable is set to `false`, so the Remove Comp. button should not appear in the ImGui.

Seems that something happens, because the button is available to delete the component
image

markmeliaa:
Okay, looks like when the component windows were decoupled from the Component::Draw() the `canBeRemoved` functionality was not taken into account, let me open an issue for this.

markmeliaa:
Done, addressed in #176

PaablooCH:
> If this were the cause of this issue it would crash everytime we delete something, as it stands it seems to crash randomly on develop. We certainly need to clear the various collections in the engine when deleting, but again, I seriously doubt this is what's causing that. Regardless, I think a better way of doing this would be to exploit the deleters of smart pointers. What you did here would need to be replicated in the Quadtree, and also do a similar thing for Components and Resources. Hopefully you see why I don't consider this the best approach. I talked a bit about this in #166

it's true that crash randomly but only in the last scene, if you load the first one it happens on every father that you want to delete (forget floor it crashes always but for different reasons). I don't see why using the deleters of smart pointers would improve the solutions, having in mind that the problem was that we not clear the vectors, what is the solution that you propose?

gerard-queralt:
Thank you for pointing out another way to force the crash, the more information we have the better. And sorry if I didn't explain myself properly, I'll try to lay out my idea. I'll try to keep it short.
The definition of a smart pointer like `std::unique_ptr<A>()` is actually hiding another template argument that the class has. If you hover over it, you will notice that it's setting an argument with type [std::default_delete\<A\>](https://en.cppreference.com/w/cpp/memory/default_delete). In the documentation, you can see that this is basically a class that calls `delete a` internally.
image
What I would like to do is to actually set the deleter that we want to use, and use a struct that we define. That is were we would call the removal from the collections. By doing it this way, our solution would be more modular - as I said, what you did here is perfectly correct, but would need to be replicated in Quadtree, and potentially other places, and the same thing will need to happen with other classes like Resource, which will remain in the map inside ModuleResources. Another advantatge is that we would de-couple the objects from the classes that contain them: a GameObject won't need to know in which Quadtrees it's present, or whether it's in ModuleRender's `renderList`, etc., which is related to the fact that it would be more modular.
To give you an idea of what I have in mind, this is a small prototype that I came up with a few weeks ago:
```
template<typename P>
class CollectionAwareDeleter
{
public:
	CollectionAwareDeleter(std::vector<P*>& collectionToManage) : collectionToManage(collectionToManage)
	{
	}
	~CollectionAwareDeleter() = default;

	void operator() (P* p)
	{
		collectionToManage.erase(std::remove(std::begin(collectionToManage), std::end(collectionToManage), p));
		delete p;
	}

private:
	std::vector<P*>& collectionToManage;
};
```
And this is how it could be used:
```
unique_ptr<A, CollectionAwareDeleter<A>> ptrA = unique_ptr<A, CollectionAwareDeleter<A>>(new A(), CollectionAwareDeleter<A>(rawPtrVector));
```
This sort of thing is already used in [ModuleScene](https://github.com/Horizons-Games/Axolotl-Engine/blob/develop/Source/Modules/ModuleWindow.h#L58) and [ModuleInput](https://github.com/Horizons-Games/Axolotl-Engine/blob/develop/Source/Modules/ModuleInput.h#L77), where we have unique pointers to SDL structs that need to be deleted in a particular way.
Hopefully I explained myself properly. Again, it's not that your solution is bad by any means, but hopefully you see why I think this would be easier to use in the long term.

markmeliaa (dismissed):

markmeliaa (approved):
Nice last change, good job! (Basta ya, estoy harto de esta pr...)

gerard-queralt (approved):

PR #174, opened 2023-03-01
feat [MousePicking]: Imguizmo implemented

MacXxs:
Functionalities added:
- Ability to modify transforms directly with the gizmo
- Being able to use snapping and change its values in the x, y, and z axis
- Changing between Local and World modes to apply transformations
- Added a viewManipulator on the top right corner of the scene to snap directly on the axis selected from the manipulator (relative to where the camera is pointing)


markmeliaa (changes requested):
I already told you but nice work with the ImGuizmo Miki!

The only "complain" that I have is that the ImGuizmo section is too long to fit in the screen (at least in mine), so we could remove the Mouse Coords of there if they are not really needed so everything fits alright even in small screens, what do you say?

image

Oh, and I just found a bug trying this branch out, it is the same bug that it was reported for the 'scenemanagement' branch in the ClickUp (https://app.clickup.com/t/865bujq80), it seems to happen when a new scene is loaded after changing how the matrices work in the `ComponentTransform` files, so maybe that bug should be fixed before merging this.

Source/Modules/ModuleEngineCamera.cpp
```diff
@@ -738,31 +738,42 @@ EFrustumMode ModuleEngineCamera::GetFrustumMode() const
 	return frustumMode;
 }
 
+Frustum* ModuleEngineCamera::GetFrustum()
+{
+	return &this->frustum;
```
PaablooCH, line 743: "this" is not necessary

Source/Modules/ModuleEngineCamera.cpp
```diff
@@ -738,31 +738,42 @@ EFrustumMode ModuleEngineCamera::GetFrustumMode() const
...
-	{
-		return false;
+
+			ray = frustum.UnProjectLineSegment(normalizedX, normalizedY);
+
+			return true;
+		}
+		else
+		{
+			return false;
```
PaablooCH, line 775: the else is not necessary, you can put the return outside and will execute

PaablooCH (changes requested):
Good job Miki, one thing apart the code, if you move something with the arrows or the "squares" while there is something behind, the mousepicking selects that object, losing the first

Source/Modules/ModuleEngineCamera.cpp
MacXxs: yeah true, I have the bad habit of using it, i'll remove it!

Source/Modules/ModuleEngineCamera.cpp
MacXxs: true!

PaablooCH (approved):
I see all good, nice work

gerard-queralt (approved):
Thank you Miki, looks awesome!

PaablooCH (approved):

markmeliaa (approved):
Nice Miki, good job!

PR #178, opened 2023-03-02
perf: ComponentBoundingBoxes Integration into GameObject

sncarbo:


PaablooCH (changes requested):
It crashes for me instantly when i load the last scene.
The problem seems to be in the initialization of the scene, where the directional amb ambient light are nulls
image
here is the call stack:
image

sncarbo:
@PaablooCH The issue is that in the parse from/to JSON there is an attribute associated to the component of a `GameObject` called `"type"`, and one of the types is `"Component_Bounding"`. As `ComponentBoundingBoxes` no longer exists, it gives an error.

I am going to check how to solve it.

PaablooCH:
> @PaablooCH The issue is that in the parse from/to JSON there is an attribute associated to the component of a `GameObject` called `"type"`, and one of the types is `"Component_Bounding"`. As `ComponentBoundingBoxes` no longer exists, it gives an error.
> 
> I am going to check how to solve it.

Okey, i see that you are modifying the jsons, I will look at them again later

sncarbo:
I think I solved it. I removed the "Component_Bounding" from the JSON code of each scene by hand. I also fixed the conflicts with the Quadtree.

I have tested and, in theory, there are no bugs or crashes related to this section. Still, I ask you to check it thoroughly.

gerard-queralt (changes requested):
Thanks Santi, great job! Hopefully you can address my small complain.
I noticed while reviewing this MR that bounding boxes are not actually drawn. It's not an issue of this branch, since it also happens in develop. We should look into why.

PaablooCH:
it's seems that the "draw bounding box" doesn't work, it didn't show the orange box
image

sncarbo:
I think I've fixed it. The bounding box was being drawn, but it displayed behind (because of the order in which the modules are in Application).

If you activate to draw the bounding box of the first GameObject of the hierarchy, nothing will be displayed because the bounding box of this one is `{0.0, 0.0, 0.0}, {0.0, 0.0, 0.0}`. The bounding box of the children are drawn perfectly. I had thought about programming not to draw the bounding box if the volume of this one is 0, but I thought that it is a calculation that would have to be performed at each iteration that it is drawn and that would affect efficiency.

Thank you @gerard-queralt and @PaablooCH for being helpful.

gerard-queralt:
Thank you for looking into this, Santi. It seems that it's still not drawing the bounding boxes properly. For one, it only draws the BB of the selected GameObject (which I don't think it's what we want). And it doesn't draw the BB of an empty GameObject containing other GameObjects, which I also don't think it's what we want. Not only that, but that BB should probably not be 0, it should be the BB encapsulating all the BB of it's children. But I might be wrong on that.

sncarbo:
I think it's working perfectly now! If a GameObject has activated the drawing of its bounding box, its children's bounding boxes are also drawn.

Dedado3000:
Gerard ask me to explain how boundingBoxes works.

First: The main Goal of the BB is for select in which Quadtrees you will allocate the object or for Raycast, at this moment the BB will be size 0 if it has no mesh and the size of the mesh if it contains one, but the size of the BB will no depend only of a mesh (for example, if you have a component light maybe you can have a little BB to make you selectable).

Second: Having this into a count, the BB from an object must not contains it childrens BB because you don't need that "extra" in which Quadtree you are in, BUT, I see a good implementation having a "HyerarchyDrawBoundingBox" that draws all the BB from your childrens

sncarbo:
> HyerarchyDrawBoundingBox

I know exactly how a Bounding Box works. The bounding box of a GameObject does not contain the one of its children, but they are drawn this way so that it is understood which is the bounding box of each one.

I recommend that you look at the code that I have just uploaded because that `GameObject::HyerarchyDrawBoundingBox` method that you have mentioned is literally what is done in the `WindowInspector`.

gerard-queralt:
Santi, I was just asking him to leave this here for posterity's sake, because I asked whether the BB of the parent should contain the bounding box of its children, and I wanted his answer to be recorded. You clearly know how BB work, no one questioned that. Sorry if David's comment came off the wrong way, that wasn't our intention.

gerard-queralt (changes requested):
I'm sorry to keep requesting changes, but there's a couple things I'm not sure of. First, when you un-mark the "Draw BB" checkbox for the parent, it keeps its children marked. The other thing is that it seems like it only draws the BB of the selected GameObject. Take a look when you can ;)

sncarbo:
> I'm sorry to keep requesting changes, but there's a couple things I'm not sure of. First, when you un-mark the "Draw BB" checkbox for the parent, it keeps its children marked. The other thing is that it seems like it only draws the BB of the selected GameObject. Take a look when you can ;)

No problem! I'll look at it right away. I really appreciate you looking to see if there are any issue or not.

gerard-queralt:
Sorry, it seems like if you enable the drawing of a child's BB and then select it's father it disables the child you had selected. If you select a sibiling and then the parent it seems to work, so it's probably something very minor.

sncarbo:
Please, @gerard-queralt and @PaablooCH, check that everything is fine. As you both requested changes in the pull request, approval from both of you is required.

gerard-queralt (approved):
Thank you for your work and hearing out all my requests Santi, appreciate it :)

Source/DataModels/Components/ComponentMeshRenderer.cpp
```diff
@@ -172,8 +169,6 @@ void ComponentMeshRenderer::SetMesh(const std::shared_ptr<ResourceMesh>& newMesh
 	if (IsMeshLoaded())
 	{
 		mesh->Load();
-		ComponentBoundingBoxes* boundingBox =
-			static_cast<ComponentBoundingBoxes*>(GetOwner()->GetComponent(ComponentType::BOUNDINGBOX));
-		boundingBox->Encapsule(mesh->GetVertices().data(), mesh->GetNumVertices());
+		this->GetOwner()->Encapsule(mesh->GetVertices().data(), mesh->GetNumVertices());
```
PaablooCH, line 172: this `this` is redundant

Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -554,3 +565,17 @@ void GameObject::MoveDownChild(GameObject* childToMove)
 		}
 	}
 }
+
+void GameObject::CalculateBoundingBoxes()
+{
+	ComponentTransform* transform =
+		static_cast<ComponentTransform*>((this)->GetComponent(ComponentType::TRANSFORM));
```
PaablooCH, line 572: `this` redundant

PaablooCH (changes requested):
now prints all fine, but just a few programming rules before merging pls

PaablooCH (approved):
Approved 👍

gerard-queralt (approved):

PR #183, opened 2023-03-03
Light Intensity cap in order create over bright lights

sncarbo:


AlexisDieu (approved):

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentDirLight.cpp
```diff
@@ -36,8 +36,11 @@ void WindowComponentDirLight::DrawWindowContents()
 			ImGui::SetNextItemWidth(80.0f);
 			ImGui::PushStyleVar(ImGuiStyleVar_FramePadding, ImVec2(5.0f, 1.0f));
 			float intensity = asDirLight->GetIntensity();
-			if (ImGui::DragFloat("##Intensity", &intensity, 0.01f, 0.0f, 1.0f))
+			if (ImGui::DragFloat("##Intensity", &intensity, 0.01f, 0.0f, 8.0f))
 			{
+				if (intensity > 8.0f)
+					intensity = 8.0f;
+
```
OriolSors, line 43: Good! Without this conditional we could modify the value by typing!

OriolSors (approved):
Fine change Santi! Thanks!

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentDirLight.cpp
```diff
@@ -36,8 +36,11 @@ void WindowComponentDirLight::DrawWindowContents()
 			ImGui::SetNextItemWidth(80.0f);
 			ImGui::PushStyleVar(ImGuiStyleVar_FramePadding, ImVec2(5.0f, 1.0f));
 			float intensity = asDirLight->GetIntensity();
-			if (ImGui::DragFloat("##Intensity", &intensity, 0.01f, 0.0f, 1.0f))
+			if (ImGui::DragFloat("##Intensity", &intensity, 0.01f, 0.0f, 8.0f))
 			{
+				if (intensity > 8.0f)
```
PaablooCH, line 41: All the ifs must be with {}, following the programming rules

PaablooCH (changes requested):
Nice catch, especially with the ifs as oriol says, just one thing:

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentDirLight.cpp
sncarbo: Exactly! `DragFloat` does not allow to set a minimum and maximum limit if you enter the text by hand, which I think they should change!

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentDirLight.cpp
sncarbo: Absolutely right! I'll change it immediately. I also take this as an opportunity to set the limit to zero and avoid negative numbers in the light intensity.

PaablooCH: Perf

PaablooCH (approved):

sncarbo:
I applied the requested changes! Thanks @OriolSors and @PaablooCH for being helpful!

OriolSors (commented):
Great!

OriolSors (dismissed):

MacXxs:
Hi Santi, everything looks fine, but i wouldn't put such a low cap on the light intensity. For example unity has no cap (it can be infinit), and it actually makes sense 'cause only close objects to the light will look burnt on higher intensities, but if the light is enough far away, even at a 100 intenisity the scene can look alright, so actually i would move to cap from 8 to 500 per example, or even 5000!

On top of that, maybe we should add this top cap as a #default value in the ComponentLight class, so we can change it from there instead having to change it manually from every function on the windowComponent classes.

image

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentPointLight.cpp
```diff
@@ -78,8 +78,17 @@ void WindowComponentPointLight::DrawWindowContents()
 			ImGui::SetNextItemWidth(80.0f);
 			ImGui::PushStyleVar(ImGuiStyleVar_FramePadding, ImVec2(5.0f, 1.0f));
 			float intensity = asPointLight->GetIntensity();
-			if (ImGui::DragFloat("##Intensity", &intensity, 0.01f, 0.0f, 1.0f))
+			if (ImGui::DragFloat("##Intensity", &intensity, 0.01f, 0.0f, 8.0f))
```
MacXxs, line 81: Hi Santi, everything looks fine, but i wouldn't put such a low cap on the light intensity. For example unity has no cap (it can be infinit), and it actually makes sense 'cause only close objects to the light will look burnt on higher intensities, but if the light is enough far away, even at a 100 intenisity the scene can look alright, so actually i would move to cap from 8 to 500 per example, or even 5000!

On top of that, maybe we should add this top cap as a #default value in the ComponentLight class, so we can change it from there instead having to change it manually from every function on the windowComponent classes.

image

MacXxs (changes requested):

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentPointLight.cpp
sncarbo: Perfect! I'm going to change it. I will add a `protected` `const static float` in `WindowComponentLight` so that its children can access it.

sncarbo: I just saw that there is no inheritance between WindowComponentPointLight, WindowComponentDirLight and WindowComponentSpotLight with WindowComponentLight, so I will do it another way, although I think that should have inheritance.

MacXxs: Hi, adding the define on the ComponentLight class should be enough. Like this:

image

MacXxs: Then you should be able to call it normally on the WindowComponent for each light, 'cause it's a value stablished on the parent of the lights, and each of its child classes are imported on the window.

image

sncarbo: I am not a big fan of using macros for this particular case because a macro does not respect any kind of scope. Also, macros are very dependent on the compiler configuration. It is fine to use them if they are macros used throughout the system, but some members, as in this case, must be encapsulated.

MacXxs: Okey, if it's better, for me it's okey adding inheritance, as long as adding the extra includes isn't worse than simply using the macro, it's fine by me!

MacXxs (dismissed):

PaablooCH (dismissed):

sncarbo:
I made one last change to replace 'const' with 'constexpr' to avoid reserving memory. @PaablooCH @MacXxs @OriolSors @AlexisDieu

MacXxs (approved):

PR #186, opened 2023-03-05
Checked all methods returning references to local objects

xavi1111:
Checked all methods returning references to local objects and changed the ones that did. The only ones found are the modified ones.


gerard-queralt (approved):
Thank you for your work Xavi!

AuxPort66 (approved):
👍

PR #187, opened 2023-03-06
feat: FileSystem rework

AuxPort66:
- Resources are stored as weak pointers in the Module's map and returned as shared pointers
- Components now decide the livespan of resource (Shared pointer)
- Engine Mode added to Loads
- Texture import/load options
- Load options and UID from meta if pre-exist
- Resource Window
- Inspector now have the capability of inspect resources and edit their load/import options
- Filesystem cache now only store resources when are requested from scene


Source/DataModels/Resources/Resource.h
```diff
@@ -48,6 +50,7 @@ class Resource
 	virtual void InternalLoad() = 0;
 	virtual void InternalUnload() = 0;
 
+	virtual bool ChildChanged() const;
```
gerard-queralt, line 53: We could probably move this stuff into the EditorResource, I don't think this will be needed outside Engine mode, will it? But probably fine to leave it here for now.

AuxPort66: This should go into EditorResource. Like assetPath, fileName, and changed. (And probably libpath too?)

PR #191, opened 2023-03-06
fix [lights]: "fixed" spotlight angle

MacXxs:
Clickup task: https://app.clickup.com/t/865bk5axr

Capped the maximum angle for the spotlights to 90 degrees. There wasn't actually anything to fix in this case, when the angle for the spotlights was greater than 90 degrees (1.57 radians), the effect that seemed to bug out the light was actually the light bending into itself (?) 'cause the angle was greater than 90 degrees, causing the light to shadow itself (like the effect of a black whole curving light around it in its event horizon).

Other than capping the light, i've changed the imgui to show the values in degrees instead of radians, so it's more clear the angle that is being used for the light.

### DIFF
--- a/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentSpotLight.cpp
+++ b/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentSpotLight.cpp
@@ -1,7 +1,9 @@
 #include "WindowComponentSpotLight.h"
 
 #include "Application.h"
+
 #include "ModuleScene.h"
+
 #include "DataModels/Scene/Scene.h"
 
 #include "DataModels/Components/ComponentSpotLight.h"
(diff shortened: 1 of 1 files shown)


PaablooCH (changes requested):
i don't know if this is a problem or is how it works but when a put an angle with 90º it lights outside of the cone
image
like i said, i don't know if it's the draw that is wrong, the light or is the correct behaviour

MacXxs:
> i don't know if this is a problem or is how it works but when a put an angle with 90º it lights outside of the cone image like i said, i don't know if it's the draw that is wrong, the light or is the correct behaviour

This already happened in the second assigment, it isn't a problem of the light itself 'cause the light it's at a 90 degrees, it's a problem with the debug draw and at the moment i haven't found a solution

PaablooCH (approved):
okey approved 👍

markmeliaa (approved):
Nice change Miki, well done!

PR #197, opened 2023-03-11
fix: Guizmo transformations not being applied since last changes in develop

MacXxs:
Solved the problem where transformations where not propertly being applied when modifying the objects with the guizmo


PaablooCH (approved):
nice 👍

PR #201, opened 2023-03-12
perf: efficient rendering program corrections

DADSILVER:
Modified how the rendering program is managed so that it works as the rest.

### DIFF
--- a/Source/DataModels/Program/Program.cpp
+++ b/Source/DataModels/Program/Program.cpp
@@ -2,8 +2,12 @@
 
 #include "Globals.h"
 #include "GL/glew.h"
+#include <MathGeoLib/Include/Math/float3.h>
 
-Program::Program(unsigned vertexShader, unsigned fragmentShader) : id (glCreateProgram())
+Program::Program(unsigned vertexShader, unsigned fragmentShader,
+	const std::string& vtxShaderFileName, const std::string& frgShaderFileName, const std::string& programName) :
+	vertexShaderFileName(vtxShaderFileName), fragmentShaderFileName(frgShaderFileName), programName(programName),
+	id(glCreateProgram())
 {
 	glAttachShader(id, vertexShader);
 	glAttachShader(id, fragmentShader);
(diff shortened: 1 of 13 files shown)


PaablooCH (approved):
works fine!!

Source/Modules/ModuleProgram.h
```diff
@@ -19,31 +20,32 @@ class ModuleProgram : public Module
 	~ModuleProgram() override;
 
 	bool Start() override;
-
 	bool CleanUp() override;
 
-	void CreateProgram(unsigned int vtxShader, unsigned int frgShader);
+	void UpdateProgram(std::string& vtxShaderFileName, std::string& frgShaderFileName, int programType,
```
gerard-queralt, line 25: This parameters can probably be const references

Source/Modules/ModuleProgram.h
```diff
@@ -19,31 +20,32 @@ class ModuleProgram : public Module
...
 {
-	return programs[static_cast<int>(type)].get();
-}
-
-inline const unsigned ModuleProgram::GetProgram() const
-{
-	return program;
+	if (programs.size() != 0) 
+	{
+		return programs[static_cast<int>(type)].get();
```
gerard-queralt, line 45: Should probably check the recieved type is not PROGRAM_TYPE_SIZE

Source/DataModels/Program/Program.cpp
```diff
@@ -29,14 +33,15 @@ Program::Program(unsigned vertexShader, unsigned fragmentShader) : id (glCreateP
...
+	CleanUp();
 }
 
 void Program::Activate()
 {
 	glUseProgram(id);
 }
 
-void Program::Deactivate()
+void Program::Desactivate()
```
gerard-queralt, line 44: "Desactivate" is not a word, should be "Deactivate". Sorry to be picky :P

gerard-queralt (changes requested):
Commented a few small things, but otherwise great changes!

Source/DataModels/Program/Program.cpp
DADSILVER: Oky

gerard-queralt (approved):
Thank you David!

PaablooCH (approved):

PR #202, opened 2023-03-13
feat: Undo-Redo

DGMC117:
This initial implementation of the functionality has a new module (ModuleCommand) and the implementation of the undo command for DrawEnableComponent.


Source/DataModels/Commands/CommandComponentEnabled.h
```diff
@@ -0,0 +1,39 @@
...
+	Component* comp;
+	bool* enable;
+	bool desiredValue;
+};
+
+CommandComponentEnabled::CommandComponentEnabled(Component* comp, bool* enable, bool desiredValue) : comp(comp), enable(enable), desiredValue(desiredValue)
+{}
+
+CommandComponentEnabled::~CommandComponentEnabled()
+{}
```
markmeliaa, line 24: If, in the future, `ComandComponentEnabled` requires a .cpp, this constructor and destructor should be moved to the .cpp, as that is what the programming rules say.

Source/Modules/ModuleCommand.h
```diff
@@ -0,0 +1,43 @@
...
+	void Undo();
+	void Redo();
+private:
+	std::list<std::unique_ptr<Command>> commandList;
+	std::list<std::unique_ptr<Command>>::iterator commandListIterator;
+	const static int commandLimit = 15;
+};
+
+template<typename C, typename ...Args, std::enable_if_t<std::is_base_of<Command, C>::value, bool>>
+void ModuleCommand::CreateAndExecuteCommand(Args&& ...args)
```
markmeliaa, line 31: Shouldn't this function be in the .cpp instead of in the .h?

markmeliaa (changes requested):
First of all, this should be updated with develop asap, as there are some things that have changed (and with this I mean bring develop to this branch, not just fixing the conflicts that this pr has).

Everything else is correct, I've checked the functionality and works as intended when enabling and disabling components, great job!

Source/Engine.vcxproj.user
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="Current" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <ShowAllFiles>false</ShowAllFiles>
+    <ShowAllFiles>true</ShowAllFiles>
```
PaablooCH, line 4: before merge change that variable to false again, you only have to push but deactivating 
this -> image

PaablooCH (changes requested):
I dont find any problem, just one thing:

Source/Modules/ModuleCommand.h
gerard-queralt: Template functions can't be in the .cpp file, unfortunately.

Source/Engine.vcxproj.user
gerard-queralt: This file should be in the `.gitignore`, not sure why it was tracked anyway.

Source/DataModels/Commands/CommandComponentEnabled.h
```diff
@@ -0,0 +1,39 @@
...
+public:
+	CommandComponentEnabled(Component* comp, bool* enable, bool desiredValue);
+	virtual ~CommandComponentEnabled() override;
+
+	std::unique_ptr<Command> Execute() override;
+
+private:
+	Component* comp;
+	bool* enable;
+	bool desiredValue;
```
gerard-queralt, line 17: What's the difference between `enable` and `desiredValue`, and why is the first one a pointer to a boolean?

Source/DataModels/Commands/CommandComponentEnabled.h
```diff
@@ -0,0 +1,39 @@
...
+};
+
+CommandComponentEnabled::CommandComponentEnabled(Component* comp, bool* enable, bool desiredValue) : comp(comp), enable(enable), desiredValue(desiredValue)
+{}
+
+CommandComponentEnabled::~CommandComponentEnabled()
+{}
+
+std::unique_ptr<Command> CommandComponentEnabled::Execute()
```
gerard-queralt, line 26: Wouldn't it be easier to do
```
if (...)
else ...
std::unique_ptr<CommandComponentEnabled> reverse = std::make_unique<CommandComponentEnabled>(comp, enable, !desiredValue);
return reverse;
```

gerard-queralt (commented):
Left a couple questions, but overall great job! This will be a great addition to the engine :)

Source/Modules/ModuleCommand.h
markmeliaa: rip, alright then.

PaablooCH (approved):

Source/DataModels/Commands/CommandComponentEnabled.h
```diff
@@ -0,0 +1,46 @@
...
+	bool desiredValue;
+};
+
+CommandComponentEnabled::CommandComponentEnabled(Component* comp, bool* enable, bool desiredValue) : comp(comp), 
+						enable(enable), desiredValue(desiredValue)
+{}
+
+CommandComponentEnabled::~CommandComponentEnabled()
+{
+	delete comp;
```
gerard-queralt, line 26: This will cause a double deletition of the pointer, since this Component is managed by a unique pointer somewhere else. I could be wrong, but I'm pretty sure this will cause an error when running it.

PaablooCH (approved):

PaablooCH (approved):

Source/DataModels/Commands/CommandComponentEnabled.h
DGMC117: This has been fixed. After discussing we saw that the pointer was incorrect.

markmeliaa (approved):

PR #205, opened 2023-03-14
feat: gameplay and static cameras

Dedado3000:
This is the first aproximation to and static camera, it contains a refactor of cameraComponent to contain the new camera implemented in the last milestone


Dedado3000:
I Will close this pull request bc i will pull all the implementation (Camera and player)

PR #206, opened 2023-03-15
feat: Game Mode allows debug options

gerard-queralt:
Triggered by `Alt+J`. Currently displays light debug stuff, game objects bounding boxes and an ImGui window that does nothing but can be used for showing FPS, disabling/enabling certain options and commands like teleport down the line.

### DIFF
--- a/Source/Modules/ModuleEditor.cpp
+++ b/Source/Modules/ModuleEditor.cpp
@@ -5,6 +5,7 @@
 #include "ModuleRender.h"
 #include "FileSystem/ModuleFileSystem.h"
 
+#ifdef ENGINE
 #include "Windows/WindowMainMenu.h"
 #include "Windows/EditorWindows/WindowConsole.h"
 #include "Windows/EditorWindows/WindowScene.h"
@@ -13,6 +14,9 @@
 #include "Windows/EditorWindows/WindowHierarchy.h"
 #include "Windows/EditorWindows/WindowEditorControl.h"
 #include "Windows/EditorWindows/WindowResources.h"
+#else
+#include "Windows/WindowDebug.h"
+#endif
 
 #ifdef DEBUG
 #include "optick.h"
@@ -31,7 +35,7 @@ static bool consoleOpened = true;
 static bool aboutOpened = false;
 static bool propertiesOpened = true;
 
-ModuleEditor::ModuleEditor() : mainMenu(nullptr), scene(nullptr), windowResized(false)
+ModuleEditor::ModuleEditor() : scene(nullptr), windowResized(false)
 {
 }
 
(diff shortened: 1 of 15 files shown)


markmeliaa:
I don't know if it's just me, but this branch does not compile in my machine :(

image

markmeliaa (commented):
I don't know if it's just me, but this branch does not compile in my machine.

image

markmeliaa:
Wait, brain fart lol, forgot to change into Game mode

Source/Modules/ModuleEditor.cpp
```diff
@@ -6,13 +6,18 @@
...
+#ifdef ENGINE
 #include "Windows/EditorWindows/WindowConsole.h"
 #include "Windows/EditorWindows/WindowScene.h"
 #include "Windows/EditorWindows/WindowConfiguration.h"
 #include "Windows/EditorWindows/WindowInspector.h"
 #include "Windows/EditorWindows/WindowHierarchy.h"
 #include "Windows/EditorWindows/WindowEditorControl.h"
 #include "Windows/EditorWindows/WindowResources.h"
+#else
+#include "Windows/EditorWindows/EditorWindow.h"
```
markmeliaa, line 18: So now, instead of not compiling the `ModuleEditor` at all during GAME, we will but without the non-necessary windows, is that correct?

Source/Modules/ModuleInput.cpp
```diff
@@ -170,6 +170,11 @@ update_status ModuleInput::Update()
 
     }
 
+    if (keysState[SDL_SCANCODE_LALT] == KeyState::REPEAT && keysState[SDL_SCANCODE_J] == KeyState::DOWN)
```
markmeliaa, line 173: Idk why if I maintain pressed `ALT+J` this keeps executing, the `KeyState::DOWN` is supposed to only trigger the first frame when a certain key is pressed if I am not mistaken.

markmeliaa (commented):
Great addition, this really gives value to our GAME part.

markmeliaa (approved):

gerard-queralt:
No Mark, you were actually right, I forgot to test it in engine so it was missing the include. Thank you for noticing!

Source/Modules/ModuleEditor.cpp
gerard-queralt: Correct, thought this includes will likely grow as the debug window does. I don't love it but I tried to keep it as light as possible while making the minimum changes :/

Source/Modules/ModuleInput.cpp
gerard-queralt: You are right, I'll create an issue to fix this in ModuleInput. Nice catch!

markmeliaa: Weird because when we developed the raycast the `KeyState::DOWN` worked as intended, maybe it has anything to do with the GAME mode? idk

markmeliaa (approved):

AuxPort66 (approved):

PR #209, opened 2023-03-17
feat: Create new scene

MarcGonzalezMoratona:


Piterga12 (commented):
When we are in a new scene and we click on save scene the save as window should pop up, as we havent saved it yet for the first time.

PR #210, opened 2023-03-18
fix: Quadtree height expanding bug

OriolSors:
I solved the crash found by @PaablooCH when loading the Corridor proper fbx.

ClickUp task: https://app.clickup.com/t/865bz5xyr

- The problem was caused by an error when expanding the height of the Quadtree: the value was not modified properly and then it didn't pass the EntireInQuadrant requirement, so it tries to expand also in X and Z axis. This gives a nullptr because the quadtree doesn't need to expand in these axis so it crashes.
- I modified a bit the new height calculation, making it more adjusted to the new object added. This is because we saw that high quadtree heights caused FPS troubles.


PaablooCH (approved):
Works fine 👍

MacXxs (approved):
Alright :thumbsup:

PaablooCH (approved):

PR #211, opened 2023-03-18
feat: Create resource

AuxPort66:
- Now we can right click on the hierarchy and directly create a gameobject with a default material and a couple of default meshes 
- Added a right click in the hierarchy to quickly create gameobjects with light components
- Added button in the lower filebrowser to create new materials

image


PaablooCH (approved):
Now i can create as much david as i want 12/10. One thing that i notice is that in FileBrowser Window at the bottom appears the ability to save things and i think that should not be there, no?

AuxPort66:
> The search bar in the file browser comes from develop, because whoever programmed the input file browser edited the general file browser. So yes, shouldn't be there.

markmeliaa (approved):
Nice addition, I run it a couple of times and everything runs correctly gj!

markmeliaa:
> The search bar in the file browser comes from develop, because whoever programmed the input file browser edited the general file browser. So yes, shouldn't be there.

So should we delete it in a separate pr or what?

AuxPort66:
>We need this for the blockout and we need to start the blockout this week, so yes, in a separate PR because it's not a priority right now. Maybe in a UI rework PR? I will discuss it with @DGMC117

PR #213, opened 2023-03-19
fix: Camera Focus movement and rotation not aligned

OriolSors:
Now that the Camera refactor is finally done, I solved this bug which appeared during the Concept Discovery.

- The problem was that the Focus stops at the frame when the final rotation is achieved. This has been aligned also with position and its Lerp.
- I modified the Slerp algorithm for not needing the current rotation and only use the current front direction.
- The speed has been adjusted too.

### DIFF
--- a/Source/DataModels/Camera/Camera.h
+++ b/Source/DataModels/Camera/Camera.h
@@ -73,7 +73,7 @@ public:
 	void SetPlaneDistance(float zNear, float zFar);
 	void SetPosition(const float3& position);
 	void SetOrientation(const float3& orientation);
-	void SetLookAt(const float3& lookAt);
+	void SetLookAt(const float3& lookAt, bool& isSameRotation);
 	void SetMoveSpeed(float speed);
 	void SetRotationSpeed(float speed);
 	void SetFrustumOffset(float offset);
--- a/Source/DataModels/Camera/Camera.h
+++ b/Source/DataModels/Camera/Camera.h
@@ -110,7 +110,6 @@ protected:
 
 	float4x4 projectionMatrix;
 	float4x4 viewMatrix;
-	Quat currentRotation;
 	float aspectRatio;
 	float acceleration;
 	float moveSpeed;
--- a/Source/DataModels/Camera/Camera.h
+++ b/Source/DataModels/Camera/Camera.h
@@ -110,7 +110,7 @@ protected:
 
 	float4x4 projectionMatrix;
 	float4x4 viewMatrix;
-	Quat currentRotation = Quat::identity;
+	Quat currentRotation;
 	float aspectRatio;
 	float acceleration;
 	float moveSpeed;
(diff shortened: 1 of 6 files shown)


AlexisDieu (dismissed):
The focus is working correctly now good job ! the speed also look more interesting like that and i don't find any trouble !

Source/DataModels/Camera/Camera.h
```diff
@@ -73,7 +73,7 @@ class Camera
 	void SetPlaneDistance(float zNear, float zFar);
 	void SetPosition(const float3& position);
 	void SetOrientation(const float3& orientation);
-	void SetLookAt(const float3& lookAt);
+	void SetLookAt(const float3& lookAt, bool& isSameRotation);
```
PaablooCH, line 76: like the float is not necessary to pass the bool by reference

Source/DataModels/Camera/CameraEngine.cpp
```diff
@@ -247,13 +247,23 @@ void CameraEngine::Focus(const OBB& obb)
...
+	else 
+	{ 
+		position = position.Lerp(endposition, App->GetDeltaTime() * rotationSpeed * 2);
+		SetPosition(position);
+	}
+
+	bool isSameRotation = false;
+	SetLookAt(boundingSphere.pos, isSameRotation);
+
+	if (isSamePosition && isSameRotation) isFocusing = false;
```
PaablooCH, line 266: following the programming rules that must be between {}

PaablooCH (changes requested):
Very smooth, i found just one thing:
if you rotate focusing an object sometimes the camera does a jump and rotate like 180º
camera-rotation

AlexisDieu:
> Very smooth, i found just one thing:
> if you rotate focusing an object sometimes the camera does a jump and rotate like 180º
> camera-rotation

For this issue it’s maybe more an issue with the mouse wrap so I am gonna look at it because this bug is kinda rare tbh I tried to re try it and it never happened again

OriolSors:
> Very smooth, i found just one thing: if you rotate focusing an object sometimes the camera does a jump and rotate like 180º camera-rotation [ camera-rotation ](https://user-images.githubusercontent.com/62140600/226199674-9a374433-5a33-4fb9-9d80-82b5b7c44221.gif) [ ](https://user-images.githubusercontent.com/62140600/226199674-9a374433-5a33-4fb9-9d80-82b5b7c44221.gif)

As @AlexisDieu has said, this is due to the Mouse Warp and it's out of this scope

Source/DataModels/Camera/Camera.h
OriolSors: I pass by reference because I modify the variable inside, not for efficiency

Source/DataModels/Camera/CameraEngine.cpp
OriolSors: Done

PaablooCH:
> As @AlexisDieu has said, this is due to the Mouse Warp and it's out of this scope

Okey 👍, so all good

PaablooCH (approved):

MacXxs (approved):
Everything looks fine!

PR #215, opened 2023-03-19
feat: added state to windows through meta's files

elbuho12Jesus:
The metafiles were created to read when the engine starts with the previous window states and to write when the engine closes the last window states.

### DIFF
--- a/Source/DataModels/Resources/Resource.h
+++ b/Source/DataModels/Resources/Resource.h
@@ -13,6 +13,7 @@ enum class ResourceType
 	Scene,
 	Material,
 	SkyBox,
+	Window,
 };
 
 class Resource
--- a/Source/DataModels/Windows/EditorWindows/EditorWindow.cpp
+++ b/Source/DataModels/Windows/EditorWindows/EditorWindow.cpp
@@ -1,31 +1,102 @@
 #include "EditorWindow.h"
 
+#include "physfs.h"
+#include "Application.h"
+#include "../FileSystem/ModuleFileSystem.h"
+#include "../FileSystem/UniqueID.h"
+#include "../FileSystem/Json.h"
+
+
+
+
+
 EditorWindow::EditorWindow(const std::string& name) : Window(name), flags(ImGuiWindowFlags_None),
 	focused(false)
-{
-}
+{}
 
 EditorWindow::~EditorWindow()
-{
+{	
 }
 
-void EditorWindow::Draw(bool& enabled)
+void EditorWindow::Start()
 {
+	load_meta(name.c_str(), enabled);
+}
+
+void EditorWindow::load_meta(std::string name, bool& enable)
+{
+	char* binaryBuffer = {};
+	std::string lib = "Lib/StateWindow/" + name + ".meta";	
+	
+	if (App->fileSystem->Load(lib.c_str(), binaryBuffer) > 0) {
+
+
+		rapidjson::Document doc;
+		Json json(doc, doc);
+		json.fromBuffer(binaryBuffer);
+		std::string buffer = std::string(binaryBuffer);
+		size_t i = buffer.find("State");
+		size_t j = buffer.find("\n", i);
+		if (i != std::string::npos)
+		{
+			if (buffer.substr(i + 8, j - (i + 8)) == "true")
+			{
+				enable = true;
+			}
+			else
+			{
+				enable = false;
+			}
+
+		}
+	}
+	else {
+		enable = true;
+	}	
+	delete binaryBuffer;
+
+}
+
+void EditorWindow::Update_meta(bool enabled,std::string name)
+{
+	
+		rapidjson::Document doc;
+		Json json(doc, doc);
+		json["UID"] = UniqueID::GenerateUID();
+		json["Type"] = "Window";
+		json["State"] = enabled;
+		rapidjson::StringBuffer buffer;
+		json.toBuffer(buffer);
+		std::string lib = "Lib/StateWindow/" + name + ".meta";				
+		App->fileSystem->Save(lib.c_str(), buffer.GetString(), (unsigned int)buffer.GetSize());	
+}
+
+
+
+void EditorWindow::Draw()
+{	
 	if (enabled)
 	{
 		if (ImGui::Begin(name.c_str(), &enabled, flags))
-		{
+		{			
+
 			ImGui::SetWindowSize(name.c_str(), GetStartingSize(), ImGuiCond_Once);
 
 			DrawWindowContents();
-			focused = ImGui::IsWindowFocused();
+			focused = ImGui::IsWindowFocused();			
 		}
 		ImGui::End();
-	}
+	}		
 	//The call to ImGui::Begin can change the value of io_enabled
 	//so using "else" will cause the window to remain focused for an extra frame
 	if (!enabled)
 	{
 		focused = false;
 	}
+	
+}
+
+void EditorWindow::CleanUp() 
+{
+	Update_meta(enabled, name);
 }
--- a/Source/DataModels/Windows/EditorWindows/EditorWindow.cpp
+++ b/Source/DataModels/Windows/EditorWindows/EditorWindow.cpp
@@ -18,47 +18,7 @@ EditorWindow::~EditorWindow()
 {	
 }
 
-void EditorWindow::Start()
-{
-	LoadMeta(name.c_str(), enabled);
-}
-
-void EditorWindow::LoadMeta(const std::string name,bool& enable)
-{
-	char* binaryBuffer = {};
-	std::string lib = "Lib/StateWindow/" + name + ".meta";	
-	
-	if (App->fileSystem->Exists(lib.c_str())) {
-
-		App->fileSystem->Load(lib.c_str(), binaryBuffer);
-		rapidjson::Document doc;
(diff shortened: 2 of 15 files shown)


Source/DataModels/Windows/EditorWindows/EditorWindow.cpp
```diff
@@ -1,31 +1,102 @@
...
+	std::string lib = "Lib/StateWindow/" + name + ".meta";	
+	
+	if (App->fileSystem->Load(lib.c_str(), binaryBuffer) > 0) {
+
+
+		rapidjson::Document doc;
+		Json json(doc, doc);
+		json.fromBuffer(binaryBuffer);
+		std::string buffer = std::string(binaryBuffer);
+		size_t i = buffer.find("State");
```
gerard-queralt, line 38: You can probably do `enabled = json["State"]` instead of all of this.

Source/DataModels/Windows/EditorWindows/EditorWindow.cpp
```diff
@@ -1,31 +1,102 @@
...
 }
 
-void EditorWindow::Draw(bool& enabled)
+void EditorWindow::Update_meta(bool enabled,std::string name)
 {
+	
+		rapidjson::Document doc;
+		Json json(doc, doc);
+		json["UID"] = UniqueID::GenerateUID();
+		json["Type"] = "Window";
```
gerard-queralt, line 66: Why do we need a `UID` and `Type` fields if we are not using them in de-serialization?

Source/DataModels/Windows/SubWindows/SubWindow.cpp
```diff
@@ -4,13 +4,14 @@ bool SubWindow::defaultEnabled = true;
 
 SubWindow::SubWindow(const std::string& name) : Window(name), flags(ImGuiTreeNodeFlags_None)
 {
+	enabled = defaultEnabled;
```
gerard-queralt, line 7: With your changed, the `defaultEnabled` variable can be removed. It was only used as a default parameter for these classes `Draw` method, but now the method doesn't have a parameter.

Source/DataModels/Windows/Window.h
```diff
@@ -7,14 +7,14 @@ class Window
 public:
 	virtual ~Window() {}
 
-	virtual void Draw(bool& enabled) = 0;
+	virtual void Draw() = 0;
 
 	const std::string& GetName() const;
-	
+	bool enabled;
```
gerard-queralt, line 13: Class members should be private and accessed with getter and setter functions.

Source/DataModels/Windows/WindowMainMenu.cpp
```diff
@@ -8,20 +8,22 @@ bool WindowMainMenu::defaultEnabled = true;
 WindowMainMenu::WindowMainMenu(const std::vector< std::unique_ptr<EditorWindow> >& editorWindows) : 
 	Window("Main Menu"), showAbout(false)
 {
+	
+	enabled = defaultEnabled;
```
gerard-queralt, line 12: As said above, `defaultEnabled` is not needed

Source/DataModels/Windows/WindowMainMenu.cpp
```diff
@@ -50,9 +53,9 @@ void WindowMainMenu::DrawAbout()
 {
 	if (ImGui::MenuItem("About"))
 	{
-		showAbout = !showAbout;
+		enabled = !showAbout;
```
gerard-queralt, line 56: This looks wrong, though I cannot test it because I can't run your branch. To me, this reads like you are enabling/disabling the entire main menu when the "About" button is pressed, which is not the intended behaviour.

Source/DataModels/Windows/WindowMainMenu.h
```diff
@@ -10,11 +10,10 @@ class WindowMainMenu : public Window
 
 	static const std::string repositoryLink;
 
-	void Draw(bool& enabled = defaultEnabled) override;
+	void Draw() override;
 
 	bool IsWindowEnabled(int windowIndex) const;
-	void SetWindowEnabled(int windowIndex, bool enabled);
-
+	void SetWindowEnabled(int windowIndex, bool &enabled);
```
gerard-queralt, line 16: Not sure if this is needed anymore, since now we keep a reference instead of a regular value

Source/DataModels/Windows/WindowMainMenu.h
```diff
@@ -28,15 +27,15 @@ class WindowMainMenu : public Window
...
 {
-	return windowNamesAndEnabled[windowIndex].second;
+	return windowNamesAndEnabled[windowIndex].second.get();	
 }
 
-inline void WindowMainMenu::SetWindowEnabled(int windowIndex, bool enabled)
+inline void WindowMainMenu::SetWindowEnabled(int windowIndex, bool &enabled)
 {
-	windowNamesAndEnabled[windowIndex].second = enabled;
+	windowNamesAndEnabled[windowIndex].second.get() = &enabled;
```
gerard-queralt, line 40: If this method ends up being necessary, this is wrong. Should be
```
windowNamesAndEnabled[windowIndex].second = enabled;
```
I'm not sure what your code is doing exactly in terms of the `referece_wrapper`, but it's modifying it instead of setting a new one, which might have side-effects. Better to use it as is intended.

Source/DataModels/Resources/Resource.h
```diff
@@ -13,6 +13,7 @@ enum class ResourceType
 	Scene,
 	Material,
 	SkyBox,
+	Window,
```
gerard-queralt, line 16: You have declared a new type of ResourceType, but not actually created a Resource class, which most likely means a new type of Resource is not needed. Resources should be something used by the game, like textures and meshes.

Source/FileSystem/ModuleResources.cpp
```diff
@@ -511,6 +513,8 @@ const std::string ModuleResources::GetNameOfType(ResourceType type)
 		return "Materials";
 	case ResourceType::SkyBox:
 		return "SkyBox";
+	case ResourceType::Window:
+		return "StateWindow";
```
gerard-queralt, line 517: As said in an earlier comment, resources should be things used by the game, which the windows are not. Your config files should not be placed in Assets and Lib. I think it'd be better to create a new Config folder inside Game for that, but maybe there's a better option.

Source/Modules/ModuleEditor.cpp
```diff
@@ -65,16 +65,22 @@ bool ModuleEditor::Start()
 {
 	ImGui_ImplSDL2_InitForOpenGL(App->window->GetWindow(), App->renderer->context);
 	ImGui_ImplOpenGL3_Init(GLSL_VERSION);
+	for (int i = 0; i < windows.size(); ++i) {
```
gerard-queralt, line 68: Bracket should be in the line below

Source/Modules/ModuleEditor.cpp
```diff
@@ -65,16 +65,22 @@ bool ModuleEditor::Start()
...
+		windows[i].get()->Start();
+	}
 
 	return true;
 }
 
 bool ModuleEditor::CleanUp()
 {
+	for (int i = 0; i < windows.size(); ++i) {
```
gerard-queralt, line 77: Bracket should be in the line below

Source/DataModels/Windows/EditorWindows/EditorWindow.cpp
```diff
@@ -1,31 +1,102 @@
...
+	load_meta(name.c_str(), enabled);
 }
 
-EditorWindow::~EditorWindow()
+void EditorWindow::load_meta(std::string name, bool& enable)
 {
+	char* binaryBuffer = {};
+	std::string lib = "Lib/StateWindow/" + name + ".meta";	
+	
+	if (App->fileSystem->Load(lib.c_str(), binaryBuffer) > 0) {
```
gerard-queralt, line 31: This doesn't seem to detect if the file was read. Use `ModuleFileSystem::Exists` instead

gerard-queralt (changes requested):
Left some comments regarding coding style and general engine workflow considerations. Also, I can't run this branch on my computer because it's trying to read non-existant meta files. If you delete your local `Lib` folder you'll be able to reproduce it. Left a suggestion on what the issue might be.

gerard-queralt:
Thinking more about this, wouldn't it be easier to serialize and store the dictionary in `WindowMainMenu`? And you could pass that Json as a constructor parameter instead of the current vector of windows, and then save it in `ModuleEditor::CleanUp`. This would require almost no changes to the current structure, though it might have some downsides that I'm not thinking of.

Source/DataModels/Windows/EditorWindows/EditorWindow.cpp
elbuho12Jesus: > Thinking more about this, wouldn't it be easier to serialize and store the dictionary in `WindowMainMenu`? And you could pass that Json as a constructor parameter instead of the current vector of windows, and then save it in `ModuleEditor::CleanUp`. This would require almost no changes to the current structure, though it might have some downsides that I'm not thinking of.

I don't quite understand what you mean, do you suggest that I don't load WINDOWS in `WindowMainManu.cpp` but JSONS?

PR #216, opened 2023-03-20
feat: Adding Wireframe mode and other options to debug mode

totisama:


AuxPort66 (dismissed):

Source/Modules/ModuleEditor.h
```diff
@@ -25,6 +25,7 @@ class ModuleEditor : public Module
 	void Resized();
 
 	const WindowScene* GetScene() const;
+	const std::unique_ptr<WindowDebug>* GetDebugOptions() const;
```
gerard-queralt, line 28: This is a pointer to a unique pointer, which doesn't make sense. 
```suggestion
const std::unique_ptr<WindowDebug> GetDebugOptions() const;
```
All the calls to `get()` can be removed as well. So, instead of doing `App->editor->GetDebugOptions()->get()->...` you should do `App->editor->GetDebugOptions()->...`

Source/Modules/ModuleEditor.h
```diff
@@ -42,4 +43,9 @@ class ModuleEditor : public Module
 inline const WindowScene* ModuleEditor::GetScene() const
 {
 	return scene;
-}
\ No newline at end of file
+}
+
+inline const std::unique_ptr<WindowDebug>* ModuleEditor::GetDebugOptions() const
+{
+	return &debugOptions;
```
gerard-queralt, line 50: Once you change the member, this should be
```suggestion
return debugOptions.get();
```
And update the method signature to return a raw pointer.

gerard-queralt (changes requested):
Left some comments regarding the use of unique pointers,

gerard-queralt (approved):

AuxPort66 (approved):

PR #218, opened 2023-03-20
feat[Engine Utility]: ctrl_c, ctrl_v, ctrl_d and ctrl_x

DADSILVER:
Copy, paste, cut and duplicate functions are added. Right now they can only be done with a single GameObject.


Source/DataModels/Scene/Scene.cpp
```diff
@@ -102,6 +102,41 @@ GameObject* Scene::CreateGameObject(const char* name, GameObject* parent)
 	return gameObject;
 }
 
+GameObject* Scene::CreateGameObject(const char* name, GameObject* newObject, GameObject* parent)
```
markmeliaa, line 105:
```suggestion
GameObject* Scene::DuplicateGameObject(const char* name, GameObject* newObject, GameObject* parent)
```

markmeliaa: Naming this function the same as the one that we already have that creates empty GameObjects may be a bit misleading, as this does not create a GameObject but rather duplicates one that already exists.

Source/Modules/ModuleEditor.cpp
```diff
@@ -210,3 +244,53 @@ void ModuleEditor::SetResourceOnInspector(const std::weak_ptr<Resource>& resourc
...
+	{
+		if (App->scene->GetSelectedGameObject())
+		{
+			App->scene->GetLoadedScene()->
+				CreateGameObject(copyObject->GetName(), copyObject, App->scene->GetSelectedGameObject());
+		}
+		else
+		{
+			App->scene->GetLoadedScene()->
+				CreateGameObject(copyObject->GetName(), copyObject, App->scene->GetLoadedScene()->GetRoot());
```
markmeliaa, line 270:
```suggestion
				DuplicateGameObject(copyObject->GetName(), copyObject, App->scene->GetLoadedScene()->GetRoot());
```

Source/Modules/ModuleEditor.cpp
```diff
@@ -210,3 +244,53 @@ void ModuleEditor::SetResourceOnInspector(const std::weak_ptr<Resource>& resourc
...
 }
+
+void ModuleEditor::CopyAnObject()
+{
+	if (App->scene->GetSelectedGameObject() != App->scene->GetLoadedScene()->GetRoot())
+	{
+		delete copyObject;
+		copyObject = new GameObject(*App->scene->GetSelectedGameObject());
+	}
+
```
markmeliaa, line 255: Here wouldn't be interesting to add an else, so when the user does CtrlC in the root or any of the unique lights, any past copied objects are erased? If not if could feel a bit weird to CtrlC and object, then CtrlC several times the root (for whatever reason), and when the pressing CtrlV the pasted object would be GameObject selected before the root was tried to be copied, it's just a suggestion.

Source/Modules/ModuleEditor.cpp
```diff
@@ -210,3 +244,53 @@ void ModuleEditor::SetResourceOnInspector(const std::weak_ptr<Resource>& resourc
...
+	App->scene->GetLoadedScene()->DestroyGameObject(gameObject);
+}
+
+void ModuleEditor::DuplicateAnObject()
+{
+	if (App->scene->GetSelectedGameObject() && App->scene->GetSelectedGameObject() != App->scene->GetLoadedScene()->GetRoot())
+	{
+		App->scene->GetLoadedScene()->
+			CreateGameObject(App->scene->GetSelectedGameObject()->GetName()
```
markmeliaa, line 293:
```suggestion
			DuplicateGameObject(App->scene->GetSelectedGameObject()->GetName()
```

markmeliaa (changes requested):
Some issues that I had while trying out this branch:
- The Ambient Light and the Directional Light should not be duplicated or deleted, as they are (for now), unique to each scene.
- When a GameObject with children is duplicated (either with CtrlD or CtrlV) when one of the child is selected, this same child is selected in all the GameObjects that were duplicated (and deleting that child breaks the engine).
- When closing the engine after having duplicated some objects, the engine breaks when shutting down.

Everything else worked correctly as far as I tried (sorry for this exhaustive review).

Source/Modules/ModuleEditor.h
```diff
@@ -30,13 +31,20 @@ class ModuleEditor : public Module
...
+
 	std::vector<std::unique_ptr<EditorWindow> > windows;
 	std::unique_ptr<WindowMainMenu> mainMenu = nullptr;
 	std::unique_ptr<WindowDebug> debugOptions = nullptr;
 
 	WindowInspector* inspector;
 	WindowScene* scene;
 	bool windowResized;
+
+	GameObject* copyObject;
```
gerard-queralt, line 47: Why not make the copy object a unique pointer?

gerard-queralt (changes requested):
Added a few comments on top of what Mark already said. Thanks for your work David!

Source/DataModels/Scene/Scene.h
```diff
@@ -31,6 +31,7 @@ class Scene
 	bool IsInsideACamera(const AABB& aabb) const;
 
 	GameObject* CreateGameObject(const char* name, GameObject* parent);
+	GameObject* CreateGameObject(const char* name, GameObject*, GameObject* parent);
```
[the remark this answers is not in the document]
DADSILVER: Done thx

Source/Modules/ModuleEditor.cpp
```diff
@@ -210,3 +244,53 @@ void ModuleEditor::SetResourceOnInspector(const std::weak_ptr<Resource>& resourc
...
+}
+
+void ModuleEditor::PasteAnObject()
+{
+	if(copyObject)
+	{
+		if (App->scene->GetSelectedGameObject())
+		{
+			App->scene->GetLoadedScene()->
+				CreateGameObject(copyObject->GetName(), copyObject, App->scene->GetSelectedGameObject());
```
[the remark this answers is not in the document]
DADSILVER: Done thx

Source/Modules/ModuleEditor.cpp
DADSILVER: Done thx

Source/Modules/ModuleEditor.cpp
```diff
@@ -210,3 +244,53 @@ void ModuleEditor::SetResourceOnInspector(const std::weak_ptr<Resource>& resourc
 	this->inspector->SetResource(resource);
 #endif
 }
+
+void ModuleEditor::CopyAnObject()
+{
+	if (App->scene->GetSelectedGameObject() != App->scene->GetLoadedScene()->GetRoot())
```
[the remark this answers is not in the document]
DADSILVER: Oky, i was not sure if directional light could be copy.

Source/Modules/ModuleEditor.cpp
DADSILVER: Done thx

Source/Modules/ModuleEditor.cpp
DADSILVER: Done thx

Source/DataModels/Components/ComponentAmbient.cpp
```diff
@@ -6,6 +6,11 @@ ComponentAmbient::ComponentAmbient() : ComponentLight(LightType::AMBIENT, false)
 {
 }
 
+ComponentAmbient::ComponentAmbient(ComponentAmbient& componentAmbient):
```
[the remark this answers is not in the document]
DADSILVER: Its not and ambient lights not need a copy

Source/Modules/ModuleEditor.cpp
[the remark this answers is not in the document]
markmeliaa: Well, IMO we should have multiple directional lights in the future, but for now it is unique.

Source/Modules/ModuleEditor.h
DADSILVER: I think it is better for memory management of something that is going to be created and deleted all the time.

AuxPort66 (approved):
Ole!

gerard-queralt (approved):

PR #219, opened 2023-03-20
feat: Create new scene

MarcGonzalezMoratona:


Source/DataModels/Windows/WindowMainMenu.cpp
```diff
@@ -27,51 +30,129 @@ void WindowMainMenu::Draw(bool& enabled)
...
+		ImGui::Text("Do you want to save the scene?\nAll your changes will be lost if you don't save them.");
+		ImGui::Separator();
+
+		if (ImGui::Button("Save scene", ImVec2(120, 0)))
 		{
-			ImGui::MenuItem(windowNameAndEnabled.first.c_str(), NULL, &windowNameAndEnabled.second);
+			std::string filePathName = App->scene->GetLoadedScene()->GetRoot()->GetName();
+			if (filePathName != "New Scene")
+			{
+				App->scene->SaveSceneToJson(filePathName + ".axolotl");
```
gerard-queralt, line 57:
```suggestion
App->scene->SaveSceneToJson(filePathName + SCENE_EXTENSION);
```

Prefer using the global `SCENE_EXTENSION` instead of `".axolotl"`.

Source/DataModels/Windows/WindowMainMenu.cpp
```diff
@@ -27,51 +30,129 @@ void WindowMainMenu::Draw(bool& enabled)
...
 	{
-		for (std::pair<std::string, bool>& windowNameAndEnabled : windowNamesAndEnabled)
+		ImGui::Text("Do you want to save the scene?\nAll your changes will be lost if you don't save them.");
+		ImGui::Separator();
+
+		if (ImGui::Button("Save scene", ImVec2(120, 0)))
 		{
-			ImGui::MenuItem(windowNameAndEnabled.first.c_str(), NULL, &windowNameAndEnabled.second);
+			std::string filePathName = App->scene->GetLoadedScene()->GetRoot()->GetName();
+			if (filePathName != "New Scene")
```
gerard-queralt, line 55: This is a bit picky, but I don't like having this type of conventions in fields that the user can modify. With this checks, we are arbitrarily preventing the user from having "New Scene" as a scene name. For now it's probably fine, but we should look into a better way of achieving this.

Source/DataModels/Windows/WindowMainMenu.cpp
```diff
@@ -27,51 +30,129 @@ void WindowMainMenu::Draw(bool& enabled)
...
+				quitEvent.type = SDL_QUIT;
+				SDL_PushEvent(&quitEvent);
+			}
+			ImGui::CloseCurrentPopup();
+			openPopup = false;
 		}
-		ImGui::EndMenu();
+		ImGui::SetItemDefaultFocus();
+		ImGui::SameLine();
+		if (ImGui::Button("Close without saving", ImVec2(240, 0)))
```
gerard-queralt, line 79: Everything inside this `if` block is repeated code from the previous button's `if` block. We could probably encapsulate most of it.

Source/DataModels/Windows/WindowMainMenu.h
```diff
@@ -1,6 +1,15 @@
 #pragma once
 
 #include "EditorWindows/WindowAbout.h"
+#include "EditorWindows/ImporterWindows/WindowLoadScene.h"
+#include "EditorWindows/ImporterWindows/WindowSaveScene.h"
+
+enum class Actions
```
gerard-queralt, line 7: I think it's better if we either scope the enum inside the class body (so, put it inside the class declaration) so the signature becomes `WindowMainMenu::Actions` or rename it to something more specific.

Piterga12:
When creating a new scene without any previous save of the current one, and we click on save scene, it creates the new one first and saves the new created one as the one we had.

And second, when doing the same with a big scene such as final.axolotl, it freezes

gerard-queralt (changes requested):

Source/DataModels/Windows/WindowMainMenu.cpp
MarcGonzalezMoratona: It has been done this way because we don't have a way to check that the scene has never been saved before. Maybe in the future we can solve it

gerard-queralt (approved):
Everything looks OK, thank you for your work Marc!

Piterga12 (approved):
Everything works fine, good job.

PR #221, opened 2023-03-21
feat: Metallic Workflow

sncarbo:
Two variables have been added to the materials: `smoothness` and `metallic`, as well as the possibility to load metallic maps. In case you have loaded a metallic map, the two variables mentioned above will not work.

Captura de pantalla 2023-03-21 121015

The specular workflow has been replaced by the metallic workflow for the next milestone. Subsequently, the two will coexist as two different fragment shaders. These are the fragment shader variables to be restored in the future:
```
struct Material {
    vec3 specular_color;
    float shininess;
    int has_specular_map;
    int shininess_alpha;
    int has_shininess_map;
};

uniform sampler2D specular_map;
```


UriMurcia:
Do you know why I see it all white, am I missing something?
image

sncarbo:
Thank you very much @UriMurcia for being attentive! I think it should have been fixed by now.

AuxPort66 (approved):

MacXxs (approved):
Great work Santi! Thanks!!

PR #225, opened 2023-03-22
Fix: matrix when added child

Dedado3000:
ALV


AuxPort66 (approved):
That was fast

MacXxs (approved):
Perfect!

PR #228, opened 2023-03-22
Qwe hotkeys

UriMurcia:
Now the user can use W, Q and E to change the mode from translation, rotation and scale respectively.


MacXxs (approved):
All perfect!

Vlukkee (approved):
Works really fine!

PR #229, opened 2023-03-23
Feat: Material load/save options

AuxPort66:
This solves some problems that occurred in the modification of the material. When you modified a material, the changes were only saved in case of changing a texture, not when we modified other parameters such as smoothness.

Another problem is that the load options in modifiable resources (Like material) had problems when reimporting because information that was only in the binary was lost. That's why now in engine mode load options are imported from the .meta to avoid these problems and in game mode from the binary as it should be.

To avoid having to reimport binary all the time, I add a Apply changes button, just like the one for the texture load/import options, but to ComponentMaterial.


diegusmejia117 (approved):
Rifado!

Dedado3000 (approved):
Let see if we have less bugs now

gerard-queralt (approved):

PR #232, opened 2023-03-23
Fix: Erase double Load of Mesh/Component

AuxPort66:
This is a fast patch, I'll check it better later.

When you drag and drop 4 different fbx the first one disappears.

For some reason when you load a model, mesh and material, then unload them and THEN load the mesh and material again. This problem appears. It doesn't make any sense because it should be fine, but I deleted the first Load because it was repetitive and patch the problem temporaly for the Level Design Team.


Dedado3000 (approved):
you worked really hard

AuxPort66:
> you worked really hard

I've spent all day debugging Dx

gerard-queralt (approved):

PR #233, opened 2023-03-23
feat: Filesystem import loading icon

Piterga12:


gerard-queralt (approved):

gerard-queralt (dismissed):

AuxPort66 (changes requested):
As I already told you in class, editor scene windows like MaterialInput should keep the request instead of the Import. I leave here the change request just so it doesn't merge

AuxPort66 (approved):

gerard-queralt (approved):

PR #237, opened 2023-03-26
perf: Window hierarchy

DADSILVER:
To improve the efficiency of the hierarchy, there have been three changes:

1. Add to a GameObject the state of ChildSelected. To avoid searching in the children's vector.
2. **In the function GetChildren uses reserve() to avoid the that when we do a pushback(), the vector created doesn't need a resize.**
3. Create a vector of children to avoid using the GetChildren() function if unnecessary.

This is the improvement:

image

On the left is the time with this new implementation, and on the right is the old one. This example was done with the block out of this milestone.


markmeliaa (dismissed):
I've tried moving around some objects around in an empty scene and everything seems to work properly, good job and nice improvement of performance!

Source/DataModels/GameObject/GameObject.h
```diff
@@ -240,3 +251,12 @@ inline void GameObject::setDrawBoundingBoxes(bool newDraw)
 	drawBoundingBoxes = newDraw;
 }
 
+inline const bool GameObject::IsChildSelected()
```
gerard-queralt, line 254: This method doesn't seem to be used, unless VisualStudio is lying to me. Also, I'm not sure a getter should have side-effects like changing the state of selection. I'd prefer to use more descriptive names, but that's just me, feel free to ignore me.

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
```diff
@@ -21,40 +21,40 @@ WindowHierarchy::~WindowHierarchy()
...
+        if (children.empty())
         {
-            if (child == App->scene->GetSelectedGameObject()
-                && StateOfSelection::SELECTED == App->scene->GetSelectedGameObject()->GetStateOfSelection())
-            {
-                ImGui::SetNextItemOpen(true);
-            }
+            flags |= ImGuiTreeNodeFlags_Leaf;
+        }
+        else
```
gerard-queralt, line 49: Empty else block.

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
```diff
@@ -21,40 +21,40 @@ WindowHierarchy::~WindowHierarchy()
...
-    if (App->scene->GetLoadedScene()->GetRoot())
-    {
-        DrawRecursiveHierarchy(App->scene->GetLoadedScene()->GetRoot());
-    }
+    assert(App->scene->GetLoadedScene()->GetRoot());
+    DrawRecursiveHierarchy(App->scene->GetLoadedScene()->GetRoot(), true);
 }
 
-void WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject)
+void WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject, bool isRoot)
```
gerard-queralt, line 28: Why add a boolean to the signature? Don't think the comparisson is expensive enough to have any benefits in removing it, and in my opinion it makes the method less intuitive to use.

gerard-queralt (changes requested):
Left a few minor comments, overall great job on your improvements to the hierarchy!

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
PaablooCH, line 28: you can set the variable inside of the function and avoid sending it by parameter, like you did with children vector

PaablooCH (approved):
Nice job friend 👍 just one thing

Source/DataModels/GameObject/GameObject.h
DADSILVER: True, my bad this function had been used before some changes.

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
DADSILVER: Yes, thank you.

Issue #240, opened 2023-03-27
WindowInspector modifies the underlying string pointer of both name and tag, which is considered unsafe


gerard-queralt:
As of writing this issue (last commit in `develop` is 3af99e1e0d) there are a [couple ](https://github.com/Horizons-Games/Axolotl-Engine/blob/3af99e1e0d84e2d44fe89980c54ca27255800c7b/Source/DataModels/Windows/EditorWindows/WindowInspector.cpp#L55) [examples](https://github.com/Horizons-Games/Axolotl-Engine/blob/3af99e1e0d84e2d44fe89980c54ca27255800c7b/Source/DataModels/Windows/EditorWindows/WindowInspector.cpp#L66) in WindowInspector that use a C-style cast to remove the `const` modifier from the returned values of `GameObject::GetName`. That alone should be avoided, but in this specific instance it's used to modify the underlying string returned by `std::string::c_str`, which is unsafe according to the [documentation](https://cplusplus.com/reference/string/string/c_str/). Relevant part:
> A program shall not alter any of the characters in this sequence.

This is not a huge issue for this upcoming VS, but should be fixed.

PR #242, opened 2023-03-27
refactor: avoid modifying underlying char pointer

gerard-queralt:
Explained in #240.
Some notes:
- When the original string value is empty, the setter methods fail unless `c_str` is called or the string is over a given size (there are a couple comments where this happens). I tried to find an explanation but couldn't figure out a reason, and it's honestly mind-boggling. If anyone reading this can explain it to me I'll really appreciate it ^^'.
- I didn't change the instance of this issue in `WindowComponentMeshRenderer` because it requires more changes that I'd like (either adding it as a friend class in Resource or adding a `SetLibraryPath`, both of which I don't like). Left a comment to try to keep in mind that this should be changed.


gerard-queralt:
Also fixed a bug that's been there since at least [2.0.0](https://github.com/Horizons-Games/Axolotl-Engine/releases/tag/2.0.0). Shows that it's not really important, but well.

markmeliaa (approved):
Nice refactor, is this also related to the task we had in the last milestone about changing all the `const char *` members to `std::string` members, or this has nothing to do with that task?

gerard-queralt:
> is this also related to the task we had in the last milestone about changing all the `const char *` members to `std::string` members

No, if I'm not mistaken all members were already changed to `std::string`. The changes I did were only to the signatures of the methods that I touched for this refactor. From what I know, it's generally preferred by the C++ community to use `std::string` when possible since it's considered more idiomatic, and doesn't have major performance differences, but at the end it comes down to personal preference, so I don't mind changing it back.

In this case I also did it because the string is returned by copy, so we can't directly access the member as we did previously since we returned a pointer to the underlying string, which isn't inherently a bad thing, but kind of defeats the purpose of having private members. [Here's an example of this in Java, where this is a common issue](https://xperti.io/blogs/getters-and-setters-java-basics-common-mistakes-best-practices/#Mistake_3_Returning_the_object_reference_directly_in_the_getter). Again, not a big deal, I don't mind if you'd rather use char pointers.

markmeliaa:
> No, if I'm not mistaken all members were already changed to `std::string`. The changes I did were only to the signatures of the methods that I touched for this refactor. From what I know, it's generally preferred by the C++ community to use `std::string` when possible since it's considered more idiomatic, and doesn't have major performance differences, but at the end it comes down to personal preference, so I don't mind changing it back.
> 
> In this case I also did it because the string is returned by copy, so we can't directly access the member as we did previously since we returned a pointer to the underlying string, which isn't inherently a bad thing, but kind of defeats the purpose of having private members. [Here's an example of this in Java, where this is a common issue](https://xperti.io/blogs/getters-and-setters-java-basics-common-mistakes-best-practices/#Mistake_3_Returning_the_object_reference_directly_in_the_getter). Again, not a big deal, I don't mind if you'd rather use char pointers.

Gotcha! Thank you for explaining it more in-depth!

AuxPort66 (dismissed):

AuxPort66 (dismissed):

AuxPort66 (approved):
👀

markmeliaa (approved):

PR #251, opened 2023-03-29
fix: game object can be deleted regardless of children

### LINKED ISSUE #249: Delete game object stopped working since merge with Hirearchy optimizations
Deleting a game object in the scene while in the editor throws and error from WindowHirearchy::Draw() function.
(There must be a null pointer in there somewhere to the deleted object)

(Sometimes it crashes sometimes it doesn't).

image

gerard-queralt:
Should fix #249

### DIFF
--- a/Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
+++ b/Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
@@ -34,7 +34,7 @@ void WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject)
     sprintf_s(gameObjectLabel, "%s###%p", gameObject->GetName(), gameObject);
     ImGuiTreeNodeFlags flags = ImGuiTreeNodeFlags_OpenOnArrow | ImGuiTreeNodeFlags_OpenOnDoubleClick;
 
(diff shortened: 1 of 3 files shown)


Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
```diff
@@ -270,9 +270,9 @@ void WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject)
 
     if (nodeDrawn) // If the parent node is correctly drawn, draw its children
     {
-        for (int i = 0; i < children.size(); ++i)
+        for (GameObject* child : gameObject->GetChildren())
```
markmeliaa, line 273: Which is the difference between the former way using an additional vector `children` and this way?

gerard-queralt: While what I said is correct, @DADSILVER let me know that `GetChildren` is one of the most expensive things done in the Hierarchy window, so it's noticeably better to call it only once ^^'

gerard-queralt (commented):
I can't approve the pull request since I'm the author, but thank you David for finding a better solution than mine!

markmeliaa (approved):

PaablooCH (approved):

PR #258, opened 2023-04-02
style: added clang-format file and re-formatted all classes

gerard-queralt:
Visual Studio uses [clang-format](https://clang.llvm.org/docs/ClangFormat.html) to format the code, including how it autocompletes it. By providing our custom file, we can make it so when we use VS's autocomplete it will automatically follow our programming rules!

I want that we can all agree on the format our code should follow, so especially for this PR give as much feedback as you want. You can download the `.clang-format` file and use [this handy website](https://zed0.co.uk/clang-format-configurator/) to see the implications of each rule. You can also see the official [documentation](https://clang.llvm.org/docs/ClangFormatStyleOptions.html).

Note that the version installed with Visual Studio 2019 is 12, so we can't use definitions from newer versions (like [this](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces) awesome one that inserts braces after control statements like `if` and the like).


markmeliaa (approved):

AuxPort66 (approved):

PaablooCH:
I read the file and tested. Particularly I don't think that imposing the rules by the force is a good option instead of teaching how to follow them, what I mean is that file modifies your code while you are writing, I would like something that underlines your code instead of changing it. Also, there are things that are not in our programming rules that this file takes for valid. To conclude this type of PR I would put it first in a branch that required 12 or 15 approval before merging (or waiting to have it here before merging to develop)

gerard-queralt:
Thanks for the feedback Pablo. My idea was that, by having Visual Studio autocomplete following our coding guidelines, it would make it easier for everyone to code. Personally, I feel that by having to change what Visual Studio completes, I'm having to write that code twice, so I'd like to have this in our project. I hope you see why I disagree with you on that regard.
On the other hand, I know that some of these changes are not in our programming rules, that's why I want your feedback, so that we can see which ones we agree on and which ones we prefer to follow a different structure. And I totally agree that this type of MR should be verified by as much people as possible, and I don't intend to merge it until we have an agreement.
Let me know if you see my point regarding auto-formatting, and please point out which things you don't like about the formatting so we can all discuss it.

Source/Application.cpp
```diff
@@ -1,18 +1,18 @@
...
 #include "ModuleEditor.h"
+#include "ModuleInput.h"
+#include "ModuleProgram.h"
+#include "ModuleRender.h"
+#include "ModuleScene.h"
+#include "ModuleUI.h"
+#include "ModuleWindow.h"
 #ifndef ENGINE
-#include "ModulePlayer.h"
+#	include "ModulePlayer.h"
```
MacXxs, line 15: Hi, shouldn't the include be together with the '#' insted of tabbed? Or is it some kind of mistake?

Source/DataModels/Camera/CameraEngine.cpp
```diff
@@ -227,39 +224,37 @@ void CameraEngine::Zoom()
...
 
 void CameraEngine::Focus(const OBB& obb)
 {
-
 	Sphere boundingSphere = obb.MinimalEnclosingSphere();
 
 	float radius = boundingSphere.r;
-	if (boundingSphere.r < 1.f) radius = 1.f;
+	if (boundingSphere.r < 1.f)
+		radius = 1.f;
```
MacXxs, line 244: This should be between brackets no?

Source/DataModels/Components/ComponentAmbient.cpp
```diff
@@ -28,14 +27,14 @@ void ComponentAmbient::SaveOptions(Json& meta)
 {
 	// Do not delete these
 	meta["type"] = GetNameByType(type).c_str();
-	meta["active"] = (bool)active;
-	meta["removed"] = (bool)canBeRemoved;
+	meta["active"] = (bool) active;
```
MacXxs, line 30: I personally prefer having the casts together with the variable like this: `(bool)active;`
But if everyone else prefers it this way I don't mind.

MacXxs (commented):
As long as everyone is okey with it, I don't mind using clang! But as already stated, we need everyone to share his opinion on this 'cause this PR affects the coding for everyone involved in the project.

Source/Application.h
```diff
@@ -56,7 +55,7 @@ class Application
 	ModuleEditor* editor;
 
 private:
-	std::vector<std::unique_ptr<Module> > modules;
+	std::vector<std::unique_ptr<Module>> modules;
```
MarcGonzalezMoratona, line 58: a teacher in the past told me that using this >> together can lead to some confusion with the stream extraction operator >>

Are you sure that this part of the formatting is a good idea?

Source/Auxiliar/CollectionAwareDeleter.h
```diff
@@ -1,12 +1,12 @@
 #pragma once
 
-#include <memory>
 #include <functional>
+#include <memory>
```
MarcGonzalezMoratona, line 4: I know it's not part of the PR, but why do we have this includes separated from the ones below?

Source/DataModels/Camera/Camera.cpp
```diff
@@ -236,19 +235,20 @@ bool Camera::IsInside(const AABB& aabb)
 				break;
 			}
 		}
-		if (!onPlane) return false;
+		if (!onPlane)
```
MarcGonzalezMoratona, line 238: I like to have conditions in one line when they don't have curly braces { }

Source/DataModels/Camera/Camera.cpp
```diff
@@ -352,14 +353,14 @@ void Camera::SetLookAt(const float3& lookAt, bool& isSameRotation)
 	{
 		isSameRotation = true;
 	}
-	else 
+	else
 	{
-		float3 nextDirection = Quat::SlerpVector(currentDirection, targetDirection, App->GetDeltaTime() * rotationSpeed * 2);
+		float3 nextDirection =
```
MarcGonzalezMoratona, line 358: Probably this is happening because we set the max line length to 120 characters, I don't like to define a variable in more than a line, but I don't know if we can do this better

Source/DataModels/Components/ComponentDirLight.cpp
```diff
@@ -5,24 +5,23 @@
 #include "Modules/ModuleScene.h"
 
 #ifndef ENGINE
-#include "Modules/ModuleEditor.h"
-#include "Modules/ModuleDebugDraw.h"
+#	include "Modules/ModuleDebugDraw.h"
```
MarcGonzalezMoratona, line 8: I think this shouldn't be tabbed

Source/DataModels/Components/ComponentMaterial.cpp
```diff
@@ -118,12 +118,12 @@ void ComponentMaterial::Draw()
 
 				glActiveTexture(GL_TEXTURE6);
 				glBindTexture(GL_TEXTURE_2D, texture->GetGlTexture());
-				glUniform1f(4, material->GetNormalStrength()); //normal_strength
-				glUniform1i(6, 1); //has_normal_map
+				glUniform1f(4, material->GetNormalStrength()); // normal_strength
+				glUniform1i(6, 1);							   // has_normal_map
```
MarcGonzalezMoratona, line 122: here something strange happened

Source/DataModels/Components/ComponentTransform.cpp
```diff
@@ -110,40 +121,39 @@ void ComponentTransform::UpdateTransformMatrices()
...
+		ComponentTransform* childTransform =
+			static_cast<ComponentTransform*>(child->GetComponent(ComponentType::TRANSFORM));
 		childTransform->UpdateTransformMatrices();
 	}
 }
 
 void ComponentTransform::CalculateLightTransformed(const ComponentLight* lightComponent,
-	bool translationModified,
-	bool rotationModified)
+												   bool translationModified,
```
MarcGonzalezMoratona, line 133: here something happened

Source/DataModels/Components/UI/ComponentBoundingBox2D.cpp
```diff
@@ -109,19 +109,19 @@ void ComponentBoundingBox2D::CalculateWorldBoundingBox()
 	pivotDifference.x = -pivotPosition.x + 0.5f;
 	pivotDifference.y = pivotPosition.y - 0.5f;
 
-	worldAABB.minPoint = position.xy().Mul(float2(1.0f, -1.0f).Mul(screenFactor)) + screenSize / 2.0f
-		+ (localAABB.minPoint + pivotDifference).Mul(transform2D->GetSize().Mul(transform2D->GetScale().xy()).Mul(screenFactor));
-	worldAABB.maxPoint = position.xy().Mul(float2(1.0f, -1.0f).Mul(screenFactor)) + screenSize / 2.0f
-		+ (localAABB.maxPoint + pivotDifference).Mul(transform2D->GetSize().Mul(transform2D->GetScale().xy()).Mul(screenFactor));
+	worldAABB.minPoint = position.xy().Mul(float2(1.0f, -1.0f).Mul(screenFactor)) + screenSize / 2.0f +
+						 (localAABB.minPoint + pivotDifference)
```
MarcGonzalezMoratona, line 113: these lines look a bit weird

Source/DataModels/Components/UI/ComponentCanvas.h
```diff
@@ -25,23 +24,24 @@ class ComponentCanvas :  public Component
...
 }
 
-inline float2 ComponentCanvas::GetScreenReferenceSize() const {
+inline float2 ComponentCanvas::GetScreenReferenceSize() const
+{
 	return screenReferenceSize;
 }
 
-inline float2 ComponentCanvas::GetSize() {
+inline float2 ComponentCanvas::GetSize()
```
MarcGonzalezMoratona, line 39: these can be const. Is there a way to tell the formatter to suggest const methods when we are creating a getter? maybe if the method name starts with "Get"...

Source/DataModels/Components/UI/ComponentCanvas.h
```diff
@@ -25,23 +24,24 @@ class ComponentCanvas :  public Component
...
+{
 	return size;
 }
 
-inline float ComponentCanvas::GetScreenFactor() {
+inline float ComponentCanvas::GetScreenFactor()
```
MarcGonzalezMoratona, line 44: this can also be const

Source/DataModels/Components/UI/ComponentImage.cpp
```diff
@@ -138,32 +137,29 @@ void ComponentImage::LoadOptions(Json& meta)
...
-	float vertices[] = {
-		// positions          
-		-0.5,  0.5, 0.0f, 1.0f,
-		-0.5, -0.5, 0.0f, 0.0f,
-		 0.5, -0.5, 1.0f, 0.0f,
-		 0.5, -0.5, 1.0f, 0.0f,
-		 0.5,  0.5, 1.0f, 1.0f,
-		-0.5,  0.5, 0.0f, 1.0f
+	float vertices[] = { // positions
+						 -0.5, 0.5,	 0.0f, 1.0f, -0.5, -0.5, 0.0f, 0.0f, 0.5,  -0.5, 1.0f, 0.0f,
```
MarcGonzalezMoratona, line 161: this formatting for arrays is really weird. The array was a lot more readable before

Source/DataModels/GameObject/GameObject.h
```diff
@@ -195,9 +194,11 @@ inline const std::vector<GameObject*> GameObject::GetChildren() const
...
 	rawChildren.reserve(children.size());
 
-	if(!children.empty())
-		std::transform(std::begin(children), std::end(children), std::back_inserter(rawChildren), 
-			[] (const std::unique_ptr<GameObject>& go) { return go.get(); });
+	if (!children.empty())
+		std::transform(std::begin(children),
+					   std::end(children),
+					   std::back_inserter(rawChildren),
+					   [](const std::unique_ptr<GameObject>& go) { return go.get(); });
```
MarcGonzalezMoratona, line 201: here I think we can do something better if we configure the formatter to avoid having the body of a method in one line

Source/DataModels/Program/Program.cpp
```diff
@@ -1,12 +1,17 @@
...
-#include "Globals.h"
 #include "GL/glew.h"
+#include "Globals.h"
 #include <MathGeoLib/Include/Math/float3.h>
 
-Program::Program(unsigned vertexShader, unsigned fragmentShader,
-	const std::string& vtxShaderFileName, const std::string& frgShaderFileName, const std::string& programName) :
-	vertexShaderFileName(vtxShaderFileName), fragmentShaderFileName(frgShaderFileName), programName(programName),
+Program::Program(unsigned vertexShader,
+				 unsigned fragmentShader,
```
MarcGonzalezMoratona, line 8: I think this should be indented at the same level than the lines below

Source/DataModels/Resources/ResourceMaterial.cpp
```diff
@@ -1,10 +1,17 @@
 #include "ResourceMaterial.h"
 #include "FileSystem/Json.h"
 
-ResourceMaterial::ResourceMaterial(UID resourceUID, const std::string& fileName, const std::string& assetsPath,
-	const std::string& libraryPath) : Resource(resourceUID, fileName, assetsPath, libraryPath),
-	diffuseColor(float4(1.0f, 1.0f, 0.0f, 1.0f)), /*specularColor(float3(0.5f, 0.3f, 0.5f)),
-	shininess(512.0f),*/ normalStrength(1.0f), smoothness(0.5f), metalness(0.0f), /*shininessAlpha(false),*/
+ResourceMaterial::ResourceMaterial(UID resourceUID,
+								   const std::string& fileName,
```
MarcGonzalezMoratona, line 5: this indentation does not seem correct

Source/DataModels/Resources/ResourceMaterial.h
```diff
@@ -13,77 +13,76 @@ struct LoadOptionsMaterial
...
-	//bool HasShininessAlpha();
+	// bool HasSpecular();
+	// bool HasShininessAlpha();
 	bool HasMetallicMap();
 	bool HasMetallicAlpha();
 
 	LoadOptionsMaterial& GetLoadOptions();
 
-	//Sets
+	// Sets
```
MarcGonzalezMoratona, line 52:
```suggestion
	// Setters
```

MarcGonzalezMoratona: do we need a comment for this?

Source/DataModels/Resources/ResourceMesh.cpp
```diff
@@ -40,14 +49,14 @@ void ResourceMesh::CreateVBO()
 	glGenBuffers(1, &vbo);
 	glBindBuffer(GL_ARRAY_BUFFER, vbo);
 
-							//position			//uv				//normal		
+	// position			//uv				//normal
```
MarcGonzalezMoratona, line 52: this comment has changed indentation and maybe it lost the meaning

Source/DataModels/Resources/ResourceScene.h
```diff
@@ -10,23 +10,26 @@ struct OptionsScene
...
 
 	ResourceType GetType() const override;
 
-	void SaveImporterOptions(Json& meta) override {};
-	void LoadImporterOptions(Json& meta) override {};
+	void SaveImporterOptions(Json& meta) override{};
+	void LoadImporterOptions(Json& meta) override{};
 
-	//Getters
+	// Getters
```
MarcGonzalezMoratona, line 24: are this comments important?

Source/DataModels/Resources/ResourceSkyBox.cpp
```diff
@@ -3,134 +3,119 @@
...
+	for (int i = 0; i < textures.size(); ++i)
+	{
+		std::shared_ptr<ResourceTexture> textI = textures[i];
+
+		if (textI)
+		{
+			textI->Load();
+			std::vector<uint8_t> aux = textI->GetPixels();
+			glTexImage2D(GL_TEXTURE_CUBE_MAP_POSITIVE_X + i,
+						 0,
```
MarcGonzalezMoratona, line 39: too much indentation here

Source/DataModels/Resources/ResourceSkyBox.cpp
```diff
@@ -3,134 +3,119 @@
...
-
-    glGenBuffers(1, &vbo);
-    glBindBuffer(GL_ARRAY_BUFFER, vbo);
-
-    glBufferData(GL_ARRAY_BUFFER, sizeof(skyboxVertices), &skyboxVertices, GL_STATIC_DRAW);
-
-    ENGINE_LOG("Generated VBO %i.", vbo);
-}
+	float skyboxVertices[] = { // positions
+							   -1.0f, 1.0f,	 -1.0f, -1.0f, -1.0f, -1.0f, 1.0f,	-1.0f, -1.0f,
```
MarcGonzalezMoratona, line 85: strange formatting

Source/DataModels/Windows/EditorWindows/ImporterWindows/WindowSaveScene.cpp
```diff
@@ -7,7 +7,7 @@ WindowSaveScene::WindowSaveScene() : WindowFileBrowser()
 {
 	isSave = true;
 	dialogName = " Save As...";
-	title = ICON_IGFD_SAVE" Save As...";
+	title = ICON_IGFD_SAVE " Save As...";
```
MarcGonzalezMoratona, line 10: please check that this space is not added to the menus

Source/DataModels/Windows/EditorWindows/WindowFileBrowser.cpp
```diff
@@ -126,17 +119,21 @@ void WindowFileBrowser::DrawWindowContents()
 
 void WindowFileBrowser::Browser()
 {
-	fileDialogBrowser.OpenDialog("embedded", "Select File", ".*",browserPath, "", -1, nullptr,
-		ImGuiFileDialogFlags_NoDialog |
-		ImGuiFileDialogFlags_DisableBookmarkMode |
-		ImGuiFileDialogFlags_DisableCreateDirectoryButton);
+	fileDialogBrowser.OpenDialog("embedded",
```
MarcGonzalezMoratona, line 122: indentation can be improved here

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
```diff
@@ -21,264 +21,259 @@ WindowHierarchy::~WindowHierarchy()
...
-    {
-        for (GameObject* child : children)
-        {
-            DrawRecursiveHierarchy(child);
-        }
-        ImGui::TreePop();
-    }
+	assert(gameObject);
+
+	// Delete a GameObject with the SUPR key
```
MarcGonzalezMoratona, line 32:
```suggestion
	// Delete a GameObject with the `Delete / Supr` key
```

Supr key is for Spanish keyboard, other keyboard have `Delete`

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
```diff
@@ -21,264 +21,259 @@ WindowHierarchy::~WindowHierarchy()
...
+	if (gameObject != App->scene->GetLoadedScene()->GetRoot() &&
+		gameObject != App->scene->GetLoadedScene()->GetAmbientLight() &&
+		gameObject != App->scene->GetLoadedScene()->GetDirectionalLight())
+	{
+		if (App->input->GetKey(SDL_SCANCODE_DELETE) == KeyState::DOWN)
+		{
+			if (gameObject == App->scene->GetSelectedGameObject())
+			{
+				App->scene->SetSelectedGameObject(gameObject->GetParent()); // If a GameObject is destroyed,
+																			// change the focus to its parent
```
MarcGonzalezMoratona, line 42: indentation should be improved here

Source/Application.cpp
gerard-queralt: I'll try to compile a list of things like this that are not in the programming rules and set up a vote on Discord. I'll look into indentation options for the includes as well to see if there's one to keep the `#` next to it,

Source/DataModels/Camera/CameraEngine.cpp
gerard-queralt: Yes, but the option for this is from clang 15.0, and the one installed with Visual Studio 2019 is 12.0, so I can't add a rule for this. It will need to be done manually, unfortunately.

Source/Application.h
gerard-queralt: From what I know this is not an issue with the current compilers. After looking a bit around on the web I didn't see any source saying that this is an issue either, but of course I could've missed it. The other thing is that there's not an option for these cases in particular, only for spaces around angles in general. So this would look like
```
std::vector < std::unique_ptr < Module > > modules;
```
Which I think looks horrible :P

Source/Auxiliar/CollectionAwareDeleter.h
gerard-queralt: In general, we've tried to group includes based on what they are actually including. In most cases we have first the include of the header file (in source files), the includes for standard libraries (like these ones), the includes for Application, Globals and Modules (which are kind of the basic Engine stuff), the includes for data models (like GameObject, Resource, etc.) and the includes for external dependencies. Of course not necessarily in this order, and sometimes not even grouped at all, but this was something we agreed to try to do at the beginning of the project. One of the options in clang merges all the includes in one block, we can consider it if you want.

Source/DataModels/Camera/Camera.cpp
gerard-queralt: Thing is, they _should_ have curly braces, but as I said above there's not an automatic way to add them in our version. I'll add them manually.

Source/Application.h
MarcGonzalezMoratona: Take into account your answer, I think that we can leave it as is. Keep the angles together and let's see if everything is working fine, and only change it if we get a problem with the compiler in the future

Source/Auxiliar/CollectionAwareDeleter.h
MarcGonzalezMoratona: I think it's great as we have done until now. Thank you for the explanation!

Source/DataModels/Components/ComponentTransform.cpp
gerard-queralt: Looks like it has something to do with GitHub formatting of tabs. Here is how this line looks in Visual Studio
image

Source/DataModels/Components/UI/ComponentBoundingBox2D.cpp
gerard-queralt: Again, seems to be a GitHub thing
image

Source/DataModels/Components/ComponentTransform.cpp
MarcGonzalezMoratona: This can create confusion when reviewing code in the future, so it's a good thing to know for all the team. But regarding configuration, I think there's nothing to change

gerard-queralt: I'll try to see if there's a way to fix it in either side just in case, thanks for noticing Marc :)

Source/DataModels/Components/UI/ComponentCanvas.h
MarcGonzalezMoratona: This is something we have to agree on as a team, because maybe not everybody wants the same warnings and formatting, so I suggest to talk about it, at least with the ones who contribute the most to the project

Source/Modules/ModuleRender.cpp
```diff
@@ -1,41 +1,47 @@
...
+void __stdcall OurOpenGLErrorFunction(GLenum source,
+									  GLenum type,
+									  GLuint id,
+									  GLenum severity,
+									  GLsizei length,
+									  const GLchar* message,
+									  const void* userParam)
 {
-	const char* tmpSource = "", * tmpType = "", * tmpSeverity = "";
+	const char *tmpSource = "", *tmpType = "", *tmpSeverity = "";
```
MarcGonzalezMoratona, line 41: is it ok to separate pointer like this?

Source/DataModels/Components/UI/ComponentImage.cpp
gerard-queralt: Looks like there's not an option to keep the number of columns in an array initializer. If we want to preserve the previous format, we should do:
```c++
// clang-format off
float vertices[] = ...
//clang-format on
```
For reference, given how GitHub formats this, here's how this looks in VS:
image

Source/DataModels/Components/UI/ComponentCanvas.h
gerard-queralt: Of course, that's why I left that branch there like two months ago. But if we agree to it, we could use it.

Source/DataModels/Components/UI/ComponentImage.cpp
MarcGonzalezMoratona: In this case I think we should add clang-format off when we create formatted arrays or matrices, although maybe people can find it annoying, so I don't know what's the best option. In VS it does not look as bad as in GitHub, but maybe is not good enough

Source/DataModels/Components/ComponentTransform.cpp
gerard-queralt: I think it might be related to the fact that I chose to indent with tabs (as we agreed at the beginning of the project) instead of spaces, so the format gets mixed up in GitHub :/

gerard-queralt: That was it! [This article](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-personal-account-settings/managing-your-tab-size-rendering-preference) explains how to change the tab size in your GitHub appearance to use 4 spaces (what VS uses) instead of 8. After doing that it looks perfect in the visualization:
image

Source/DataModels/Components/UI/ComponentImage.cpp
gerard-queralt: I don't know how often we will initialize arrays or vectors this way, I haven't finished looking at your comments but I think it's rather rare, so it's probably not an issue.

MarcGonzalezMoratona: I think you will see it 3-4 times in the whole project. Not a big issue

Source/DataModels/Components/ComponentTransform.cpp
gerard-queralt: Looks like we can add an `.editorconfig` file in the repo, as per [this StackOverflow comment](https://stackoverflow.com/questions/8833953/how-to-change-tab-size-on-github/33831598#33831598). I can take a look, seems rather simple.

MarcGonzalezMoratona: I think that the best option is to have it in the repository, to have something unified for the whole team. Although it's really easy to configure for the user, but maybe we want it different in another repository

Source/DataModels/GameObject/GameObject.h
MarcGonzalezMoratona: Yes! This looks really good! Thank you for your work Gerard

Source/DataModels/Windows/EditorWindows/ImporterWindows/WindowSaveScene.cpp
gerard-queralt: It's not! This is just a feature of C++ that implicitly concatenates string literals, it's basically the same thing as writing
```c++
	title = ICON_IGFD_SAVE + " Save As...";
```
This is especially useful when breaking strings, so you can do something like:
```c++
std::string aString = "Hello"
    " World!";
```
In case you're curious, [this article](https://www.techiedelight.com/concatenate-string-literals-c-cpp/) provides a bit more information.

Source/Modules/ModuleRender.cpp
gerard-queralt: Looks like it's because multiple variables are declared in the same line. I tried breaking it up into three different declarations and it aligns the pointer to the left.

Source/FileSystem/ModuleResources.cpp
```diff
@@ -365,16 +373,14 @@ void ModuleResources::CreateAssetAndLibFolders()
...
-	//ensure this vector is updated whenever a new type of resource is added
-	std::vector<ResourceType> allResourceTypes = { ResourceType::Material,
-												  ResourceType::Mesh,
-												  ResourceType::Model,
-												  ResourceType::Scene,
-												  ResourceType::Texture,
-												  ResourceType::SkyBox };
+	// ensure this vector is updated whenever a new type of resource is added
+	std::vector<ResourceType> allResourceTypes = {
+		ResourceType::Material, ResourceType::Mesh,	   ResourceType::Model,
```
MarcGonzalezMoratona, line 381: this vector was better formatted before

Source/DataModels/Windows/SubWindows/WindowHardware.cpp
```diff
@@ -140,20 +140,20 @@ void WindowHardware::GetSoftwareVersions()
...
 	int ramInMB = SDL_GetSystemRAM();
 	float ramInGB = ramInMB / 1000.f;
 	float ramInGb = ramInGB * 8.f;
-	//this is the value multiplied by 10 to the power of 8
-	//insted of to the power of 9, which is what Giga is
-	//this is to make rounding easier,
-	//since I can't find a way to set decimal precision
-	int ramInGbOneDecimalAux = (int)std::ceil(ramInGb * 10.f);
+	// this is the value multiplied by 10 to the power of 8
+	// insted of to the power of 9, which is what Giga is
```
MarcGonzalezMoratona, line 151:
```suggestion
	// instead of to the power of 9, which is what Giga is
```

Source/Modules/ModuleDebugDraw.cpp
```diff
@@ -1,587 +1,582 @@
 #include "ModuleDebugDraw.h"
```
MarcGonzalezMoratona, line 1: I'm not sure, but I think this module was given from the teachers. The diff is huge and really difficult to compare so I can't do a good check. But I think it's not a big issue. Most of the things to change will appear in other files

Source/FileSystem/Importers/MeshImporter.cpp
```diff
@@ -28,30 +28,26 @@ void MeshImporter::Import(const char* filePath, std::shared_ptr<ResourceMesh> re
 	delete saveBuffer;
 }
 
-void MeshImporter::Save(const std::shared_ptr<ResourceMesh>& resource, char* &fileBuffer, unsigned int& size)
+void MeshImporter::Save(const std::shared_ptr<ResourceMesh>& resource, char*& fileBuffer, unsigned int& size)
```
MarcGonzalezMoratona, line 31: This is more readable separated *&, I hope it does not conflict with other pointer rules

gerard-queralt: Unfortunately it does, but now that you point it out, I'm not sure a reference to a pointer makes sense, this could probably be just a pointer and work the same.

Source/DataModels/Components/ComponentTransform.cpp
gerard-queralt: I added the `.editorconfig` file, but even though it works when viewing files regularly, it doesn't seem to work when viewing the diff in the PR :/

AuxPort66 (dismissed):

Source/Modules/ModuleRender.cpp
```diff
@@ -259,8 +269,10 @@ update_status ModuleRender::Update()
 	glDepthFunc(GL_LEQUAL);
 	glEnable(GL_BLEND);
 	glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
-	for (std::map<float, const GameObject*>::reverse_iterator it = transparentGOToDraw.rbegin(); it != transparentGOToDraw.rend(); ++it)
-	{	
+	for (std::map<float, const GameObject*>::reverse_iterator it = transparentGOToDraw.rbegin();
+		 it != transparentGOToDraw.rend();
+		 ++it)
```
UriMurcia, line 274: Having the ++it in the second line would be possible?

UriMurcia (dismissed):
Everything is okay for me, great job and nice idea!

Dedado3000 (dismissed):
As you commented there are some standard that we can't follow due to tthe version of vs but is okey for me

Source/Modules/ModuleRender.cpp
gerard-queralt: From the looks of it not in version 12. I mean, it would be possible, but it uses the same rule as function calls, so when breaking calls we wouldn't put each parameter in a different line (see [documentation](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#binpackarguments)), which in my opinion looks worse.

markmeliaa (dismissed):

AuxPort66 (dismissed):

UriMurcia (approved):

MarcGonzalezMoratona (approved):

PR #259, opened 2023-04-02
perf[Lights]: delete program where is not necessary

PaablooCH:
When you create or assign data to a ssbo is not necessary to use any program. They are stored in globally so any shader can use them.


UriMurcia (approved):
okay, works fine!

MacXxs (approved):
Everything working 👍🏼

PR #261, opened 2023-04-02
style: add padding to spot light

PaablooCH:
Added the padding to avoid the calculation of the offset for each light

### DIFF
--- a/Source/DataModels/Components/ComponentSpotLight.h
+++ b/Source/DataModels/Components/ComponentSpotLight.h
@@ -11,6 +11,7 @@ struct SpotLight
 	float3 aim;
 	float innerAngle = 0.0f;
 	float outAngle = 0.0f;
+	float3 padding = { 0, 0, 0 };
 };
 
 class Json;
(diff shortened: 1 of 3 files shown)


MacXxs:
Hi Pablo!

This is not a fix, it's more of a "style" type of PR, we aren't actually fixing anythin that wasn't working , so i would change the type to style.

There were a couple of lines that were still using the "hack" of putting the manual calculation of the offset for the spotlights so i've changed it. On top of that, Carlos told me to put the padding both on the shader and the struct for readability purposes so I've added the padding on the shader too.

A part from this, all the program->BindUniformBlock calls for the lights aren't needed ('cause we have the binding points already) so i've removed them from the code (this should have us extra performance).

I've done the changes directly 'cause it's faster and less time consuming, hope you are okey with it!

MacXxs (approved):

PaablooCH:
> Hi Pablo!
> 
> This is not a fix, it's more of a "style" type of PR, we aren't actually fixing anything that wasn't working , so i would change the type to style.
> 
> There were a couple of lines that were still using the "hack" of putting the manual calculation of the offset for the spotlights so i've changed it. On top of that, Carlos told me to put the padding both on the shader and the struct for readability purposes so I've added the padding on the shader too.
> 
> A part from this, all the program->BindUniformBlock calls for the lights aren't needed ('cause we have the binding points already) so i've removed them from the code (this should gives us extra performance).
> 
> I've done the changes directly 'cause it's faster and less time consuming, hope you are okey with it!

ok, thanks 👍

MacXxs (approved):

AuxPort66 (approved):

PR #265, opened 2023-04-03
feat: created ComponentScript with some semblance of reflection

gerard-queralt:
Note that we need to check that the scripts in a separate project can inherit from this new class, so we can use it as an interface for the scripts that the user will create. If this approach is not compatible with hot reloading, we would need to so something similar to what's done in #263

### DIFF
--- a/Source/Auxiliar/Reflection/Field.h
+++ b/Source/Auxiliar/Reflection/Field.h
@@ -1,0 +1,32 @@
+#pragma once
+
+#include <string>
+#include <functional>
+
+enum class FieldType
+{
+	FLOAT
+};
+
+template<typename T>
+struct Field
+{
+public:
+	const std::string name;
+	const std::function<T(void)> getter;
+	const std::function<void(const T&)> setter;
+	FieldType type;
+
+	Field(const std::string& name,
+		  const std::function<T(void)>& getter,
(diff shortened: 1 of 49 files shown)


markmeliaa (approved):
Nice approach, good job!

Source/Auxiliar/Reflection/Field.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include <string>
+#include <functional>
+
+enum class FieldType
+{
+	FLOAT
```
markmeliaa, line 8: How many field types should we support? Anything else outside the basic variable types (STRING, INT, FLOAT...)? Maybe GameObjects or some Components?

markmeliaa (approved):

Source/Auxiliar/Reflection/Field.h
gerard-queralt: This is something that we should discuss with the game team. Strings are an easy one, GameObjects are also pretty important I would imagine, and then maybe things like MathGeoLib's `float3`. I don't know if we'll need `int`s, since we can just use `float`s, but it'd be easy to add. But I don't imagine we won't forget something, so as you guys get going and require more stuff, we can add in the engine whatever you need!

PR #268, opened 2023-04-04
feat: Bounding box inside ComponentTransform

AuxPort66:
This moves the bounding box inside the ComponentTransform so we can have 2D and 3D objects.

It also solves an error that happened in the hierarchy when clicking on the canvas, caused by a merge conflict


Source/DataModels/Components/ComponentTransform.h
```diff
@@ -35,13 +39,20 @@ class ComponentTransform : public Component
 	const float3& GetGlobalRight() const;
 	const float3& GetGlobalScale() const;
 
+	const float4x4& GetLocalMatrix() const;
+	const float4x4& GetGlobalMatrix() const;
+
+	const AABB& GetLocalAABB();
+	const AABB& GetEncapsuledAABB();
+	const OBB& GetObjectOBB();
+	const bool isDrawBoundingBoxes() const;
```
gerard-queralt, line 48: Should be in PascalCase, and I don't think it makes any sense to return a `const bool` by copy.
```
	bool IsDrawBoundingBoxes() const;
```

AlexisDieu (approved):

PR #269, opened 2023-04-05
feat: Skinning FileSystem

sncarbo:


DADSILVER (commented):
Good job, but let me tell you some suggestions.

Source/DataModels/Resources/ResourceMesh.h
```diff
@@ -63,6 +91,8 @@ class ResourceMesh : virtual public Resource
 	void CreateEBO();
 	void CreateVAO();
 
+	static constexpr unsigned int bonesPerVertex = 4;
```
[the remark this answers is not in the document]
sncarbo: Done! Thank you for noticing.

DADSILVER (approved):

Source/DataModels/Resources/ResourceMesh.h
```diff
@@ -33,24 +48,37 @@ class ResourceMesh : virtual public Resource
 	unsigned int GetNumVertices();
 	unsigned int GetNumFaces();
 	unsigned int GetNumIndexes();
+	unsigned int GetNumBones();
```
MarcGonzalezMoratona, line 51: I think this and the rest of the getters of this class can be const

Source/DataModels/Resources/ResourceMesh.h
```diff
@@ -146,6 +187,21 @@ inline const std::vector<std::vector<unsigned int> >& ResourceMesh::GetFacesIndi
...
+}
+
+inline const std::vector<Attach>& ResourceMesh::GetAttaches() const 
+{
+	return attaches;
+}
+
+inline const std::vector<unsigned int>& ResourceMesh::GetNumWeights() const
+{
+	return this->numWeights;
```
MarcGonzalezMoratona, line 202: can we do it this way? If it works I think is more simple
```suggestion
	return numWeights;
```

Source/DataModels/Resources/ResourceMesh.h
```diff
@@ -196,3 +257,36 @@ inline void ResourceMesh::SetFacesIndices(const std::vector<std::vector<unsigned
...
 }
+
+inline void ResourceMesh::SetBones(const std::vector<Bone>& bones)
+{
+	this->bones = bones;
+}
+
+inline void ResourceMesh::SetAttachResize()
+{
+	this->attaches.resize(this->numVertices);
```
MarcGonzalezMoratona, line 268: do you think we can avoid using `this->` to simplify this kind of methods?

Source/FileSystem/Importers/MeshImporter.cpp
```diff
@@ -166,5 +213,50 @@ void MeshImporter::Load(const char* fileBuffer, std::shared_ptr<ResourceMesh> re
...
+
+	std::vector<Bone> bones;
+	std::vector<unsigned int> allNumWeights;
+	resource->SetAttachResize();
+	unsigned int vertexId = 0u;
+	float vertexWeight = 0.0f;
+	unsigned int numWeights = 0u;
+	for (unsigned int i = 0; i < resource->GetNumBones(); ++i)
+	{
+		Bone bone;
```
MarcGonzalezMoratona, line 228: if this bone is only used in the scope of this loop, is there a way to delete its memory once we have finished working with it?

Source/DataModels/Resources/ResourceMesh.h
sncarbo: Done! Thank you for pointing this out.

Source/DataModels/Resources/ResourceMesh.h
sncarbo: I think that according to the programming rules it has to be in this format.

MarcGonzalezMoratona (dismissed):

Source/DataModels/Resources/ResourceMesh.h
```diff
@@ -71,12 +103,16 @@ class ResourceMesh : virtual public Resource
...
 	unsigned int numIndexes;
+	unsigned int numBones;
 	unsigned int materialIndex;
 	std::vector<float3> vertices;
 	std::vector<float3> textureCoords;
 	std::vector<float3> normals;
 	std::vector<float3> tangents{};
 	std::vector<std::vector<unsigned int> > facesIndices;
+	std::vector<Bone> bones;
+	std::vector<Attach> attaches;
```
PaablooCH, line 114: Why don't u make these vectors a vector of pointers? will be easier to pass them through the getters

Source/DataModels/Resources/ResourceMesh.h
```diff
@@ -116,36 +152,56 @@ inline unsigned int ResourceMesh::GetNumIndexes()
...
+}
+
+inline const std::vector<Attach>& ResourceMesh::GetAttaches() const 
+{
+	return attaches;
+}
+
+inline const std::vector<unsigned int>& ResourceMesh::GetNumWeights() const
+{
+	return this->numWeights;
```
PaablooCH, line 202: this-> is not necessary

Source/DataModels/Resources/ResourceMesh.h
PaablooCH, line 268: this-> is not necessary in both variables

Source/DataModels/Resources/ResourceMesh.h
```diff
@@ -196,3 +257,36 @@ inline void ResourceMesh::SetFacesIndices(const std::vector<std::vector<unsigned
...
+}
+
+inline void ResourceMesh::SetAttachResize()
+{
+	this->attaches.resize(this->numVertices);
+}
+
+inline void ResourceMesh::SetAttachBones(const unsigned int vertexId, const unsigned int boneId)
+{
+	this->attaches[vertexId].bones[this->attaches[vertexId].numBones] = boneId;
```
PaablooCH, line 273: this-> is not necessary

Source/DataModels/Resources/ResourceMesh.h
```diff
@@ -196,3 +257,36 @@ inline void ResourceMesh::SetFacesIndices(const std::vector<std::vector<unsigned
...
+inline void ResourceMesh::SetNumWeights(const std::vector<unsigned int>& numWeights)
+{
+	this->numWeights = numWeights;
+}
+
+inline void ResourceMesh::IncrementAttachNumBones(const unsigned int vertexId)
+{
+	if (this->attaches[vertexId].numBones < bonesPerVertex)
+	{
+		++this->attaches[vertexId].numBones;
```
PaablooCH, line 290: this-> is not necessary

Source/DataModels/Resources/ResourceMesh.h
```diff
@@ -196,3 +257,36 @@ inline void ResourceMesh::SetFacesIndices(const std::vector<std::vector<unsigned
...
+}
+
+inline void ResourceMesh::SetNumWeights(const std::vector<unsigned int>& numWeights)
+{
+	this->numWeights = numWeights;
+}
+
+inline void ResourceMesh::IncrementAttachNumBones(const unsigned int vertexId)
+{
+	if (this->attaches[vertexId].numBones < bonesPerVertex)
```
PaablooCH, line 288: this-> is not necessary

PaablooCH (changes requested):
Hi, good job. I see some things in the code:

Source/DataModels/Resources/ResourceMesh.h
sncarbo: Done! I have removed all the `this` that you mentioned above.

Source/DataModels/Resources/ResourceMesh.h
sncarbo: The struct destructors need to be explicitly defined and I had thought that this was not necessary as the size in bytes should not be very large.

PaablooCH (approved):

MacXxs (approved):
Everything working! 👍🏼

PR #270, opened 2023-04-05
refactor[Engine/animation]: remove extra nodes from hierarchy

MacXxs:
Removed the drawing of the extra nodes that assimp creates when importing models, with the information of the pre_rotations and pre_translations applied to the bones (this commit only affects the imgui drawing of the hirearchy, we must keep the nodes themselves for the animations)

<b>Before</b>             |  <b>After</b>
:-------------------------:|:-------------------------:
image  |  image

### DIFF
--- a/Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
+++ b/Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
@@ -29,108 +29,153 @@ void WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject)
 {
     assert(gameObject);
 
-    // Delete a GameObject with the SUPR key
-    if (gameObject != App->scene->GetLoadedScene()->GetRoot() &&
-        gameObject != App->scene->GetLoadedScene()->GetAmbientLight() &&
-        gameObject != App->scene->GetLoadedScene()->GetDirectionalLight())
+    // Do not draw assimp extra nodes for translations and rotations
+    if (gameObject->GetName().find("$AssimpFbx$") == std::string::npos)
     {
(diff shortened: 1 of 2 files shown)


DADSILVER (approved):

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
```diff
@@ -29,108 +29,153 @@ void WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject)
...
+            App->scene->ChangeSelectedGameObject(gameObject);
+        }
 
-    if (ImGui::BeginPopupContextItem("RightClickGameObject", ImGuiPopupFlags_MouseButtonRight))
-    {
-        if(gameObject->GetComponent(ComponentType::TRANSFORM) != nullptr)
+        if (ImGui::BeginPopupContextItem("RightClickGameObject", ImGuiPopupFlags_MouseButtonRight))
         {
-            if (ImGui::MenuItem("Create Empty child"))
+            if (gameObject->GetComponent(ComponentType::TRANSFORM) != nullptr)
```
markmeliaa, line 101: This if its a bit useless, every gameobject has to have a transform, its not a possiblity, its an obligation, so this if will always resolve to `true`.

markmeliaa (changes requested):
Nice! Just one comment.

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
markmeliaa: Wait, just noticed that you did not add this line of code, idk why GitHub shows like you changed those, so don't mind me.

markmeliaa (approved):

PaablooCH (approved):

PR #271, opened 2023-04-05
Refactor : merge the ComponentMaterial into ComponentMeshRenderer

AlexisDieu:


PaablooCH (changes requested):
I see two things:

1. If you set a color, delete the material and load one again, the window doesn't update with the new data:
image

2. the buttons apply and reset are not present
image

AlexisDieu (commented):
You right i test on develop, this issue is not there because it's using the apply and reset button but when i merge develop into this branch these button was not there i am gonna try to add them back/fix this issue, idk if we need the apply button if we change it automatically

PaablooCH:
> idk if we need the apply button if we change it automatically
Professors told us that having those buttons was necessary to avoid mistakes with the material

PaablooCH (changes requested):
Alexis, the component doesn't work like in develop. In develop it seems that the purpose of the buttons is to update the material once you click on "apply", now in this branch you can change the diffuse color and it is updated in real time, but in develop, you need to apply first the changes before see them. I suppose that you need to adapt it as in develop. Tell me if i'm wrong

AlexisDieu (commented):
I know what you mean but the changement of diffuse color is only visual, if you delete the material and get it back you will see it will be the same color as the beginning but if you save it apply the current diffuse color (if it's not working like i said tell me i will modify it but for me it's working) for texture you need to apply to see the result but i planned to modify it later to have the visual of the texture then apply it

PaablooCH (approved):

MacXxs (approved):

PR #272, opened 2023-04-07
feat: get animation info from assimp

MarcGonzalezMoratona:


Source/DataModels/Resources/ResourceAnimation.h
```diff
@@ -7,36 +7,65 @@
...
+	void SetChannels(std::unordered_map<std::string, ResourceAnimation::Channel*> channels);
 
 protected:
 	void InternalLoad() override {};
 	void InternalUnload() override;
 
 private:
-	float duration;
-	std::unordered_map<char*, Channel*> channels;
+	double duration;
```
PaablooCH, line 39: why you change to double? We really need that precision?

Source/DataModels/Resources/ResourceModel.h
```diff
@@ -55,6 +56,9 @@ class ResourceModel : virtual public Resource
 
 	void AppendNode(Node* node);
 
+	void SetAnimations(const std::vector<std::shared_ptr<ResourceAnimation>>& animations);
+	const std::vector<std::shared_ptr<ResourceAnimation>>& GetAnimations() const;
```
PaablooCH, line 60: Should this not return a vector of raw pointers?

Source/DataModels/Windows/EditorWindows/WindowInspector.cpp
```diff
@@ -149,6 +149,13 @@ void WindowInspector::InspectSelectedGameObject()
 				}
 			}
 
+			if (!lastSelectedGameObject->GetComponent(ComponentType::ANIMATION)) {
+				if (ImGui::MenuItem("Create Animation Component"))
+				{
+					AddComponentAnimation();
```
PaablooCH, line 155:
```suggestion
					App->scene->GetSelectedGameObject()->CreateComponent(ComponentType::ANIMATION);
```
you can delete the function at 357

Source/Globals.h
```diff
@@ -24,7 +24,7 @@ enum class TextureType
 
 // Application -------------
 #define TITLE "Axolotl Engine"
-#define VERSION "1.0.0"
+#define VERSION "0.0.3"
```
PaablooCH, line 27:
```suggestion
#define VERSION "0.3.0"
```
our last tag was that

PaablooCH (changes requested):
One thing a part of my changes is (idk if it's only me) all the animation files are outside of its filters, it's just a reminder for later.
image

Source/DataModels/Resources/ResourceAnimation.h
MarcGonzalezMoratona: Assimp defines mDuration as a double, that's why we did it the same way
https://assimp.sourceforge.net/lib_html/structai_animation.html

But if you think that a float is enough we can change it

MarcGonzalezMoratona:
> One thing a part of my changes is (idk if it's only me) all the animation files are outside of its filters, it's just a reminder for later. image

I had some problems adding new files to the project, I will try to fix it. Thank you for noticing!

Source/DataModels/Resources/ResourceAnimation.h
PaablooCH: ok, thanks for the explanation

Source/DataModels/Windows/EditorWindows/WindowInspector.cpp
MarcGonzalezMoratona: we are encapsulating into methods all the logic related to components. There are other methods such as AddComponentPlayer, AddComponentRigidBody and AddComponentMockState that are defined in a similar way. I think that it's good to have it like this for reutilisation purposes. I would suggest to do a refactor in the future if we don't find this useful

Source/DataModels/Resources/ResourceModel.h
MarcGonzalezMoratona: methods for GetMeshes and GetMaterials were returning a vector of shared_ptr. I don't know which approach is better, so I did something similar. Why do you think it's better to return a vector of raw pointers here?

PaablooCH: i find working with raw is easier than smart, also you don't need to bring shared pointer's logic with you if the purpose of calling GetAnimation is to get only the info in ResourceAnimation

gerard-queralt: When working with resources we should always use shared/weak pointers, since they could be deleted at any point. If someone wants to get this resources we should assume that they intend to extend the lifespan of the object. Furthermore, converting the inner vector of shared to a vector of raw is a relatively expensive operation, and as I said I don't think it has any benefit in this case to justify it.

PaablooCH (dismissed):

MacXxs (changes requested):
'm getting this error when drag and dropping the Gunan model with animations provided by the teachers:

image

It seems that one of the animations when requesting the resource during the load function is trying to get a resource with an empty path so it breaks.

Also this PR doesn't have the latest changes in animation, so they should be merged in this branch before pushing the pr.

MacXxs (approved):

PR #277, opened 2023-04-13
chore[main]: allow manual pipeline

gerard-queralt:
Cherrypick of #276. According to the docs (linked in the original MR), the file needs to be in the default branch of the repository in order for the changes to take effect.

### DIFF
--- a/.github/workflows/msbuild.yml
+++ b/.github/workflows/msbuild.yml
@@ -10,6 +10,7 @@ on:
     branches: [ "main" ]
   pull_request:
     branches: [ "**" ]
+  workflow_dispatch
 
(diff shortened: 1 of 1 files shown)


AuxPort66 (approved):

Dedado3000 (approved):
this branch commits into main, is this ok?

gerard-queralt:
As I said in the description of this MR, this change is only applied if present in the default branch of our repository, which is `main`. I also did this change in `develop` (in fact, this is just a cherrypick), but I figured there wasn't any need to either wait for another release or merge the entirety of `develop` onto `main` just for this.

PR #278, opened 2023-04-14
FEAT: Basic game play mode

DADSILVER:
- OnPlay and OnStop states are created in the application.
- An own timer is created for the OnPlay state.
- Frustum for the camera is changed to a raw pointer because it has no copy constructor.
- Copy constructors are created for player and rigid-body components.
- To exit the OnPlay state, press ctrl+Q.
- The player is copied for the Onplay state. In the stop, the player copied is deleted and the actual player is turned on to enable.


Source/Modules/ModulePlayer.cpp
```diff
@@ -126,10 +156,29 @@ void ModulePlayer::LoadNewPlayer()
...
 		if (camera->GetParent()->GetComponent(ComponentType::PLAYER))
 		{
+#ifdef ENGINE
+			SetPlayer(std::make_unique<GameObject>(static_cast<GameObject&>(*camera->GetParent())));
+			lastPlayer = camera->GetParent();
+			// look for the player's camera
+			std::vector<GameObject*> children = player.get()->GetChildren();
+			for (auto child : children)
+			{
+				if (!child->GetComponentsByType<ComponentCamera>(ComponentType::CAMERA).empty())
```
PaablooCH, line 166: could you create cameraPlayer outside this if and check if its null, saving making this call?

Game/Assets.zip
PaablooCH, line 1: delete this zip

PaablooCH (changes requested):
I didn't find anything wrong testing it. Just 3 things

Source/DataModels/Camera/Camera.cpp
```diff
@@ -47,7 +46,7 @@ Camera::Camera(Camera& camera)
 	lastMouseX(camera.lastMouseX),
 	lastMouseY(camera.lastMouseY),
 	mouseState(camera.mouseState),
-	frustum(std::move(camera.frustum))
+	frustum(new Frustum(*camera.frustum))
```
gerard-queralt, line 49: Instead of changing the frustum to a raw pointer (which you aren't deleting, if I didn't miss it), you could do something like
```suggestion
	frustum(std::unique_ptr<Frustum>(new Frustum(*camera.frustum)))
```

Source/DataModels/Timer/Timer.cpp
```diff
@@ -11,6 +11,7 @@ Timer::~Timer()
 void Timer::Start()
 {
 	startTick = SDL_GetTicks();
+	stopped = false;
```
gerard-queralt, line 14: This is changing the functionality of the class. If I recall correctly, a Timer isn't supposed to be reusable after calling `Stop`. We would need to make sure this doesn't break anything, at the very least.

gerard-queralt (changes requested):
Left a comment regarding `Camera::frustum` and a minor thing with `Timer`.

I don't particularly like the idea of having the Application as a whole be aware of whether we are on play mode or not. Wouldn't it make more sense to store that information (boolean and Timer) in `ModuleScene`? In particular, for Scripting, a script's `Init` and `Start` need to be called when entering play mode, and `CleanUp` when exiting. As it stands, we would need two additional methods in `ModuleScene` to call them to do that functionality. If we moved `OnPlay` and `OnStopPlay`, we could have everything together in that same method body. I'd like to know your thoughts on this, it's possible I'm missing something important.

MacXxs:
> Left a comment regarding `Camera::frustum` and a minor thing with `Timer`.
> 
> I don't particularly like the idea of having the Application as a whole be aware of whether we are on play mode or not. Wouldn't it make more sense to store that information (boolean and Timer) in `ModuleScene`? In particular, for Scripting, a script's `Init` and `Start` need to be called when entering play mode, and `CleanUp` when exiting. As it stands, we would need two additional methods in `ModuleScene` to call them to do that functionality. If we moved `OnPlay` and `OnStopPlay`, we could have everything together in that same method body. I'd like to know your thoughts on this, it's possible I'm missing something important.

For Animation I need the OnPlay, OnStop and OnUpdate on Application as it is referenced to be done this way in the ppt. On top of this, if we put the OnPlay, OnStop and OnUpdate on the Application every module and component that needs to access this state simply needs check the application's isOnPlayMode state. I wouldn't change where things need to be implemented, and do it as it is stipulated on the documentation that we are given.

AuxPort66 (changes requested):
We should have the ability to use the mouse in Play mode by hitting a key. This is essential to be able to use the editor with the scripting while we test on Play mode.

Source/Modules/ModulePlayer.cpp
```diff
@@ -126,10 +158,33 @@ void ModulePlayer::LoadNewPlayer()
...
+			//	}
+			//}
+			SetPlayer(camera->GetParent());
+			cameraPlayer = static_cast<ComponentCamera*>(camera->GetComponent(ComponentType::CAMERA))->GetCamera();
+			cameraPlayer->SetAspectRatio(App->editor->GetAvailableRegion().first / App->editor->GetAvailableRegion().second);
+			App->scene->GetLoadedScene()->GetRootQuadtree()->RemoveGameObjectAndChildren(player);
+			//lastPlayer->Disable();
+			App->camera->SetSelectedCamera(0);
+#else
 			SetPlayer(camera->GetParent()->GetParent()->RemoveChild(camera->GetParent()));
```
gerard-queralt, line 182: I'm pretty sure this line won't work after your change to `SetPlayer`. Can you try compiling in Game mode just to make sure?

Source/Modules/ModulePlayer.h
```diff
@@ -21,17 +21,34 @@ class ModulePlayer : public Module
...
 	void LoadNewPlayer();
+	void UnloadNewPlayer();
 
 	bool IsStatic();
 
+	void SetReadyToEliminate(bool readyToEliminate);
+
 private:
-	std::unique_ptr <GameObject>  player;
+	GameObject*  player;
```
gerard-queralt, line 36: Why was this changed?

PaablooCH (dismissed):

gerard-queralt (dismissed):

gerard-queralt (dismissed):
I'm trying on Mark's branch ([scripting/component-script-window](https://github.com/Horizons-Games/Axolotl-Engine/tree/scripting/component-script-window)), so this might be inacurate. But holding `Ctrl+A` ends up locking the mouse in place if you don't exit play mode fast enough. Could you check this doesn't happen in your branch?

Source/Application.cpp
```diff
@@ -100,3 +100,29 @@ bool Application::CleanUp()
...
+{
+	isOnPlayMode = false;
+	input->SetShowCursor(true);
+	player->UnloadNewPlayer();
+	onPlayTimer->Stop();
+	scene->OnStop();
+}
+
+void Application::OnPause()
+{
```
gerard-queralt, line 127: Not a big deal, but I would leave the call to `scene->OnPause()` ready, even if it's not currently used.

Source/Application.h
```diff
@@ -31,12 +29,18 @@ class Application
...
 	bool CleanUp();
 
+	void OnPlay();
+	void OnStop();
+	void OnPause();
+
 	void SetMaxFrameRate(int maxFrames);
 	int GetMaxFrameRate() const;
 	float GetDeltaTime() const;
+	bool GetIsOnPlayMode() const;
```
gerard-queralt, line 39:
```suggestion
	bool IsOnPlayMode() const;
```

Source/Application.h
```diff
@@ -49,19 +53,19 @@ class Application
...
-#ifndef ENGINE
 	ModulePlayer* player;
-#endif // ENGINE
 	ModuleDebugDraw* debug;
 	ModuleEditor* editor;
 
 private:
 	std::vector<std::unique_ptr<Module> > modules;
 	std::unique_ptr<Timer> appTimer;
+	std::unique_ptr<Timer> onPlayTimer;
```
gerard-queralt, line 63: Now that I whink about it, maybe both timers could be references? Maybe it's not possible.
```suggestion
	Timer onPlayTimer;
```

Source/DataModels/Components/ComponentMaterial.cpp
```diff
@@ -0,0 +1,384 @@
+#pragma warning (disable: 4804)
```
gerard-queralt, line 1: This entire file should not be added.

Source/DataModels/Components/ComponentPlayer.cpp
```diff
@@ -9,6 +9,12 @@ ComponentPlayer::ComponentPlayer(bool active, GameObject* owner)
 {
 }
 
+ComponentPlayer::ComponentPlayer(const ComponentPlayer& componentPlayer) :
```
gerard-queralt, line 12: I think we already talked about this, but could you remind me why we need these copy constructors? Looks pretty redundant to me.

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
```diff
@@ -36,7 +37,7 @@ void WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject)
     {
         if (App->input->GetKey(SDL_SCANCODE_DELETE) == KeyState::DOWN)
         {
-            if (gameObject == App->scene->GetSelectedGameObject())
+            if (gameObject == App->scene->GetSelectedGameObject() && gameObject !=App->player->GetPlayer())
```
gerard-queralt, line 40: I know we currently require a Player in Game/Play mode, but that shouldn't be the case. Is there a reason beyond that that we couldn't delete the player in play mode? Feels a bit arbitrary, but I could be missing something.

Source/Modules/ModulePlayer.cpp
```diff
@@ -138,10 +181,20 @@ void ModulePlayer::LoadNewPlayer()
...
+			return;
 		}
 	}
+	isPlayerLoad = false;
+	ENGINE_LOG("Player is not load");
 }
 
+void ModulePlayer::UnloadNewPlayer()
+{
+	App->camera->SetSelectedCamera(-1);
```
gerard-queralt, line 194: Not really related to this PR so feel free to ignore this, but `SetSelectedCamera` should probably receive an `enum`. @Dedado3000

gerard-queralt (changes requested):
Thank you for your work David, left a few comments.

Source/Application.cpp
DADSILVER: Sure.

Source/Application.h
DADSILVER: I don't think so. In this case both timers are only used in the application. And they are used during the whole execution, so I think it is worth to have them unique to not have to deal with the delete.

Source/DataModels/Components/ComponentMaterial.cpp
DADSILVER: this file does not exist anymore, no?

markmeliaa: yup, it does not

Source/DataModels/Components/ComponentPlayer.cpp
DADSILVER: At the beginning, the player was copied at the time of executing the scene in order not to make a temporary save and reload the scene. But I have left it in case you need to do a CTRL + c at some point.

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
DADSILVER: If it is deleted I am sure it will crash because it is not controlled. but it also seems a bit strange to be able to delete a player during the execution. maybe later if for some reason it is interesting we can add that freature. :)

PaablooCH (approved):

Source/Application.h
gerard-queralt: Sorry, that's not what I mean. I think you are thinking about using a `Timer*`. In which case, yeah, I think it's better to have it as a unique pointer. But I'm wondering if it could be a plain `Timer`, since it will never be null and it doesn't need to be passed around. If I'm not mistaken it should also be more efficient, but I could be wrong. As I said in the original comment, maybe it's not possible, but wanted to ask just in case.

Source/DataModels/Components/ComponentPlayer.cpp
gerard-queralt: Yes, but that was not my question. This copy constructor is the same that is implicitly declared with every class. That is, I think you can delete it and the copy will work just fine. If that's the case, would there be any reason to keep this constructor around?
You can check in `scripting-main` that we don't declare an explicit copy constructor, but we still copy the component just fine because the default one is exactly what we need.

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
gerard-queralt: I imagine it would crash, so not really something to change in this PR. But I was wondering if there was any reason for this check besides that. Ideally, there shouldn't be any issues with deleting the player, since it should be a regular GameObject for the most part.

Source/Application.h
DADSILVER: Ahhh, a see your point. I Will change thx.

markmeliaa (dismissed):
Everything correct, we can merge this whenever, so merging scripting will be easier.

gerard-queralt (approved):
Thank you for your work David!

AuxPort66 (approved):

PR #279, opened 2023-04-14
Feat : Add and remove materials

AlexisDieu:
- Allow to add and remove a material from a mesh
- Can change a material for an another one
- Use a default material when no one is binded

### DIFF
--- a/Source/DataModels/Batch/GeometryBatch.h
+++ b/Source/DataModels/Batch/GeometryBatch.h
@@ -50,19 +50,22 @@ public:
 	GeometryBatch();
 	~GeometryBatch();
 
+	const int GetFlags() const;
+
 	void CreateVAO();
 
 	void AddComponentMeshRenderer(ComponentMeshRenderer* newComponent);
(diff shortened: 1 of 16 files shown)


Source/DataModels/Batch/GeometryBatch.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "Math/float3.h"
+#include "Resources/ResourceMaterial.h"
```
PaablooCH, line 4: this include is not necessary we already have it in forward declaration

PaablooCH (changes requested):
Just one thing, but the other part of the code are good. After this we only have 1 job 🎊

Source/DataModels/Batch/GeometryBatch.h
AlexisDieu: Done thanks I forgot to remove it after tested

PaablooCH (approved):

MacXxs (approved):

PR #280, opened 2023-04-15
feat: Bones GO Hierarchy & Skinning CPU Algorithm

sncarbo:
Now each ComponentMeshRenderer contains a vector with the bone hierarchy that is affected by each transformation of the global matrix of its GameObject.


MacXxs:
Hi Santi! Remember to remove the metas and the meshes! hahahah

MacXxs (changes requested):
Hi Santi! Remember to remove the metas and the meshes! hahahah

sncarbo:
> Hi Santi! Remember to remove the metas and the meshes! hahahah

Done Miki!

Source/DataModels/Scene/Scene.cpp
```diff
@@ -340,6 +364,101 @@ GameObject* Scene::SearchGameObjectByID(UID gameObjectID) const
...
+		{
+			isNode = false;
+			isParentNode = false;
+
+			for (const Bone& bone : bones)
+			{
+				if (child->GetParent()->GetName() == bone.name)
+				{
+					isParentNode = true;
+					break;
```
MarcGonzalezMoratona, line 383: Why do you need this break?

MarcGonzalezMoratona (changes requested):
Hi Santi! I think that the meta files are still there. Did you commit the last changes?
I also see that the pipeline is not working. I pressed "retry" because it seems to be a timeout.

Source/DataModels/Scene/Scene.cpp
MarcGonzalezMoratona: I suppose that this method is making a search and when you find the root bone you skip the next iterations of the loop. Can you confirm?

sncarbo: What the method essentially does is to look for the rootbone within the bones of each GO. The rootbone is the one whose name appears within the mesh bone names of that GO and whose parent's name **does not** appear. In other words, you could say that if the name of the parent appears, we already know that it is not the rootbone. That's why we put the `break`, because we don't need to look any further because we already know it's not the rootbone.

sncarbo:
> Hi Santi! I think that the meta files are still there. Did you commit the last changes?
> 
> I also see that the pipeline is not working. I pressed "retry" because it seems to be a timeout. (SOLVED)

I have deleted the metas and meshes from the animation branch, not the others. Do you want me to delete them all?

MacXxs:
> > Hi Santi! I think that the meta files are still there. Did you commit the last changes?
> > 
> > I also see that the pipeline is not working. I pressed "retry" because it seems to be a timeout. (SOLVED)
> 
> I have deleted the metas and meshes from the animation branch, not the others. Do you want me to delete them all?

Hey Santi, yeah we should only push the code modified, not the metas nor the meshes, so remove them all please!

sncarbo:
> > > Hi Santi! I think that the meta files are still there. Did you commit the last changes?
> > > I also see that the pipeline is not working. I pressed "retry" because it seems to be a timeout. (SOLVED)
> > 
> > 
> > I have deleted the metas and meshes from the animation branch, not the others. Do you want me to delete them all?
> 
> Hey Santi, yeah we should only push the code modified, not the metas nor the meshes, so remove them all please!

Done!

Game/Assets/Animation/Take 001.anim
MacXxs, line 1: This too should be removed ahaha. With this out of the way it should be everything! 😵‍💫

MacXxs (changes requested):

Game/Assets/Animation/Take 001.anim
sncarbo: I deleted it!

MacXxs (approved):

PR #281, opened 2023-04-15
refactor: split move parent

PaablooCH:
See -> https://app.clickup.com/t/865c1tzt0
I change the return of some function that returns unique_pointer to a simple raw pointer. First I release the unique and then i return the raw. This solution is more clean and remove the necessity of doing constantly move(unique), the unique is created once we need it.


gerard-queralt:
Thank you for your work Pablo. If you don't mind, I would like to come to an agreement as a team on whether these methods should use unique or raw pointers. As I explained in a comment in the ticket you linked, talking with Carlos after class he made it very clear that not having unique pointers in the method's signature is just his personal preference and doesn't really affect performance in any way, and that we were free to disagree with him.
I know you prefer working with raw pointers, but in this case we would need to remember to _always_ call `delete` in the cases we are not calling `LinkChild`, otherwise we will have memory leaks; in my opinion, using unique pointers makes this easier. Again, I know you disagree with me, but as I said at the beginning, I think we should reach an agreement as a team.
Aside from this, everything looks great to me. Thanks again for your work!

PaablooCH:
> talking with Carlos after class he made it very clear that not having unique pointers in the method's signature is just his personal preference and doesn't really affect performance in any way, and that we were free to disagree with him. 

i saw it, thats why i explain why i make these changes.

> but in this case we would need to remember to always call delete in the cases we are not calling LinkChild, otherwise we will have memory leaks

at the moment we only use Unlink before Link. I cannot find a purpose to unlink a child unless we want to remove it and for that case we have the option delete. Also doing a `delete pointer` I think that we can achieve it.

gerard-queralt:
> Also doing a `delete pointer` I think that we can achieve it.

Of course, but I don't think there's any advantatge on calling `delete` instead of allowing the unique pointer to delete itself by going out of scope. I worry that at some point someone might forget about it and cause a preventable memory leak.

MacXxs:
I personally prefer to follow the prefrences of the teachers so I'm inclined in using raw pointers and deleting them when needed,

markmeliaa (approved):
Everything looks and works well, thanks Pablo!

gerard-queralt:
Would you guys mind if we discuss this before merging? Maybe in the next engine meeting or in some recess. Not like this is needed for the vertical slice anyway.

gerard-queralt:
I'll ask about this in tomorrow's engine meeting. It goes without saying, but everyone is free to come and have their voices heard.

gerard-queralt (approved):

PR #283, opened 2023-04-16
feat: Engine/animation controller animation

MacXxs:
* Now when the controller is on play state the transformations of the first Animation are applied to the bones of the Model.
* Fixed AnimationImportaion problems with the position and rotations

<b>TODO: When David's implementation for the OnPlay/OnStop modes for Application is merged we need to adapt the controller to play the animations or stop playing them relative to the Application's OnPlay/OnStop.</b>

### DIFF
--- a/Source/DataModels/Animation/AnimationController.cpp
+++ b/Source/DataModels/Animation/AnimationController.cpp
@@ -4,7 +4,7 @@
 #include "Resources/ResourceAnimation.h"
 
 AnimationController::AnimationController() :
-    currentTime(0.0f), loop(true), play(false), resource(nullptr)
+    currentTime(0.0f), isLooping(true), isPlaying(false), resource(nullptr)
 {
 }
 
@@ -15,25 +15,25 @@ AnimationController::~AnimationController()
 void AnimationController::Play(std::shared_ptr<ResourceAnimation> resource, bool loop)
 {
     this->resource = resource;
-    this->loop = loop;
-    play = true;
+    isLooping = loop;
+    isPlaying = true;
 }
 
 void AnimationController::Stop()
 {
-    play = false;
+    isPlaying = false;
 }
 
 void AnimationController::Update()
 {
-    if (play && resource)
+    if (isPlaying && resource)
     {
         float duration = resource->GetDuration();
 
         currentTime += App->GetDeltaTime();
         if (currentTime > duration)
         {
-            if (loop)
+            if (isLooping)
             {
                 currentTime = 0.0f;
             }
--- a/Source/DataModels/Animation/AnimationController.cpp
+++ b/Source/DataModels/Animation/AnimationController.cpp
@@ -1,0 +1,95 @@
+#include "AnimationController.h"
+#include "Application.h"
+
+#include "Resources/ResourceAnimation.h"
+
+AnimationController::AnimationController(ResourceAnimation* resource)
+{
+	currentTime = 0.0f;
+	loop = true;
+    play = false;
+	this->resource = resource;
+}
+
+AnimationController::~AnimationController()
+{
+}
+
+void AnimationController::Play()
+{
+    play = true;
+}
+
+void AnimationController::Stop()
+{
+    play = false;
+}
+
+void AnimationController::Update()
+{
+    if (play)
+    {
+        float duration = resource->GetDuration();
+
+        currentTime += App->GetDeltaTime();
+        if (currentTime > duration)
+        {
+            if (loop)
+            {
+                currentTime = 0.0f;
+            }
+            else
+            {
+                currentTime = duration;
+            }
+        }
+    }
+}
+
+void AnimationController::GetTransform(char* name, float3& pos, Quat& rot)
+{
+    ResourceAnimation::Channel* channel = resource->GetChannel(name);
+    float currentSample = (currentTime * (channel->positions.size() - 1)) / resource->GetDuration();
+
+    int first = floor(currentSample);
+    int second = ceil(currentSample);
+
+    float3 firstPos = channel->positions[first];
+    float3 secondPos = channel->positions[second];;
+    Quat firstQuat = channel->rotations[first];
+    Quat secondQuat = channel->rotations[second];
+    float lambda = currentSample - first;
+
+    pos = Interpolate(firstPos, secondPos, lambda);
+    rot = Interpolate(firstQuat, secondQuat, lambda);
+}
+
+float3 AnimationController::Interpolate(const float3& first, const float3& second, float lambda)
+{
+	return first * (1.0f - lambda) + second * lambda;
+}
+
+Quat AnimationController::Interpolate(const Quat& first, const Quat& second, float lambda)
+{
+    Quat result;
+    float dot = first.Dot(second);
+
+    if (dot >= 0.0f) // Interpolate through the shortest path
+    {
+        result.x = first.x * (1.0f - lambda) + second.x * lambda;
+        result.y = first.y * (1.0f - lambda) + second.y * lambda;
+        result.z = first.z * (1.0f - lambda) + second.z * lambda;
+        result.w = first.w * (1.0f - lambda) + second.w * lambda;
+    }
+    else
+    {
+        result.x = first.x * (1.0f - lambda) - second.x * lambda;
+        result.y = first.y * (1.0f - lambda) - second.y * lambda;
+        result.z = first.z * (1.0f - lambda) - second.z * lambda;
+        result.w = first.w * (1.0f - lambda) - second.w * lambda;
+    }
+
+    result.Normalize();
+
+    return result;
+}
--- a/Source/DataModels/Animation/AnimationController.cpp
+++ b/Source/DataModels/Animation/AnimationController.cpp
@@ -3,20 +3,19 @@
 
 #include "Resources/ResourceAnimation.h"
 
-AnimationController::AnimationController(ResourceAnimation* resource)
+AnimationController::AnimationController() :
+    currentTime(0.0f), loop(true), play(false)
 {
-	currentTime = 0.0f;
-	loop = true;
-    play = false;
-	this->resource = resource;
 }
 
 AnimationController::~AnimationController()
 {
 }
 
-void AnimationController::Play()
+void AnimationController::Play(std::shared_ptr<ResourceAnimation> resource, bool loop)
 {
+    this->resource = resource;
+    this->loop = loop;
     play = true;
 }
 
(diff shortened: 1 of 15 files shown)


Source/DataModels/Scene/Scene.cpp
```diff
@@ -253,16 +257,20 @@ void Scene::DestroyGameObject(GameObject* gameObject)
...
 {
 	std::shared_ptr<ResourceModel> resourceModel = App->resources->RequestResource<ResourceModel>(model);
-	//resourceModel->Load();
-
+	std::vector<std::shared_ptr<ResourceAnimation>> animations = resourceModel->GetAnimations();
 	std::string modelName = App->fileSystem->GetFileName(model);
 
 	GameObject* gameObjectModel = CreateGameObject(modelName.c_str(), GetRoot());
+
+	if (animations.size() > 0)
```
MarcGonzalezMoratona, line 265:
```suggestion
	if (!animations.empty())
```

maybe it's easier to understand like this

Source/DataModels/Components/ComponentAnimation.h
```diff
@@ -2,17 +2,31 @@
...
 
+	void DrawBones(GameObject* parent, GameObject* lastNonExtraNode) const;
+
 	void SaveOptions(Json& meta) override;
 	void LoadOptions(Json& meta) override;
 
 private:
+	AnimationController* controller;
+	std::vector<std::shared_ptr<ResourceAnimation>> animations;
+	unsigned int animationIx;
```
MarcGonzalezMoratona, line 31: what does animationIx mean?

Source/DataModels/Components/ComponentAnimation.cpp
```diff
@@ -1,23 +1,100 @@
...
+	controller->Play(animations[animationIx], true);
+	controller->Stop();
 }
 
 void ComponentAnimation::Update()
 {
-	// Empty for now
+	controller->Update();
+
+	if (animations.size() > 0 && controller->GetPlay())
```
MarcGonzalezMoratona, line 45:
```suggestion
	if (!animations.empty() && controller->GetPlay())
```

Source/DataModels/Components/ComponentAnimation.cpp
```diff
@@ -1,23 +1,100 @@
...
+					static_cast<ComponentTransform*>(child->GetComponent(ComponentType::TRANSFORM));
+				transform->SetPosition(pos);
+				transform->SetRotation(float4x4(rot));
+			}
+		}
+		static_cast<ComponentTransform*>(owner->GetComponent(ComponentType::TRANSFORM))->UpdateTransformMatrices();
+	}
 }
 
 void ComponentAnimation::Draw()
```
MarcGonzalezMoratona, line 66: maybe this method can be const?

Source/DataModels/Components/ComponentAnimation.h
```diff
@@ -2,17 +2,31 @@
...
 	ComponentAnimation(const bool active, GameObject* owner);
 	~ComponentAnimation() override;
 
+	AnimationController* GetController();
+	void SetAnimations(std::vector<std::shared_ptr<ResourceAnimation>> animations);
+
 	void Update() override;
 	void Draw() override;
 
+	void DrawBones(GameObject* parent, GameObject* lastNonExtraNode) const;
```
MarcGonzalezMoratona, line 23: what is this lastNonExtraNode doing?

Source/DataModels/Animation/AnimationController.h
```diff
@@ -0,0 +1,43 @@
...
+public:
+	AnimationController();
+	~AnimationController();
+
+	void Play(std::shared_ptr<ResourceAnimation> resource, bool loop);
+	void Stop();
+	
+	void Update();
+
+	bool GetTransform(char* name, float3& pos, Quat& rot);
```
MarcGonzalezMoratona, line 21: some of this methods can also be const

Source/DataModels/Animation/AnimationController.h
```diff
@@ -0,0 +1,43 @@
...
+	bool GetPlay();
+	bool GetLoop();
+
+private:
+	float3 Interpolate(const float3& first, const float3& second, float lambda);
+	Quat Interpolate(const Quat& first, const Quat& second, float lambda);
+
+	float currentTime;
+	bool loop;
+	bool play;
```
MarcGonzalezMoratona, line 31: what do you think about renaming this variable to `isPlaying`?

Source/DataModels/Animation/AnimationController.cpp
```diff
@@ -0,0 +1,102 @@
...
+        pos = Interpolate(firstPos, secondPos, lambda);
+        rot = Interpolate(firstQuat, secondQuat, lambda);
+
+        return true;
+    }
+    
+    return false;
+}
+
+float3 AnimationController::Interpolate(const float3& first, const float3& second, float lambda)
```
MarcGonzalezMoratona, line 74:
```suggestion
float3 AnimationController::Interpolate(const float3& firstPos, const float3& secondPos, float lambda)
```

Source/DataModels/Animation/AnimationController.cpp
```diff
@@ -0,0 +1,102 @@
...
+{
+	return first * (1.0f - lambda) + second * lambda;
+}
+
+Quat AnimationController::Interpolate(const Quat& first, const Quat& second, float lambda)
```
MarcGonzalezMoratona, line 79:
```suggestion
Quat AnimationController::Interpolate(const Quat& firstQuat, const Quat& secondQuat, float lambda)
```

Source/DataModels/Animation/AnimationController.cpp
MarcGonzalezMoratona: if you accept the suggestion don't forget to change the body of the method

Source/DataModels/Animation/AnimationController.cpp
MarcGonzalezMoratona: if you accept the suggestion don't forget to change the body of the method

Source/DataModels/Animation/AnimationController.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+bool AnimationController::GetTransform(char* name, float3& pos, Quat& rot)
+{
+    ResourceAnimation::Channel* channel = resource->GetChannel(name);
+
+    if (channel)
+    {
+        float currentSample = (currentTime * (channel->positions.size() - 1)) / resource->GetDuration();
+
+        int first = floor(currentSample);
```
MarcGonzalezMoratona, line 56: I don't know if I understood the method propertly, but if you here are getting the position and rotation of a specific keyframe from a channel in the animation maybe we can call this `currentKeyframe` and the second can be the `nextKeyframe`

MarcGonzalezMoratona (commented):
Thanks for your contribution Miki! I added some comments to have a better understanding of the code and to help readability but I think everything is fine in terms of execution

Source/DataModels/Scene/Scene.cpp
MacXxs: It's actually the same 'cause the empty function for vectors calculates if size() > 0. So i'm doing this 'cause it should be faster

Source/DataModels/Components/ComponentAnimation.h
MacXxs: It's the index in the vector of animations. An fbx can have multiple animations (or clips if do you undersand it better)

Source/DataModels/Components/ComponentAnimation.cpp
MacXxs: Which one? UpdateTranmsforms? It can't be 'cause the it's changing the transforms of the GameObjects

Source/DataModels/Components/ComponentAnimation.h
MacXxs: It's the last node that assimp isn't adding on top of the ones of the fbx (the ones with weird names that Carlos told us not to show)

Source/DataModels/Scene/Scene.cpp
MarcGonzalezMoratona: I made a small research and I found this thread that can be interesting. It's not a big issue but it's good to know. Hope it helps!

https://stackoverflow.com/questions/743197/size-vs-empty-in-vector-why-empty-is-preferred

Source/DataModels/Components/ComponentAnimation.h
MarcGonzalezMoratona: Ahh now I understand. I didn't associate Ix to index

Source/DataModels/Animation/AnimationController.cpp
MacXxs: I would leave it as it is, the return type is self explenatory (this functions is actually a copy paste of the one from the ppt from Animation)

Source/DataModels/Animation/AnimationController.cpp
MacXxs: Same answer as before hahaha

Source/DataModels/Components/ComponentAnimation.cpp
MarcGonzalezMoratona: The draw method is only calling drawBones. I don't know is this method will have more logic in the future but maybe we can make it const or (just remove it and use the drawBones one)

MarcGonzalezMoratona: I see that it's created because it overrides the draw method from the parent, so maybe we can't remove it or make it const because the parent's definition is already done this way

Source/DataModels/Animation/AnimationController.cpp
MacXxs: It's not, they are actually the lower bound of the current sample (floor) and the higher bound (ceiling), they are just used to interpolate (but this can change in any time due to delta time variatons), i assigned them the name of first and second so It's clearer for the Interpolate functions that the teachers gave us on the PPT

Source/DataModels/Components/ComponentAnimation.h
MarcGonzalezMoratona: now I remember, thank you!

Source/DataModels/Animation/AnimationController.h
MacXxs: They cannot 'cause we need the values to be modified. The functionality of this function is to return the position and rotation needed to be applied, instead of creating two functions (one GetRotation and one GetPosition) with this we get both the rotation and position that need to be applied to the bone (if the boolean returned is true)

Source/DataModels/Components/ComponentAnimation.cpp
MacXxs: Yep if you were refering to the draw, we cannot make it const for the reason you just mentioned!

Source/DataModels/Animation/AnimationController.h
MacXxs: Yeah i can change it to isPlaying no worries!

Source/DataModels/Scene/Scene.cpp
MacXxs: In this case 'cause we are using a vector it should take constant time to to the std::size() calculus and it will take O(1). And 'cause we aren't going to change the container to a list it won't be O(n). But sure i can change it if you thing its more readable no problem!

MarcGonzalezMoratona: both implementations are good for me, use the one you feel more comfortable with. But at least I think we learnt something new from this

MarcGonzalezMoratona (approved):

Source/DataModels/Scene/Scene.cpp
MarcGonzalezMoratona: if you are going to use empty(), I think this one is missing

MacXxs: True!

> both implementations are good for me, use the one you feel more comfortable with. But at least I think we learnt something new from this

Yeah for sure!

Source/DataModels/Animation/AnimationController.cpp
MarcGonzalezMoratona: well, let's use the ppt names for the variables (although maybe they are not the most descriptive names)

Source/DataModels/Scene/Scene.cpp
MacXxs: > if you are going to use empty(), I think this one is missing

changed!

MarcGonzalezMoratona (approved):

sncarbo (approved):
I have not found any bugs and I believe that the implementation of the new structures you introduced are correct.

AuxPort66 (changes requested):
When you try to copy and object with the new component explodes because you need to add it inside the CopyComponent() function on line 302 in GameObject

PR #286, opened 2023-04-16
feat: functional scripting with hot reloading

gerard-queralt:
- Ability to create scripts with members modifiable through the graphical interface. **Scripts can't be created through the graphical interface, you need to do so before running the engine**.
- Scripts can be hot reloaded, meaning, you can edit the contents of the script while the engine is running. You can trigger it by saving (`Ctrl+S`) a script, even if there are no changes. **Hot reloading only works in source (.cpp) files, not headers (.h)**.
- **There's an empty project that's not used as of now. This is where user scripts should go, but we cut it for time**.


markmeliaa:
Why does this branch now have 120 commits? lol

gerard-queralt:
> Why does this branch now have 120 commits? lol

Because we inadvertedly merged [BasicPlayMode](https://github.com/Horizons-Games/Axolotl-Engine/tree/BasicPlayMode) into it, so it has not only the changes in that branch, but also the new stuff that wasn't in develop when [scripting-main](https://github.com/Horizons-Games/Axolotl-Engine/tree/scripting-main) was created. If scripting needs to be merged to develop before #278 I'll just try to cherrypick the changes so the only commits merged are actually from the scripting feature.

TLDR: don't worry about it :)

markmeliaa (approved):
Venga mergeamos esto o que?

ericdacal (approved):

PR #290, opened 2023-04-17
feat: Different Fragment Shaders integration

sncarbo:
Henceforth, the desired workflow (Default or Specular) can be selected via the MeshRenderer component in the ImGui.


sncarbo:
> There is a bug regarding skybox, please wait to approve until is fixed.

Done! Check the pr please whenever you are available.

MacXxs:
Hi Santi, I've fixed the bugs i've found:
* The ambient not working with the specular shader
* The specularity being broken for spot and point lights
* And some wrong calculus for the pbr

MacXxs (approved):

PaablooCH (commented):
Is this behaviour normal?
luces

The first one use specular shader, the second default

MacXxs:
Hi Santi! The specular color should only be applied once the button apply is pressed!

Vídeo sin título ‐ Hecho con Clipchamp

MacXxs:
I'm still working on the specular shader 'cause the specular is still not working correctly relatively to the smoothness slider, once i have it finished i'll push the changes!

MacXxs (dismissed):

PaablooCH (changes requested):
I have found two bugs:

1. in metallic i delete and add a normal texture:
call stack:
image
line (componentMeshRender): 
image

2. the second one happens when i use the specular and i apply some changes, it seems that when it loads some textures are corrupted (i suppose that the old one, because the pointer is pointing to null position and somehow it is not updated), sorry but i couldn't replicate it.

sncarbo:
> I have found two bugs:
> 
> 1. in metallic i delete and add a normal texture:
>    call stack:
>    image
>    line (componentMeshRender):
>    image
> 2. the second one happens when i use the specular and i apply some changes, it seems that when it loads some textures are corrupted (i suppose that the old one, because the pointer is pointing to null position and somehow it is not updated), sorry but i couldn't replicate it.

I'm removing the normal texture and adding it again (by pressing the Apply button or not) and I don't get the error you mentioned.

For the second bug I would ask you to add the code trace or an image because only with the description it is not understood.

MacXxs (approved):

PR #293, opened 2023-04-17
perf: inlined candidate methods to be inlined

xavi1111:


markmeliaa:
Hi! I have a question regarding this pr, is it truly beneficial for us? Will it improve our engine's performance? IIRC, adding header files into another header files was one of the practices that the teachers told us to avoid, as forward declartions are much less costly, and with this pr what we are doing is removing those forward declarations and adding header files into header files just to make a bunch of funcions inline, so idk if the tradeoff is positive or if we should ask the teachers about it or what.

Source/Engine.vcxproj.user
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="Current" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <ShowAllFiles>false</ShowAllFiles>
+    <ShowAllFiles>true</ShowAllFiles>
```
PaablooCH, line 4: turn this to false

PaablooCH (commented):
not compile in game mode and showAllFiles must be false. Also i'm with marc, first let's ask the teachers

PaablooCH (changes requested):
i forgot to request the changes

markmeliaa:
Then those functions that were inlined in this pr and that use unique pointers should go back to the .cpp files, shouldn't they? This way we would set the declaration of those classes back to forward declarations (no problem with those methods that were inlined and did not required to import any header file).

markmeliaa:
Also idk why this does not compile in DebugGame mode, but we should fix that also before merging this.

gerard-queralt:
> Then those functions that were inlined in this pr and that use unique pointers should go back to the .cpp files, shouldn't they? 

Yes! Xavi is already working on it.

> but we should fix that as well before merging this.

Also yes, in fact I think it's configured so a PR can't be merged unless it passes the pipeline.

markmeliaa:
> Also yes, in fact I think it's configured so a PR can't be merged unless it passes the pipeline.

Oh, that's nice, I didn't know it!

xavi1111:
Gathering your feedback I tried to keep as many inline methods as possible without having too much new imports on header files (I think only three are left now). I also find it a good idea to ask the teachers if the imports left are beneficial to the project or we should revert the changes. Thank you for your fast responses.

Source/DataModels/Scene/Scene.h
```diff
@@ -115,6 +116,11 @@ class Scene
 	std::vector<GameObject*> nonStaticObjects;
 };
 
+inline bool Scene::IsInsideACamera(const AABB& aabb) const
```
markmeliaa, line 119: Why did you decide to inline this method? It does not seem trivial.

markmeliaa:
> Gathering your feedback I tried to keep as many inline methods as possible without having too much new imports on header files (I think only three are left now). I also find it a good idea to ask the teachers if the imports left are beneficial to the project or we should revert the changes. Thank you for your fast responses.

Thanks to you! The remaining includes are from external libraries if I did not miss anything, so I see it okay-ish to leave those includes in the header files.

Source/DataModels/Scene/Scene.h
gerard-queralt: I agree with you that this doesn't seem trivial. However, it's worth noting that the `inline` keyword serves only as a hint to the compiler, so it may ignore it and not actually inline it. Here's the [documentation](https://en.cppreference.com/w/cpp/language/inline). Relevant part:

> Since this meaning of the keyword `inline` is non-binding, compilers are free to use inline substitution for any function that's not marked `inline`, and are free to generate function calls to any function marked `inline`

This is to say that even if we inline functions that we shouldn't, in theory the compiler should be smart enough to ignore us :P

markmeliaa: Then if it's possible that the compiler will ignore this as `inline` why don't we turn this into how it was before (non inlined)? This way we could avoid one header being included into another header again :)

xavi1111: The function called inside is not trivial but that does not mean that the method encapsulating it cannot be inlined to save the cost of calling the initial function. This way we should save one function call (Theoretically) if the compiler deems appropiate to really inline the method. That said you are probably right that this inline is not worth the import. So maybe it's better to revert this change. 

Sources to the first assesment : 
- https://stackoverflow.com/questions/27904237/does-it-makes-sense-to-define-an-inline-function-which-makes-calls-to-other-func
- https://stackoverflow.com/questions/67967657/usage-of-static-inline-to-functions-which-calls-other-larger-functions-in-c

xavi1111: I reverted the changes to avoid the import. Thanks again for the fast input.

markmeliaa (dismissed):

gerard-queralt (dismissed):
Thanks for your work Xavi, everything looks good to me! I left some comments, but as I said it's not really related to this PR, so we can do this later down the line.

markmeliaa:
Seems like some conflicts arised due to the previous merges to develop, those should be fixed before merging this pr too.

DADSILVER (changes requested):
A few things.

Source/DataModels/Program/Program.h
DADSILVER, line 1: None of these functions are inline as they contain opengl functions.

DADSILVER (changes requested):
A few things.

Source/DataModels/Program/Program.h
xavi1111: Hi could you elaborate in this statement. I don't understand what you are referring to. Could it be something similar to what was discussed with Mark? or is it something else?. Thank you in advance.

gerard-queralt (approved):

Source/DataModels/Program/Program.h
DADSILVER: Sorry for the delay, those functions are internally from OpenGL so we don't know what they are doing. Inline should only be those "simple" methods. Therefore those methods that call other more complex functions should not be inline.
https://www.mygreatlearning.com/blog/inline-functions-in-cpp/#:~:text=Inline%20functions%20are%20commonly%20used,definition%20of%20the%20called%20function.

gerard-queralt: If I may interject, I think Xavi was referring to a conversation he had with Mark in this same PR adressing a very similar concern: https://github.com/Horizons-Games/Axolotl-Engine/pull/293#discussion_r1170345909
TLDR: Inlining a call to another method is perfectly fine.

DADSILVER (approved):

PR #295, opened 2023-04-18
refactor: Update and Draw not methods of Component, use interfaces instead

gerard-queralt:


gerard-queralt:
It's mainly from a design standpoint. If you check the methods I deleted, the majority of Draws and Updates were empty. So then, why force everyone to overload two functions that only a few use? And even if we don't make them pure virtual (so don't force an overload), each instance would still "carry" an empty method (which is also added to the virtual table). That's why I thought using these interfaces was a better idea.
I heard that you guys were concerned about how this relates to virtual tables, but I wasn't able to find anything on the internet on whether multiple inheritance has any impact on performance. Though, to be fair, it's not like inheritance has any noticeable impact on performance anyway. That said, I have no problem with asking the teachers, of course.
Also, I can't really cherrypick this, since the vector I added here uses the new interface, but I could of course replicate this change in another branch.

AuxPort66 (approved):

Vlukkee (approved):

gerard-queralt:
I'll merge this since it's not only needed, it should also fix the error reported in #307. We can discuss after the VS my approach and your alternatives.

PR #296, opened 2023-04-18
UI rework

AuxPort66:
- **perf:** Moved the vertex vector and image VAO to ModuleUI to have a general mesh for all UI elements.
- **refactor:** ModuleCamera now has the orthogonal projection.
- **fix:** The raycast did not check if the gameObject had transform3D and it exploded when detecting a UI component with transform2D (I think this was already there, maybe a merge removed it, I don't know)
- **fix:** The canvas factor was not updated when resizing the window scene.
- **feat:** The shader and the image draw have been updated to be able to draw translucent elements
- **feat:** For the above, a new UniformFloat4 has also been added to the program


Source/DataModels/Components/UI/ComponentButton.h
```diff
@@ -22,19 +22,19 @@ class ComponentButton :
 	void SetClicked(bool clicked);
 	void SetHovered(bool hovered);
 
-	void SetColorClicked(float3 colorClicked);
-	void SetColorHovered(float3 colorHovered);
+	void SetColorClicked(const float4 colorClicked);
+	void SetColorHovered(const float4 colorHovered);
```
gerard-queralt, line 26:
```suggestion
	void SetColorHovered(const float4& colorHovered);
```

Same for the method above.

Source/DataModels/Components/UI/ComponentImage.h
```diff
@@ -19,29 +19,25 @@ class ComponentImage :
...
 	std::shared_ptr<ResourceTexture>& GetImage();
-	float3& GetColor();
-	float3 GetFullColor();
-	void SetImage(const std::shared_ptr<ResourceTexture>& image);
+	float4 GetColor() const;
 
-	void LoadVBO();
-	void CreateVAO();
+	void SetImage(const std::shared_ptr<ResourceTexture>& image);
+	void SetColor(const float4 color);
```
gerard-queralt, line 28:
```suggestion
	void SetColor(const float4& color);
```

Source/DataModels/Program/Program.h
```diff
@@ -16,6 +17,7 @@ class Program
 
 	void BindUniformFloat4x4(const std::string& name, const float* data, bool transpose);
 	void BindUniformFloat3(const std::string& name, const float3 data);
+	void BindUniformFloat4(const std::string& name, const float4 data);
```
gerard-queralt, line 20:
```suggestion
	void BindUniformFloat4(const std::string& name, const float4& data);
```
Should do the same for the other methods, I guess it slipped by the original PR.

Source/Modules/ModuleCamera.h
```diff
@@ -47,11 +47,15 @@ class ModuleCamera : public Module
 	int GetSelectedPosition();
 	void SetSelectedPosition(int newSelected);
 
-	
+	const float4x4  GetOrthoProjectionMatrix() const;
```
gerard-queralt, line 50: Don't see the point in returning a const **copy** of the member. I would either return a non-const copy or a const reference, up to you.

Source/Modules/ModuleUI.h
```diff
@@ -19,5 +19,18 @@ class ModuleUI : public Module
 
 	void DrawChildren(GameObject* gameObject);
 	void RecalculateCanvasSizeAndScreenFactor();
+	void LoadVBO();
+	void CreateVAO();
+	const unsigned int GetQuadVAO() const;
```
gerard-queralt, line 24: Same as previous comment, don't think there's any point in returning a const primitive.

gerard-queralt (commented):
Thank you for your work Sara. Left a few unimportant comments.

gerard-queralt (approved):

Vlukkee (approved):

PR #300, opened 2023-04-19
feat: Scripts are "fully" funcional and accept different types of variables

markmeliaa:
Now, scripts accept not only floats, but also strings, booleans and GameObjects (that should suffice for now, we'll add the different types of Components in the future, by now we can access to any Component instantiating a GameObject in the Script and calling `GetComponent(type)` on it).

Also, the ScriptComponentWindow was improved, you can load `DefaultScript` to see how the different types of variables are shown (this window also accepts Drag and Drop of GameObjects, so you can set any GameObject of the hierarchy into the script).

In the `HelloWorldScript` you can find a couple trials that everything works as intended (a couple of bugs arised, and they should be reported but I think that they have nothing to do with this pr so they shouldn't be addressed here).

That's all, any doubt of how Scripts work please feel free to ask!


Game/Scripts/HelloWorldScript.h
```diff
@@ -0,0 +1,29 @@
...
+
+	void Init() override {};
+	void Start() override {};
+	void PreUpdate(float deltaTime) override {};
+	void Update(float deltaTime) override;
+	void PostUpdate(float deltaTime) override {};
+	void CleanUp() override {};
+
+	std::string GetHelloWorld() const;
+	void SetHelloWorld(const std::string helloWorld);
```
gerard-queralt, line 21: Thank you for your work Mark. I'll take a better look at your changes later, but I noticed something here:
```
	void SetHelloWorld(const std::string& helloWorld);
```
If this script is to serve as an example, it better set a good one! :P

markmeliaa: Whoops, thanks for noticing!

Game/Settings/WindowsStates.conf
DADSILVER, line 1: Should be deleted.

Game/Assets/Unknown
DADSILVER, line 1: Should be deleted

DADSILVER (approved):

Game/Scripts/DefaultScript.cpp
```diff
@@ -21,3 +35,33 @@ void DefaultScript::SetValue(float value)
...
+}
+
+bool DefaultScript::GetCheck() const
+{
+	return check;
+}
+
+void DefaultScript::SetCheck(bool verify)
+{
+	this->check = verify;
```
PaablooCH, line 66: `this` unnecessary

PaablooCH (changes requested):
all seems good, I request just a little change

Game/Scripts/DefaultScript.cpp
markmeliaa: Done, ty for noticing!

markmeliaa:
> Hi Mark! @Vlukkee reported that the engine crashes when deleting a component script. After checking, tuns out I forgot to check that the pointer is not null after the "remove" button has been processed, like it's done in every other window ([MeshRenderer](https://github.com/Horizons-Games/Axolotl-Engine/blob/scripting/component-script-window/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentMeshRenderer.cpp#L41), for example). Could you please fix this? Something like
> 
> ```c++
> ComponentScript* script = static_cast<ComponentScript*>(this->component);
> 
> if (script)
> {
> 	std::vector<std::string> constructors = App->scriptFactory->GetConstructors();
> 	.........
> ```
> 
> Thank you in advance and sorry for the trouble!

Done!

PaablooCH (approved):

gerard-queralt (approved):

PR #262, opened 2023-04-02
feat: scripting new project

Piterga12:

PR #301, opened 2023-04-21
feat: scripting added to the engine

gerard-queralt:
List of pull requests about scripting:

-  #262
    - Currently not used due to issues with the library, but idealy will be in the future
- #264
    - Not merged, but good insight on how to access code from another project
- #265
    - Not merged directly, but merged later with @ericdacal's branch in #286
- #286
    - Integration of a library that does hot reloading for us with our idea for scripting
- #300
   - Added more types of variables, example scripts and improved the component window. Also fixed some issues unnoticed in the previous PR

Authors:
- @Piterga12 
- @AlexisDieu 
- @totisama 
- @ericdacal 
- @markmeliaa 
- @gerard-queralt


gerard-queralt:
Depends on #278

Dedado3000 (approved):
I don't see any problem in the parts i worked on.

In the other Part, good implementation of scripting @ericdacal  and @gerard-queralt congrats

markmeliaa:
Yessss, we got this! Nice work everyone 💯

markmeliaa (dismissed):
Yessss, we got this! Nice work everyone 💯

ericdacal (approved):

AuxPort66 (dismissed):

Piterga12 (approved):
Everything works fine!

markmeliaa (dismissed):

Source/Modules/ModuleInput.cpp
```diff
@@ -6,6 +6,7 @@
 #include "ModuleCamera.h"
 #include "ModuleScene.h"
 #include "ModuleEditor.h"
+#include "ModulePlayer.h"
```
PaablooCH, line 9: why is only added this include? i mean it's the only change, you didn't add a use

Source/Modules/ModuleScene.cpp
PaablooCH, line 1: there are a couple ifs that can be merge into one, lines: 95-97, 119-121 and137-139

Source/Modules/ModuleUI.cpp
```diff
@@ -68,7 +68,7 @@ update_status ModuleUI::Update()
 	glOrtho(0, width, height, 0, 1, -1);
 	glMatrixMode(GL_MODELVIEW);
 
-	App->camera->GetCamera()->GetFrustum()->SetOrthographic(width, height);
+	App->camera->GetCamera()->GetFrustum()->SetOrthographic((float)width, (float)height);
```
PaablooCH, line 71: it is better to use static cast instead of c cast

Source/Scripting/IScript.h
```diff
@@ -0,0 +1,132 @@
...
+};
+
+inline void IScript::SetGameObject(GameObject* owner)
+{
+	this->owner = owner;
+}
+
+inline void IScript::SetApplication(Application* app)
+{
+	this->App = app;
```
PaablooCH, line 59: i guess `this` is unnecessary

PaablooCH (changes requested):
good job all of you, I notice a couple of things
PS: it doesn't compile in debugGame

Source/Scripting/ScriptFactory.h
```diff
@@ -0,0 +1,38 @@
...
+	bool IsCompiled();
+	std::vector<const char*> GetConstructors();
+private:
+	void IncludeDirs();
+
+	//Runtime Systems
+	ICompilerLogger* pCompilerLogger;
+	IRuntimeObjectSystem* pRuntimeObjectSystem;
+
+	std::vector<std::pair<ObjectId, IScript*>> allScripts;
```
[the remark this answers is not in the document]
gerard-queralt: I'm not sure if they are, to be honest. But this container is never used as a dictionary (we never search a Script by it's ObjectID). What we do is, when a hot reload happens, we iterate this collection, request the new Script instance to the library through that ObjectID, and perform the necessary updates. That's why it's a vector of pairs instead of a map. Let me know if you're still not sure about this!

Source/Modules/ModuleInput.cpp
gerard-queralt: Maybe it got mixed up in some merge. Thanks for noticing!

DADSILVER (commented):
Many thanks to all of you who have worked on this branch. Just one thing.

Game/Scripts/DefaultScript.cpp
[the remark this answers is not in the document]
gerard-queralt: Ideally yes, but the library we use doesn't seem to detect changes in the header so it doesn't trigger hot reloading. This means that, if you were to change something in an inlined method, you would need to save the source file anyway. To make it less strange for the user, we opted for this solution.
We still need to figure out if this is a limitaion of the library or something we configured wrong, but decided to leave it for after the vertical slice.
Hopefully this answers your question!

gerard-queralt:
Thank you for your feedback @PaablooCH and @DADSILVER. Tried to apply your suggestions, take a look when you can and resolve the comment if you are satisfied.

PaablooCH (dismissed):

DADSILVER (dismissed):

Dedado3000 (approved):

markmeliaa (approved):

DADSILVER (approved):

PR #303, opened 2023-04-21
feat : Efficient rendering persistent buffer and double buffer

AlexisDieu:
- Add the persistent buffer in order to have better performance
- Create double buffer to manage a change the buffer each frame for more performance and better synchronization

### DIFF
--- a/Source/DataModels/Batch/BatchManager.cpp
+++ b/Source/DataModels/Batch/BatchManager.cpp
@@ -68,5 +68,12 @@ GeometryBatch* BatchManager::CheckBatchCompatibility(const ComponentMeshRenderer
 
 void BatchManager::DrawBatch(GeometryBatch* batch, const std::vector<ComponentMeshRenderer*>& componentsToRender)
 {
-	batch->BindBatch(componentsToRender);
+	if (batch->dirtyBatch)
+	{
+		batch->ClearBuffer();
+		batch->CreateVAO();
+		batch->dirtyBatch = false;
+
+	}
+		batch->BindBatch(componentsToRender);
 }
--- a/Source/DataModels/Batch/BatchManager.cpp
+++ b/Source/DataModels/Batch/BatchManager.cpp
@@ -71,7 +71,7 @@ void BatchManager::DrawBatch(GeometryBatch* batch, const std::vector<ComponentMe
 	if (batch->dirtyBatch)
 	{
 		batch->ClearBuffer();
-		batch->CreateVAO();
+		batch->UpdateBuffer();
 		batch->dirtyBatch = false;
 
 	}
--- a/Source/DataModels/Batch/GeometryBatch.cpp
+++ b/Source/DataModels/Batch/GeometryBatch.cpp
@@ -28,12 +28,19 @@ createBuffers(true), reserveModelSpace(true), flags(0), defaultMaterial(new Reso
 	glGenVertexArrays(1, &vao);
 	glGenBuffers(1, &ebo);
 	glGenBuffers(1, &indirectBuffer);
-	glGenBuffers(1, &transforms);
+	for (int i = 0; i < DOUBLE_BUFFERS; i++)
+	{
+		glGenBuffers(1, &transforms[i]);
+	}
 	glGenBuffers(1, &verticesBuffer);
 	glGenBuffers(1, &textureBuffer);
 	glGenBuffers(1, &normalsBuffer);
 	glGenBuffers(1, &tangentsBuffer);
-	glGenBuffers(1, &materials);
+	//for (int i = 0; i < DOUBLE_BUFFERS; i++)
+	//{
+	//	glGenBuffers(1, &materials[i]);
+	//}
+		glGenBuffers(1, &materials);
 	program = App->program->GetProgram(ProgramType::MESHSHADER);
 }
 
@@ -49,7 +56,13 @@ GeometryBatch::~GeometryBatch()
 	instanceData.clear();
 	delete defaultMaterial;
 	CleanUp();
-	glDeleteSync(gSync);
+	for (int i = 0; i < DOUBLE_BUFFERS; i++)
+	{
+		if (gSync[frame])
+		{
+			glDeleteSync(gSync[frame]);
+		}
+	}
 }
 
 void GeometryBatch::FillBuffers()
@@ -108,6 +121,7 @@ void GeometryBatch::FillMaterial()
 {
 	std::vector<Material> materialToRender;
 	materialToRender.reserve(instanceData.size());
+	//Material* materials = (Material*)materialData[frame];
 	for (int i = 0; i < instanceData.size(); i++)
 	{
 		int materialIndex = instanceData[i];
@@ -143,6 +157,7 @@ void GeometryBatch::FillMaterial()
 		}
 
 		materialToRender.push_back(newMaterial);
+		//materials[i] = newMaterial;
 		materialData[i] = newMaterial;
 	}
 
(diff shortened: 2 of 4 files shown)


Source/DataModels/Batch/BatchManager.cpp
```diff
@@ -68,5 +68,12 @@ GeometryBatch* BatchManager::CheckBatchCompatibility(const ComponentMeshRenderer
 
 void BatchManager::DrawBatch(GeometryBatch* batch, const std::vector<ComponentMeshRenderer*>& componentsToRender)
 {
-	batch->BindBatch(componentsToRender);
+	if (batch->dirtyBatch)
```
PaablooCH, line 71: why you change the flow of the batch doing this? you are deleting buffers and recreating them, maybe i miss something, but i prefer our previous version

Source/DataModels/Batch/GeometryBatch.h
```diff
@@ -98,6 +105,16 @@ class GeometryBatch
 	unsigned int numTotalIndices;
 	unsigned int numTotalFaces;
 
+
+	GLsync gSync[DOUBLE_BUFFERS] = { nullptr,nullptr };
+
+	GLbitfield mapFlags = GL_MAP_WRITE_BIT |GL_MAP_PERSISTENT_BIT | GL_MAP_COHERENT_BIT;
+	GLbitfield createFlags = mapFlags | GL_DYNAMIC_STORAGE_BIT;
```
PaablooCH, line 112: At this two variables to the constructor

Source/DataModels/Batch/GeometryBatch.h
```diff
@@ -98,6 +105,16 @@ class GeometryBatch
...
+	float4x4* transformData[DOUBLE_BUFFERS];
+	//Material* materialData[DOUBLE_BUFFERS];
+	Material* materialData;
```
PaablooCH, line 115: why this is needed? i understand the purpose of having the transformData, but material is something that wouldn't change

Source/DataModels/Batch/GeometryBatch.cpp
```diff
@@ -369,16 +424,15 @@ void GeometryBatch::BindBatch(const std::vector<ComponentMeshRenderer*>& compone
...
 	}
-	
 	glBindBuffer(GL_DRAW_INDIRECT_BUFFER, indirectBuffer);
-	glBufferSubData(GL_DRAW_INDIRECT_BUFFER, 0, drawCount * sizeof(Command), &commands[0]);
-	
-	glNamedBufferSubData(transforms, 0, modelMatrices.size() * sizeof(float4x4), &modelMatrices[0]);
-
+	glBufferData(GL_DRAW_INDIRECT_BUFFER, commands.size() * sizeof(Command), &commands[0], GL_DYNAMIC_DRAW);
+	//glBufferSubData(GL_DRAW_INDIRECT_BUFFER, 0, drawCount * sizeof(Command), &commands[0]);
+	program = App->program->GetProgram(ProgramType::MESHSHADER);
```
PaablooCH, line 430: program is a class variable initializated in constructor, this line can be deleted

Source/DataModels/Batch/GeometryBatch.h
PaablooCH: in the pdf only talks about model matrix
image
so it is not necessary

PaablooCH (changes requested):
i leave you some correction but the most important is the creation of dirtyBatch, i would like to talk about it

Source/DataModels/Batch/BatchManager.cpp
AlexisDieu: Yeah exactly because I am using persistent buffer now that is immutable so i need to re create the batch every time we add or delete a component and this solution is to do that and be able to re do all persistent buffer correctly, it's not really costly because it's only done on Engine mode and it's the only way to redo the size of a persistant buffer

Source/DataModels/Batch/GeometryBatch.h
AlexisDieu: It's only talking of model in the pdf but do a persistent buffer of the materials is not that bad I think but I am open to talk about it to know which one is better, in did i tried also to do double buffer for material but is really not a good idea cause of the instance that get broken really quickly when adding or deleting but yet make a persistent buffer is not that bad (also because I redo the batch when adding or removing and the way we delete material by remplacing with a default one fit correctly with the persistnt buffer idea)

Source/DataModels/Batch/GeometryBatch.cpp
AlexisDieu: True I had a bug for a moment and I put that to test but forgot to delete it

Source/DataModels/Batch/GeometryBatch.h
```diff
@@ -65,8 +66,10 @@ class GeometryBatch
 
 	bool CleanUp();
 
-private:
+	bool dirtyBatch;
 
+private:
+	enum { DOUBLE_BUFFERS = 2 };
```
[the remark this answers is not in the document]
AlexisDieu: I think enumeration in this case may be considered better practice because it provides more type safety because it's hard typed and can only use the value defined in the enumeration but I can change it I you think it's better +enumaration can be use in switch statements and i use this value a bunch of times and in some case switch can be an idea !

Source/DataModels/Batch/GeometryBatch.cpp
```diff
@@ -300,6 +345,15 @@ void GeometryBatch::DeleteMaterial(ComponentMeshRenderer* componentToDelete)
 
 void GeometryBatch::BindBatch(const std::vector<ComponentMeshRenderer*>& componentsToRender)
 {
+	if (frame == 0)
```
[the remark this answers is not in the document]
AlexisDieu: i tried but it's not working because frame is not a boolean but a int (because it can go to 2 if we will use triple buffer in futur) what i can do it's try to calculate it with a modulo but i don't like this idea , it will be like 	frame = (frame + 1) % DOUBLE_BUFFERS;

Source/DataModels/Batch/GeometryBatch.h
AlexisDieu: Well in fact i found an issue when deleting a material only and persitant buffer so I keep the idea in mind but I am gonna use normal SSBO once again I guess

Source/DataModels/Batch/GeometryBatch.h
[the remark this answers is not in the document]
PaablooCH: now is just a number that is repeat it a bunch of times, constexpr could be also an option because is fast in compilation time. We can leave this as a tie between us and if someone wants to aport something can untie 👍

PaablooCH (changes requested):
new things:
PS: check the previous one also thx

Source/DataModels/Batch/BatchManager.cpp
AlexisDieu: In fact I did it lik that because i found that bindBatch had already a lot of if() exception and one more will make it maybe less clear to read (i also thought about move the creation of command and transforms in a function then call them in bind batch instead of doing it there) but you right it's just it was the easiest way for me and more visually clear ! For the redo of buffer you right but as i said before I would like to do a function for transforms and command and when it's dirty I call it and redo it and not all buffer because you right we didn't need to recreate vertices and stuff

sncarbo:
In terms of functionality, I don't see the Skybox and the engine freezes when I try to import something. I don't know if it's a problem with my computer or not, but I want to check it. It compiles in all modes.

In terms of code, good use of OpenGL concurrency with glFenceSync, glDeleteSync and glClientWaitSync. Good use of glBufferStorage and glBindBufferRange. There are some cases where you don't apply the programming rules, such as the brackets in the line break after an if statement.

Apply the changes that Pablo asked you to make and, when I check what I said before, I will approve the pull request. 

Very good work from my point of view.

PR #307, opened 2023-04-23
fix: don't update scripts in normal component's update flow

markmeliaa:


gerard-queralt:
`GameObject::Update` should never be called outside Game/Play mode in the first place. That's the error, not the fact that the scripts are also updated. We should take a look at that.

gerard-queralt:
The error is most definetly the fact that the transform is calling Update. Check the changes in #295, should fix this.

markmeliaa:
Right, thanks for noticing that and thanks for the quick fix, closing this!

PR #311, opened 2023-04-24
feat[Gameplay]: Player move, jump, dash, crouch and run added

DADSILVER:
The basic movements of the player are implemented to be able to pass the level.

- isOnPlayMode is used to manage when scripts are activated.
- The module player is left only as the upload and download manager of the player.
- A script is created with the player's main mobility. In addition, a dash is added as an ability, the possibility of double jumping and crouching.

If you want to test the branch you need to run it in debugGame mode by putting the CantinaWithMovScript scene and the scene module in the corresponding lib folder.


markmeliaa (approved):
Thanks for your work David, everything works as intended in Game mode.

markmeliaa:
After trying it a bit more, maybe the jump feels weird, but maybe that's more a gravity issue? idk, the player falls really slowly

PR #315, opened 2023-04-24
Feat: Wwise implementation and Audio

UriMurcia:
Here is a first approach of the audio module, components and sound effects in our engine for the VS2.
Main New classes:

- Audio Source Component
- Audio Listener Component
- Module Audio

To test this feature is necessary to have an audio listener in the scene, then, create an audio source in the same scene and call its funcion of Post Event from any script.

To use the different audios include "Auxiliar/Audio/AudioData.h" and use its audio:: namespace


AuxPort66 (changes requested):
Missing serialization of both components =/

AuxPort66 (dismissed):

AuxPort66 (approved):

gerard-queralt (approved):

PR #318, opened 2023-04-25
Feat: State machine

AuxPort66:
- StateMachine Importer
- New StateMachine Resource with Transitions, Conditions, Paremeters and States
- New StateMachineEditor (You need to select a statemachine inside some ComponentAnimation to edit the stateMachine)
- Update ComponentAnimation with StateMachine and all the Rework need it to refresh between animations


MacXxs (approved):
Works perfectly 👍🏼

PR #319, opened 2023-04-25
feat: integrate Animation into the engine

AuxPort66:
- Resource Animation and StateMachine
- Update ResourceModule and ResourceMesh (Bones)
- Update ModuleImporter and MeshImporter
- AnimationImporter and StateMachineImporter
- ComponentAnimation and ControllerAnimation

To put animations you have to create a new StateMachine (You can do this from the file browser there is a new button)

Then Import your model with drag & drop and in the parent gameObject create a componentAnimation where to place your stateMachine, hit the edit statemachine button and with the right click you can start creating transitions and states. (move with Left and Middle click)

To select the transitions you have to click on the triangle, and you can add conditions that you can then call from a script by making **SetParameter(Name, new value)** to the corresponding animation component. Component Animation will check at all times if any of its transitions are fulfilled and will apply the resource of the actual State.


markmeliaa (approved):

Source/FileSystem/ModuleResources.cpp
```diff
@@ -187,30 +208,30 @@ void ModuleResources::AddResource(std::shared_ptr<Resource>& resource, const std
 
 void ModuleResources::DeleteResource(const std::shared_ptr<EditorResourceInterface>& resToDelete)
```
gerard-queralt, line 209: Any reason this is commented? Doesn't look right to me.

Source/FileSystem/ModuleResources.cpp
```diff
@@ -292,6 +319,14 @@ void ModuleResources::ReimportResource(UID resourceUID)
 		std::shared_ptr<ResourceMaterial> materialResource = std::dynamic_pointer_cast<ResourceMaterial>(resource);
 		ReImportMaterialAsset(materialResource);
 	}
+	if (resource->GetType() == ResourceType::StateMachine)
+	{
+		std::shared_ptr<ResourceStateMachine> stateMachineResource = std::dynamic_pointer_cast<ResourceStateMachine>(resource);
+		char* saveBuffer = {};
```
gerard-queralt, line 325: I don't remember if this buffer needs to be deleted after, could you double check just to make sure? It might be allocated in some of the functions below.

Dedado3000 (approved):

Source/Modules/ModuleEditor.h
```diff
@@ -43,6 +48,8 @@ class ModuleEditor : public Module
 	std::vector<std::unique_ptr<EditorWindow> > windows;
 	std::unique_ptr<WindowMainMenu> mainMenu = nullptr;
 	std::unique_ptr<WindowDebug> debugOptions = nullptr;
+	WindowStateMachineEditor* stateMachineEditor = nullptr;
```
gerard-queralt, line 51: I think this pointer is never deleted. Prefer using unique pointers.

gerard-queralt (commented):
As I said, I won't be requesting changes so we can merge this today, but I'd appreciate if we could address my comments as soon as possible.

Source/DataModels/Windows/EditorWindows/ImporterWindows/WindowResourceInput.h
gerard-queralt, line 1: Forgot to mention it, but looks like this window is only intended for state machines. We can probably use a more descriptive name.

MacXxs (approved):

Source/DataModels/Components/ComponentAnimation.cpp
```diff
@@ -107,12 +107,20 @@ void ComponentAnimation::Update()
 				static_cast<ComponentTransform*>(owner->GetComponent(ComponentType::TRANSFORM))->UpdateTransformMatrices();
 			}
 		}
+		lastState = actualState;
+	}
+	else
+	{
+		lastState == NON_STATE;
```
gerard-queralt, line 114: This is most likely wrong
```suggestion
		lastState = NON_STATE;
```

Source/DataModels/Components/ComponentAnimation.cpp
```diff
@@ -51,10 +52,9 @@ void ComponentAnimation::SetStateMachine(const std::shared_ptr<ResourceStateMach
 
 void ComponentAnimation::Update()
 {
-
 	if (stateMachine)
 	{
-		if (actualState == 0) //Entry State 
+		if ((actualState == 0) & (lastState == NON_STATE)) //Entry State
```
gerard-queralt, line 57: I'm guessing you don't want a bit-wise comparison...
```suggestion
		if ((actualState == 0) && (lastState == NON_STATE))
```

Dedado3000 (approved):

Dedado3000 (approved):

MacXxs (dismissed):

markmeliaa (dismissed):

Source/DataModels/Components/ComponentAnimation.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+			Transition* checkedTransition = CheckTransitions(state);
+			if (checkedTransition != nullptr)
+			{
+				nextState = stateMachine->GetIdState(*checkedTransition->destination);
+			}
+		}
+		else 
+		{
+			//Pasamos a la nueva state con tal de tener algo
```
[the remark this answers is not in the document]
AuxPort66: Better eliminate that comment, that's not a real comment

Source/DataModels/Components/ComponentAnimation.cpp
```diff
@@ -0,0 +1,268 @@
...
+					break;
+				case ConditionType::FALSECONDITION:
+					conditionCheck = (std::get<bool>(value) == false);
+					break;
+				default:
+					break;
+				}
+			}
+
+			if (!conditionCheck) break;
```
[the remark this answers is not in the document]
AuxPort66: Yes that made sense before, with the last change it was left forgotten. I'm gonna do that

Source/DataModels/Resources/ResourceStateMachine.cpp
```diff
@@ -0,0 +1,161 @@
...
+		SelectConditionParameter(defaultParameters.begin()->first, condition);
+	}
+	transitions[transition].conditions.push_back(std::move(condition));
+}
+
+void ResourceStateMachine::SelectConditionParameter(const std::string& parameter, Condition& condition)
+{
+	condition.parameter = parameter;
+	const auto& it = defaultParameters.find(parameter);
+	switch (it->second.first)
```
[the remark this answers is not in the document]
AuxPort66: Currently and theoretically (No one else should use this function), the only ones that use this function is a for of parameters, so it's not possible to send a parameter that does not exist. And is used too in AddCondition but that check the parameter size, it will be the same on cost but AddCondition it's less used.

So we don't need a check there.

Source/DataModels/Windows/EditorWindows/ImporterWindows/WindowResourceInput.h
AuxPort66: Suggestions? It accepts any type of resource in statemachine state.

Source/DataModels/Resources/ResourceAnimation.h
```diff
@@ -38,7 +38,7 @@ struct Channel
 	double GetDuration() const;
 
 	void SetDuration(double duration);
-	void SetChannels(std::unordered_map<std::string, ResourceAnimation::Channel*> channels);
+	void SetChannels(const std::unordered_map<std::string, ResourceAnimation::Channel*> channels);
```
gerard-queralt, line 41: Whoops, this is still wrong, my bad.
```suggestion
	void SetChannels(const std::unordered_map<std::string, ResourceAnimation::Channel*>& channels);
```

Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -582,6 +590,27 @@ Component* GameObject::GetComponent(ComponentType type) const
 	return nullptr;
 }
 
+GameObject* GameObject::FindGameObject(const std::string& name)
+{
+	if (this->name == name)
+	{
+		return this;
+	}
+	else
```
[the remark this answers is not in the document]
AuxPort66: I think I read better the logic with the else even if it's not necessary.

Source/DataModels/Resources/ResourceStateMachine.cpp
```diff
@@ -0,0 +1,161 @@
...
+{
+}
+
+int ResourceStateMachine::GetIdState(const State& state) const
+{
+	for (int i = 0; i < states.size(); i++)
+	{
+		if (states[i]->id == state.id) return i;
+	}
+	return -1;
```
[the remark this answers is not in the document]
AuxPort66: For an int I think it's a bit overkill and that a GetIdStates returns a -1 if it doesn't find it seems the most normal given the name, like when they return an end() after a find

Source/DataModels/Components/ComponentAnimation.cpp
```diff
@@ -0,0 +1,268 @@
...
+	if (resourceState)
+	{
+		SetStateMachine(resourceState);
+	}
+
+	actualState = 0;
+	nextState = 0;
+}
+
+Transition* ComponentAnimation::CheckTransitions(State* state)
```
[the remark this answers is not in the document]
AuxPort66: I don't want a copy and I prefere a Pointer instead of a std::optional<std::reference_wrapper<Transition>>

ericdacal:
I find some bugs using the state machines:
-  Engine crash when you delete a state in Edit State Machine.
- If you change the animation of an state in Edit, and after that you run, Engine crash.

Dedado3000 (approved):
Game Design is using a subranch based in this and we aren't having any problem, for future problems we will report directly to the grafics team

AuxPort66:
> 
 Report to me. I'm working on updates on animations with part of Art department

PR #321, opened 2023-04-25
feat: basic drone behaviour (Patrol - Seek)

markmeliaa:
(and a bunch of physics fixes, thanks physics team).

How does this work? Well, you have to add the following scripts into the drone to make its behaviour work:
- PatrolBehaviourScript: given two waypoints, the enemy will make a route between those two waypoints (but set this script to disabled from the inspector window, it will be managed from another script).
- SeekBehaviourScript: given a target, the enemy will start following that target (also, set this script to disabled from the inspector window, it will be managed from another script).
- EnemyDroneScript: this script will take nothing, but needs that PatrolBehaviourScript and SeekBehaviourScript are present in this same GameObject. Depending on if Bix is near enough the drone or not, this will manage the behaviour of the drone between PATROL and SEEK.
- Finally, EnemyCollidesFloorScript, just add this to each enemy so they don't ignore the ground and fall into the infinite :I

If you run CatinaScriptsVS2.axolotl you can check this working.


Game/Scripts/EnemyCollidesFloorScript.cpp
```diff
@@ -0,0 +1,71 @@
...
+	//	bool hasHit = Physics::Raycast(line, hit, owner);
+
+	//	if (hasHit && hit.hitPoint.y > maxHeight)
+	//	{
+	//		maxHeight = hit.hitPoint.y;
+	//	}
+
+	//}
+
+	//rigidBody->SetBottomHitPoint(maxHeight);
```
Dedado3000, line 55: I see some commented code, i don't know if this is intented to  the next milestone or deprecated

Dedado3000 (approved):
If you can ask my question pls, everything else seems ok

Game/Scripts/EnemyCollidesFloorScript.cpp
markmeliaa: I think this comes from @OriolSors' commit, so I guess it is intended.

totisama (approved):
Nice job!!!

DADSILVER (approved):
Awesome work, Just only one thing.

Game/Scripts/PlayerMobilityScript.cpp
```diff
@@ -88,7 +88,7 @@ void PlayerMovilityScript::Move()
 		sizeForce = deltaTime * dashForce;
 		if (nextDash == 0)
 		{
-			nextDash = SDL_GetTicks() + 200;
+			nextDash = (float)(SDL_GetTicks()) + 200.0f;
```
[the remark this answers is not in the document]
markmeliaa: I'll upload this fix in my current branch so we can merge this one.

PR #325, opened 2023-04-26
refactor: make modules private in application

gerard-queralt:
In general, it's considered good practice to not have public members and access them through getters.

**IMPORTANT**
In C++, maps don't preserve the order the elements where inserted in, and instead are sorted by the key. This means that the order in the constructor of Application is now meaningless, and what decides the order of insertion is the order of the enum used to index the dictionary. Not ideal, but not the end of the world, we will do the same but in a different file.


gerard-queralt:
Added a forward declaration file in the last commit to make sure compilation performance remains the same. After reordering modules before, we just needed to recompile the Application source file, but now, since it changed the header, it forced a recompilation of basically everything. By adding this forward declaration, now it again only recompiles the source file.

Source/Application.h
```diff
@@ -46,25 +35,14 @@ class Application
...
+	ScriptFactory* GetScriptFactory() const;
+	template<typename M>
+	M*GetModule();
+
+private:	
 	std::unique_ptr<ScriptFactory> scriptFactory;
 
-private:
-	std::vector<std::unique_ptr<Module> > modules;
+	std::map<ModuleType, std::unique_ptr<Module>> modules;
```
AuxPort66, line 45: I don't agree with turning the most looped vector of the entire engine into a map.

gerard-queralt: Working on it as we speak.

AuxPort66 (dismissed):

Game/Scripts/PlayerMobilityScript.cpp
```diff
@@ -89,7 +89,7 @@ void PlayerMobilityScript::Move()
 	size = speed * deltaTime * 1.1f;
 
 	// Dash pressing E during 0.2 sec
-	if (App->input->GetKey(SDL_SCANCODE_E) != KeyState::IDLE && GetCanDash())
+	if (App->GetModule<ModuleInput>()->GetKey(SDL_SCANCODE_E) != KeyState::IDLE && GetCanDash())
```
Dedado3000, line 92: We must save the module input first and not call get the module 16 times in the same script

Dedado3000 (approved):
I don't know if what i comment is true, but if we need to change that i saw a few more .cpp where it happens (i don't want to overflow this)

Game/Scripts/PlayerMobilityScript.cpp
gerard-queralt: You're right, but also, this call is just an access to a map, so it's not the end of the world in terms of cost. I'll change it now that you've spotted it, but I don't think we need to enforce this in future MRs.

gerard-queralt: Managed to go from 563 calls to `App->GetModule<X>` to 459, but I think I'll create a separate ticket for this. There are not only repeated calls to get the modules, but also to the modules themselves, like `GetLoadedScene`, `GetWindow` or `GetSelectedGameObject`. This would take a few hours...

AuxPort66 (approved):

markmeliaa (approved):

PR #329, opened 2023-04-27
fix: OnPlayMode functional again

DADSILVER:
- Fix problem with ctrl+q.
- remove a raycast not used from fucn Move(...).
- Add new raycastFirst with a exceptionGameObject.
- remove save options and other features that should not be usable during onGameMode.


OriolSors (changes requested):
Very good job @DADSILVER! I tested in Game Mode and PlayMode:

- The Play issue with FPS is solved
- The Rotation of the player now works on PlayMode
- Good overload of the RaycastFirst. Hope you can use it while we implement the Bullet...

Unfortunately, the Drone/Enemy is falling down on PlayMode but is working fine in Game Mode. Can you check it?

gerard-queralt (commented):
I see you've removed a couple calls to `SetReadyToEliminate`. Could you explain what that method does? Destroy the player game object? I'm not sure just by the name.

Source/Physics/Physics.h
```diff
@@ -25,6 +25,7 @@ class Physics
 	static bool Raycast(const LineSegment& ray, RaycastHit& hit);
 	static bool Raycast(const LineSegment& ray, RaycastHit& hit, GameObject* exceptionGameObject);
 	static bool RaycastFirst(const LineSegment& ray);
+	static bool RaycastFirst(const LineSegment& ray, GameObject* exceptionGameObject);
```
gerard-queralt, line 28: Maybe `exceptionGameObject` could be a collection (C-style array or a vector), so it's more flexible.

gerard-queralt (approved):
Haven't tested, but the code looks good.

Source/Physics/Physics.h
DADSILVER: I understand that this is looking to the future, but right now we wouldn't use it and we don't know if we will ever use it. That's why I think it's better not to implement it or to wait until it's necessary :)

DADSILVER:
> I see you've removed a couple calls to `SetReadyToEliminate`. Could you explain what that method does? Destroy the player game object? I'm not sure just by the name.

Thanks for reminding me. It is a function that was used in the rigged body but is no longer needed. I have removed it.

AlexisDieu:
- First fps looks great I am around 50fps everytime
- I can move and rotate without any problem
- if i wait the drone can pass through me 
image
ezgif com-video-to-gif
- I also see some raycast issue because only with walking i can pass through object 
image
ezgif com-video-to-gif (1)

Else everything seems great !

MarcGonzalezMoratona:
I have this error when pressing ctrl + x

image
image
image

MarcGonzalezMoratona:
I see in the description that you fixed Ctrl + q but I don't see anything related to this in the code. What is this shortcut supposed to do? Can you explain me the ctrl + q changes?

DADSILVER:
> I see in the description that you fixed Ctrl + q but I don't see anything related to this in the code. What is this shortcut supposed to do? Can you explain me the ctrl + q changes?

In moduleinput in the line 198 more or less. GitHub doesn't show theses lines because the change is inside the función.

gerard-queralt:
Hi @MarcGonzalezMoratona, there was an error with deleting GameObjects that's been mostly fixed. #330 should have fixed the crash, and in #332 fixes a non-breaking bug where the non-static objects weren't removed from the list in Scene. Could you try merging develop locally to see if the crash still occurs?

MarcGonzalezMoratona:
> Hi @MarcGonzalezMoratona, there was an error with deleting GameObjects that's been mostly fixed. #330 should have fixed the crash, and in #332 fixes a non-breaking bug where the non-static objects weren't removed from the list in Scene. Could you try merging develop locally to see if the crash still occurs?

I see that supr does not remove the objects from the Hierarchy, Ctrl + X works, but not Ctrl + V or Ctrl + D.

I also see that if you create an empty GameObject, then Ctrl + X and Ctrl + V for this GameObject, there is an error:
image
image
image

MarcGonzalezMoratona:
> Hi @MarcGonzalezMoratona, there was an error with deleting GameObjects that's been mostly fixed. #330 should have fixed the crash, and in #332 fixes a non-breaking bug where the non-static objects weren't removed from the list in Scene. Could you try merging develop locally to see if the crash still occurs?

Also tried to make Ctrl+X on the scene, the parent of all GameObjects and got this error:
image
image

gerard-queralt:
Thanks Marc, I'll take a look.

gerard-queralt:
Should be fixed in #334 @MarcGonzalezMoratona

gerard-queralt (approved):

OriolSors (approved):

PR #330, opened 2023-04-28
fix: corrected use of remove_if

gerard-queralt:
After calling `std::remove_if`, you must call `erase` using the iterator it returns. Someone didn't read the full documentation ^^'


Vlukkee (approved):

Dedado3000 (approved):
Well, that was fast

PR #333, opened 2023-04-30
refactor: context menu to add a component is alphabetically sorted automatically

gerard-queralt:
I found it quite confusing that the menu to add a component wasn't alphabetically sorted. Instead of just moving stuff around, which opened us to the same thing happening again, I defined a simple structure to represent the action to add a component, created a vector of them and sorted it. This way, when adding components in the future, we will just add another entry to this vector and not care about the order.


Dedado3000 (approved):

markmeliaa (dismissed):
Great job, I find this change quite useful! 
Can I ask what does std::bind exactly do? If I did not misunderstood it, when using bind you are linking a function (in this case the function that creates the selected component) to the `action` vector, am I right?

gerard-queralt:
> Can I ask what does std::bind exactly do? If I did not misunderstood it, when using bind you are linking a function (in this case the function that creates the selected component) to the `action` vector, am I right?

The short version is that it creates a "new" function which already has parameters bound to its call. For example, this call to `std::bind` that I make in the code:
```c++
std::bind(&WindowInspector::AddComponentLight, this, LightType::SPOT)
```
Would be equivalent to this lambda expression:
```c++
[this] ()
{
    this->AddComponentLight(LightType::SPOT);
}
```
Basically, it's a cleaner way to create instances of a `std::function`.

Going a bit more in-depth, you might wonder what's the point of calls like `std::bind(&WindowInspector::AddComponentPlayer, this)`, since `WindowInspector::AddComponentPlayer` does not have any parameters in its signature. But actually, it does! All class-instance functions in C++ have an implicit parameter: the `this` pointer. In other languages like Python, this parameter is explicitly passed to the function. From one of the examples of [the official documentation](https://docs.python.org/3/tutorial/classes.html):
``` python
class MyClass:
    """A simple example class"""
    i = 12345

    def f(self): # Notice the parameter "self"!
        return 'hello world'
```

That's why passing just a reference to the function gives a compiler error:
image
Notice that the first parameter in the function (the second argument of the call) is a `WindowInspector*`.

However, `std::bind` has a lot of other cool uses, albeit probably unneeded in most cases. If you're curious, I encourage you to read [the documentation](https://en.cppreference.com/w/cpp/utility/functional/bind) and see how else it can be used!

markmeliaa:
> However, `std::bind` has a lot of other cool uses, albeit probably unneeded in most cases. If you're curious, I encourage you to read [the documentation](https://en.cppreference.com/w/cpp/utility/functional/bind) and see how else it can be used!

Thank you for the explanation, I'll take a look to the documentation!

markmeliaa (approved):

Dedado3000 (approved):

PR #332, opened 2023-04-30
fix: when deleting a GameObject, remove it from non-static

gerard-queralt:
(cherry picked from commit b464085df780a57af79a25c4de6dff32ce88eaec)

PR #334, opened 2023-05-01
fix: errors with ctrl+x and ctrl+v

gerard-queralt:
Reported by @MarcGonzalezMoratona in #329


DADSILVER (dismissed):

MarcGonzalezMoratona:
@Querfer#0370 what I see when testing is that when you do ctrl+v several times the gameobjects are created inside each other, and maybe it would be more practical to do it at the same level

MarcGonzalezMoratona:
If you create an empty gameObject, Ctrl+C and Ctrl+V, you will see your gameobject now has a child. Try to Ctrl+X a gameobject that has a child and you will see this:
image

MarcGonzalezMoratona:
This happened trying to copy an audio source:
image

MarcGonzalezMoratona:
Does it make sense that after Ctrl+X you can make Ctrl+V of the same gameObject several times?

I think that normal behaviour is to just let the user paste once after Ctrl+X and many times after Ctrl+C

gerard-queralt:
Thank for your feedback Marc. Merging #332 into this, I've fixed the error about deleting a GameObject that has been copied. However, I haven't been able to fix the crash when cutting it. Not only that, I've ran into some other crashes that I haven't quite figured out how to repro. I'll take a better look tomorrow.

AuxPort66 (approved):

Source/DataModels/Windows/EditorWindows/WindowHierarchy.h
```diff
@@ -14,4 +14,21 @@ class WindowHierarchy : public EditorWindow
...
 	void DrawRecursiveHierarchy(GameObject* gameObject);
+
+    void Create2DObjectMenu(GameObject* gameObject);
+
+    void MoveObjectMenu(GameObject* gameObject, std::vector<GameObject*>& children);
+
+    void Create3DObjectMenu(GameObject* gameObject);
+
+	bool IsModifiable(const GameObject* gameObject) const;
+    void DeleteGameObject(const GameObject* gameObject) const;
```
MarcGonzalezMoratona, line 25: maybe we could solve indentation of this methods

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
```diff
@@ -25,6 +25,36 @@ void WindowHierarchy::DrawWindowContents()
...
+
+    ModuleInput* input = App->GetModule<ModuleInput>();
+
+    if (SDL_ShowCursor(SDL_QUERY)
+        && input->GetKey(SDL_SCANCODE_LCTRL) == KeyState::REPEAT
+        || input->GetKey(SDL_SCANCODE_LCTRL) == KeyState::DOWN)
+    {
+        if (input->GetKey(SDL_SCANCODE_C) == KeyState::DOWN)
+        {
+			CopyAnObject();
```
MarcGonzalezMoratona, line 37: indentation issues

MarcGonzalezMoratona:
I can't copy, delete or cut a canvas component
image

Source/DataModels/Windows/EditorWindows/WindowHierarchy.h
gerard-queralt: Actually, it's the other way around, that's the only method that's correctly indented. That one is using tabs, as we agreed in our programming rules, while the others are using spaces. Having said that, I wouldn't bother since all of this will be fixed automatically once I'm allowed to merge #258

MarcGonzalezMoratona:
When you copy-paste an object, right now it appears inside the original one.
It will be better if it appears at the same level.

gerard-queralt:
Thank you Marc, fixed the issue with the canvas. About the parent of the copied object, I'm not sure what's best. I'll take a look at how Unity does it for reference and we can discuss what we prefer for our engine.

MarcGonzalezMoratona:
> Thank you Marc, fixed the issue with the canvas. About the parent of the copied object, I'm not sure what's best. I'll take a look at how Unity does it for reference and we can discuss what we prefer for our engine.

Ctrl + C looks great, Ctrl + X and delete (supr) are not working for me
image

gerard-queralt:
Hopefully everything works now!

MarcGonzalezMoratona (approved):

AuxPort66 (approved):

PR #335, opened 2023-05-01
feat: Uniform scale

MarcGonzalezMoratona:
Component transform has now a checkbox to scale objects proportionally in the x, y and z axis simultaneouslly


Source/DataModels/Components/ComponentTransform.h
```diff
@@ -41,11 +41,13 @@ class ComponentTransform : public Component, public Drawable
 	const AABB& GetEncapsuledAABB();
 	const OBB& GetObjectOBB();
 	bool IsDrawBoundingBoxes() const;
+	bool IsUniformScale() const;
 
 	void SetPosition(const float3& position);
 	void SetRotation(const float3& rotation);
 	void SetRotation(const float4x4& rotation);
 	void SetScale(const float3& scale);
+	void SetUniformScale(const float3& scale, const char modifiedScaleAxis);
```
gerard-queralt, line 50:
```suggestion
	void SetUniformScale(const float3& scale, char modifiedScaleAxis);
```
Not much of a point in having a `const` primitive type.

Source/DataModels/Components/ComponentTransform.h
```diff
@@ -188,9 +196,42 @@ inline void ComponentTransform::SetScale(const float3& scale)
 {
 	sca = scale;
 
-	if (sca.x <= 0) sca.x = 0.0001f;
-	if (sca.y <= 0) sca.y = 0.0001f;
-	if (sca.z <= 0) sca.z = 0.0001f;
+	if (sca.x <= 0) {
+		sca.x = 0.0001f;
```
gerard-queralt, line 200:
```suggestion
std::max(sca.x, 0.0001f);
```
Same for other instances of this pattern. Though this is just personal preference, so feel free to ignore me :P

Source/DataModels/Components/ComponentTransform.h
```diff
@@ -188,9 +196,42 @@ inline void ComponentTransform::SetScale(const float3& scale)
...
+	}
+	if (sca.y <= 0) {
+		sca.y = 0.0001f;
+	}
+	if (sca.z <= 0) {
+		sca.z = 0.0001f;
+	}
+}
+
+inline void ComponentTransform::SetUniformScale(const float3& scale, const char modifiedScaleAxis)
```
gerard-queralt, line 210: Maybe overkill, but we could add an ENGINE_LOG in case the axis is not recognized.

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentTransform.cpp
```diff
@@ -10,7 +10,7 @@
 #include "ModuleScene.h"
 
 WindowComponentTransform::WindowComponentTransform(ComponentTransform* component) :
```
gerard-queralt, line 12: Could be wrong, but I think this copy constructor is not needed and can be deleted. Not related to this PR though.

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentTransform.h
```diff
@@ -27,7 +27,9 @@ class WindowComponentTransform : public ComponentWindow
 	bool translationModified;
 	bool rotationModified;
 	bool scaleModified;
+	char modifiedScaleAxis;
```
gerard-queralt, line 30: This is just a suggestion and it's most definetly not needed for this feature, but in general, I prefer using enums in these instances, for a double purpose: one, you make sure a wrong value is never introduces, and two, whoever is reading your code can easily identify which are the accepted values.
```suggestion
	Axis modifiedScaleAxis;
```
Again, not needed and just personal preference, feel free to ignore this.

gerard-queralt (commented):
Thank you Marc! Left a few minor suggestions regarding the code, but nothing required. I'll test the branch tomorrow and let you know if I find anything strange.

Source/DataModels/Components/ComponentTransform.h
MarcGonzalezMoratona: I think it's not necessary if we have an enum for the axis

gerard-queralt: Agreed! :)

Source/DataModels/Components/ComponentTransform.h
```diff
@@ -6,6 +6,7 @@
 
 #include "Geometry/AABB.h"
 #include "Geometry/OBB.h"
+#include <DataModels/Windows/SubWindows/ComponentWindows/WindowComponentTransform.h>
```
gerard-queralt, line 9:
```suggestion
#include "DataModels/Windows/SubWindows/ComponentWindows/WindowComponentTransform.h"
```
https://stackoverflow.com/questions/21593/what-is-the-difference-between-include-filename-and-include-filename

Source/DataModels/Components/ComponentTransform.h
```diff
@@ -186,11 +195,30 @@ inline void ComponentTransform::SetRotation(const float4x4& rotation)
...
-
-	if (sca.x <= 0) sca.x = 0.0001f;
-	if (sca.y <= 0) sca.y = 0.0001f;
-	if (sca.z <= 0) sca.z = 0.0001f;
+	sca.x = std::max(scale.x, 0.0001f);
+	sca.y = std::max(scale.y, 0.0001f);
+	sca.z = std::max(scale.z, 0.0001f);
+}
+
+inline void ComponentTransform::SetUniformScale(const float3& scale, Axis modifiedScaleAxis)
```
gerard-queralt, line 203: This is picky, but I'd prefer if the three axis were always in the same position insde the `if` blocks
```c++
sca.x = ...
sca.y = ...
sca.z = ...
```
What do you think?

MarcGonzalezMoratona: We need to preserve this order because the variable sca.x, sca.y and sca.z is used to compute the other values. We could save the value in an aux variable, but it looks easier for me this way

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentTransform.cpp
MarcGonzalezMoratona: I'm using it to assign the initial value to the members. Is it possible to remove the constructor and assign the values in another place? How can we do this?

gerard-queralt: C++ has implicit copy constructors that just assign the values from the original instance to the new one. For this reason, there's no point in having constructors that basically do
```c++
this->a = original.a;
this->b = original.b;
...
```
We do need to override the implicit copy constructor when the copy process is more complex, like in the case of `GameObject`, but I'm pretty sure you can simply delete this constructor and everything will work the same.

Documentation: https://en.cppreference.com/w/cpp/language/copy_constructor
Specifically, there's a section called `Implicitly-declared copy constructor`.

gerard-queralt (approved):
Tested and everything seems to work, thank you Marc!

markmeliaa (approved):
Everything works as intended, good job Marc!

Piterga12 (approved):
Everything works fine!

PR #337, opened 2023-05-03
perf[frame_cap]: fixed frame cap not working accurately

MacXxs:
* Fixed the frame time not being accurate relatively to the frame cap
* Massive boost in performance

### DIFF
--- a/Source/Application.cpp
+++ b/Source/Application.cpp
@@ -81,12 +81,8 @@ bool Application::Start()
 
 update_status Application::Update()
 {
-	float ms;
-#ifdef ENGINE
-	(isOnPlayMode) ? ms = onPlayTimer.Read() : ms = appTimer.Read();
-#else
-	ms = appTimer.Read();
-#endif // ENGINE
+	bool playMode = isOnPlayMode;
+	float ms = playMode ? onPlayTimer.Read() : appTimer.Read();
 
 	for (const std::unique_ptr<Module>& module : modules)
 	{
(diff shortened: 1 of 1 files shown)


AlexisDieu:
tested and approved

AlexisDieu (dismissed):
Tested and approved

Source/Application.cpp
```diff
@@ -117,13 +117,12 @@ update_status Application::Update()
 
 	float dt;
 #ifdef ENGINE
-	(isOnPlayMode) ? dt = (onPlayTimer.Read() - ms) / 1000.0f : dt = (appTimer.Read() - ms) / 1000.0f;
+	(isOnPlayMode) ? dt = (onPlayTimer.Read() - ms) : dt = (appTimer.Read() - ms);
```
gerard-queralt, line 120:
```suggestion
	dt = isOnPlayMode ? onPlayTimer.Read() - ms : appTimer.Read() - ms;
```

Source/Application.cpp
```diff
@@ -81,12 +81,8 @@ bool Application::Start()
 
 update_status Application::Update()
 {
-	float ms;
-#ifdef ENGINE
-	(isOnPlayMode) ? ms = onPlayTimer.Read() : ms = appTimer.Read();
-#else
-	ms = appTimer.Read();
-#endif // ENGINE
+	bool playMode = isOnPlayMode;
```
gerard-queralt, line 84: Even if it's redundant, maybe we could change this to
```suggestion
#ifdef ENGINE
	bool playMode = isOnPlayMode;
#else
	bool playMode = false;
#endif
```
Just so it's obvious to the reader what will happen in Game. But as I said it's redundant, since `isOnPlayMode` will always be `false` in Game.

MacXxs: Actually it should be "playMode = true" when it's in game mode. But even then, i woudln't add this change, not a fan of having redundant code xD and also the boolean is self explenatory i think it's pretty clear as it is.

gerard-queralt (approved):

AlexisDieu (approved):

PR #338, opened 2023-05-03
feat: Import scenes into scenes

Dedado3000:
New Button Added that allows yo to import a scene into another scene, giving facilities to merge work in diferent parts of level design.

Knowed Bug:
You can't import an scene with an script that have a pointer to a gameobject, Im currently working on it but doesn't affect the motive for what we use this now.


Source/DataModels/Windows/WindowMainMenu.h
```diff
@@ -3,6 +3,7 @@
 #include "EditorWindows/WindowAbout.h"
 #include "EditorWindows/ImporterWindows/WindowLoadScene.h"
 #include "EditorWindows/ImporterWindows/WindowSaveScene.h"
+#include "WindowImportScene.h"
```
gerard-queralt, line 6: Use a forward declaration. Now that I notice, we should use them for the other windows as well, if possible.

Source/Modules/ModuleScene.cpp
```diff
@@ -316,6 +316,36 @@ void ModuleScene::LoadSceneFromJson(const std::string& filePath)
 #endif // !ENGINE
 }
 
+void ModuleScene::ImportFromJson(const std::string& filePath)
```
gerard-queralt, line 319: Could this be done in the (currently unused) `SceneImporter`? This way we could do asyncronous importing just reusing the code in `ModuleResources`. It would require actually implementing the `SceneResource` class as well, but that's also a bit overdue in my opinion, especially once we start having many scenes.

Source/Modules/ModuleScene.cpp
```diff
@@ -387,6 +417,66 @@ void ModuleScene::SetSceneFromJson(Json& json)
...
+		if (obj->GetComponent(ComponentType::TRANSFORM) != nullptr)
+		{
+			//Quadtree treatment
+			AddGameObject(obj);
+
+		}
+
+	}
+	App->GetModule<ModuleEditor>()->RefreshInspector();
+	loadedScene->AddSceneCameras(loadedCameras);
```
gerard-queralt, line 469: Maybe create a new method, I don't know, `MergeScenes`, that encapsulates this logic? Following the above comment, maybe it can also be in `SceneImporter`.

gerard-queralt (commented):
Added a few questions, would like to know yours and @AuxPort66's opinion.

Source/Modules/ModuleScene.cpp
Dedado3000: I don't know if using asyncronous importing there have problems because what will you do while is loading the scene to include in the scene? you can't work until it finish

gerard-queralt: "Asyncronous" might have been a strong word. Currently what's done when a resource is imported is display a "Loading" window, so the user has some visual feedback. That's what I meant :P

Source/Modules/ModuleScene.cpp
```diff
@@ -477,6 +567,93 @@ std::vector<GameObject*> ModuleScene::CreateHierarchyFromJson(Json& jsonGameObje
...
 }
 
+std::vector<GameObject*> ModuleScene::InsertHierarchyFromJson(Json& jsonGameObjects)
+{
+	std::vector<GameObject*> gameObjects{};
+	std::unordered_map<UID, GameObject*> gameObjectMap{};
+	std::unordered_map<UID, UID> childParentMap{};
+	std::unordered_map<UID, std::pair<bool, bool>> enabledAndActive{};
+	//Map created in order to create new ID to the object (now you can import 2 times the same scene)
+	std::unordered_map<UID, UID> uidMap{};
```
AuxPort66, line 577: Like I say before, changing the UIDs of a GameObject inside the scene it's problematic, not only for the dependencies of their childrens, that have the UID, also for scripting, because we can save a reference to a gameObject, and that reference is saved with the UID.

So we are destroying the scripting dependencies of all scenes, even the ones that don't need to change the UID because they don't have the same root.

In any case, if you want to load 2 identical scenes, it's an acceptable risk that your scripting will be pointing to the original gameObject, you can check that they are the same scene and do this by looking at the UID of the root, but you can have this functionality with a ctrl+c ctrl +v, you don't need a full load function for this.

AuxPort66 (changes requested):

Source/DataModels/Windows/WindowMainMenu.h
Dedado3000: I don't know what but i can't use a forward declaration, can we check together? im feeling so dumb

gerard-queralt: It's possible it can't be done in this case ^^'

gerard-queralt: Nope, you definitely can, make sure to include the actual class in the source file.

Source/WindowImportScene.h
gerard-queralt, line 1: Put these new files in the proper folder.

gerard-queralt (dismissed):

Source/WindowImportScene.h
Dedado3000: Check it guapetón

Source/Modules/ModuleScene.cpp
Dedado3000: Guapetona, is okey now?

Source/DataModels/Windows/WindowMainMenu.h
Dedado3000: Okey they where other includes from imguy needed that where included in  a include from an include, and obviously i was missing it

AuxPort66 (dismissed):

gerard-queralt (approved):

AuxPort66 (approved):

PR #340, opened 2023-05-03
fix: can move object up and down, corrected use of iterators

gerard-queralt:
Branched from #334, so that must be merged first.

### DIFF
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -32,6 +32,12 @@
 
 #include "Scene/Scene.h"
 
+enum class GameObject::DirectionToMove
+{
+	UP,
+	DOWN
+};
+
 // Root constructor
 GameObject::GameObject(const std::string& name, UID uid) :
 	GameObject(name, nullptr, uid, true, true, StateOfSelection::NO_SELECTED, false)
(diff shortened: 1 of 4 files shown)


Source/DataModels/GameObject/GameObject.h
```diff
@@ -103,6 +103,9 @@ class GameObject
 
 	bool IsAChild(const GameObject* child);
 
+	enum class DirectionToMove;
```
MarcGonzalezMoratona, line 106: why do we need this to declare an enum class? Is it mandatory? Last time I did an enum class I think I didn't do that

MarcGonzalezMoratona:
Thank you for your work Gerard! I left some questions, but the code looks great. I will check now that everything is fixed

MarcGonzalezMoratona (approved):

Source/DataModels/GameObject/GameObject.h
gerard-queralt: We need to, because we use it as a parameter of the method. Could you share where you think you did that? It's possible I could be missing something.
If you are asking about the forward declaration, it's just to cut compilation time. I personally don't like it and it's probably really, really unimpactful, but well. Actually, the more I write about it, the more I dislike it. I'd like to hear your thoughts on that.

Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -32,6 +32,12 @@
 
 #include "Scene/Scene.h"
 
+enum class GameObject::DirectionToMove
```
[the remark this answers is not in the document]
gerard-queralt: I'm not sure, I feel like `Direction` is too broad. How do you feel about something like `HierarchyDirection`, or `HierarchyMovement`?

Source/DataModels/GameObject/GameObject.h
MarcGonzalezMoratona: In the uniform scaling PR we did an enum class for the axis and I didn't forward declared I think. Maybe it's a different use case, but I would like to understand the differences

Source/DataModels/GameObject/GameObject.cpp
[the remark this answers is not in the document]
MarcGonzalezMoratona: HierarchyDirection it's good for me, but well, you don't have to change it, it was just a suggestion

Source/DataModels/GameObject/GameObject.h
gerard-queralt: The idea of forward declarations is to avoid having to recompile a class because another one changed. In this case, this forward declaration would be useful if we expected the enum class to change with relative frequency, but I don't think we will ever allow objects to be moved diagonally in the hierarchy :P

Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -635,6 +641,46 @@ bool GameObject::IsAChild(const GameObject* child)
...
+
+	auto childIterator = std::find_if(childrenVectorBegin, childrenVectorEnd,
+		[child](const std::unique_ptr<GameObject>& otherChild)
+		{
+			return otherChild.get() == child;
+		});
+	if (childIterator == childrenVectorEnd)
+	{
+		ENGINE_LOG("Object being moved (%s) is not a child of this (%s)",
+			std::to_string(child->GetUID()).c_str(), std::to_string(this->GetUID()).c_str());
```
markmeliaa, line 657: Maybe instead of the UID we could display by Console the actual name of the GameObject? (Same for the other situations in which something is printed through the console).

markmeliaa (dismissed):
Everything looks and works correctly, left a small suggestion. Any idea why this stopped working? It is weird that it previously worked but now it only works when an GameObject has an actual child.

gerard-queralt:
> Any idea why this stopped working? It is weird that it previously worked but now it only works when an GameObject has an actual child.

Looking at the history, I think it was some of the changes David did optimizing the window. Previously we checked the children of the parent, but he changed it to check your own children. Easy to mix up with so many game objects to consider ^^'

Source/DataModels/GameObject/GameObject.cpp
gerard-queralt: Yes, makes more sense. I'm too used to debug by looking at the UID :P

markmeliaa (approved):

MarcGonzalezMoratona (approved):
Good job!

PR #341, opened 2023-05-03
feat: Physics bullet integration

xavi1111:


Source/ModulePhysics.cpp
```diff
@@ -0,0 +1,43 @@
...
+}
+
+ModulePhysics::~ModulePhysics()
+{
+}
+
+bool ModulePhysics::Init()
+{
+	//collision configuration contains default setup for memory, collision setup. Advanced users can create their own configuration.
+	btDefaultCollisionConfiguration* collisionConfiguration = new btDefaultCollisionConfiguration();
```
gerard-queralt, line 15: All the uses of `new` inside this method don't have the accompaning `delete`s, which means the allocated memory is never freed. If we don't need to save this objects as members, can't we just use references?
```suggestion
	btDefaultCollisionConfiguration collisionConfiguration;
```
And if we need to save them, consider using unique pointers.

Source/ModulePhysics.cpp
```diff
@@ -0,0 +1,43 @@
...
+	btSequentialImpulseConstraintSolver* solver = new btSequentialImpulseConstraintSolver;
+
+	btDiscreteDynamicsWorld* dynamicsWorld = new btDiscreteDynamicsWorld(dispatcher, overlappingPairCache, solver, collisionConfiguration);
+
+	dynamicsWorld->setGravity(btVector3(0, -10, 0));
+
+    return true;
+}
+
+update_status ModulePhysics::Update()
```
gerard-queralt, line 33: Can be deleted.

Source/ModulePhysics.cpp
```diff
@@ -0,0 +1,43 @@
...
+{
+    
+    return update_status::UPDATE_CONTINUE;
+}
+
+
+bool ModulePhysics::CleanUp()
```
gerard-queralt, line 40: Can be deleted.

Source/ModulePhysics.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "Module.h"
+
+class ModulePhysics
+{
+public:
+	ModulePhysics();
+	~ModulePhysics();
```
gerard-queralt, line 9: Missing `override`.
```suggestion
	~ModulePhysics() override;
```

Source/ModulePhysics.h
```diff
@@ -0,0 +1,18 @@
...
+
+	bool Init();
```
gerard-queralt, line 11: Missing `override`.
```suggestion
	bool Init() override;
```

gerard-queralt (changes requested):

Source/ModulePhysics.cpp
xavi1111: Yeah I agree, but this code is meerly to check the integration of the library it is not intented to stay as is.  It is the HelloWorld provided by the library and the only reason it is in this PR is to check that it compiles in more than one machine (mine).

gerard-queralt: Ah whoops, didn't realize this wasn't pointed to `develop`. Carry on then! :P

OriolSors (approved):

Dedado3000 (approved):

PR #342, opened 2023-05-04
fix: solve deleter

Dedado3000:
I worked really hard in solve the deleter, pls accept as soon as posible


AuxPort66 (approved):

gerard-queralt (approved):
We need more people like you, thank you for your tremendous work!

PR #343, opened 2023-05-04
feat: script headers trigger hot reloading, simpler REGISTER_FIELD macro

gerard-queralt:


Game/Scripts/DefaultScript.cpp
```diff
@@ -23,42 +23,13 @@ void DefaultScript::Update(float deltaTime)
...
-{
-	this->sentence = sentence;
-}
-
 GameObject* DefaultScript::GetCharacter() const
 {
 	return character;
 }
 
 void DefaultScript::SetCharacter(GameObject* character)
```
markmeliaa, line 31: Couldn't this get and set now be inline in the header file? Should we enforce inlining simple functions in scripting as we do in other classes?

markmeliaa (approved):
GJ!

DADSILVER (approved):

Game/Scripts/DefaultScript.cpp
gerard-queralt: Yes it could! About enforcing it, well, that's up to the gameplay team in my opinion. It should be better for performance, but it's not like this methods will be called every frame. In general, they will be called at the start for de-serialization, and at the end for serialization, both in Game and Engine. Specifically in Engine, only the methods of the scripts of the selected game object will be called each frame. All of this to say, the impact in performance is probably negligible, so choose whatever you feel most comfortable with ;)

PR #346, opened 2023-05-04
Fix: Rigid body serialization

AuxPort66:
I saw that the RigidBody component didn't save the gravity value, so the changes were lost.

I have checked the rest of components too and this was the only thing.

### DIFF
--- a/Source/DataModels/Components/ComponentRigidBody.cpp
+++ b/Source/DataModels/Components/ComponentRigidBody.cpp
@@ -253,6 +253,7 @@ void ComponentRigidBody::SaveOptions(Json& meta)
 	meta["useRotationController"] = (bool)GetUseRotationController();
 	meta["KpForce"] = (float)GetKpForce();
 	meta["KpTorque"] = (float)GetKpTorque();
+	meta["gravity.Y"] = (float)GetGravity().y;
 }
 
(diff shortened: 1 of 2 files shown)


Source/DataModels/Components/ComponentRigidBody.cpp
```diff
@@ -253,6 +253,7 @@ void ComponentRigidBody::SaveOptions(Json& meta)
 	meta["useRotationController"] = (bool)GetUseRotationController();
 	meta["KpForce"] = (float)GetKpForce();
 	meta["KpTorque"] = (float)GetKpTorque();
+	meta["gravity.Y"] = (float)GetGravity().y;
```
gerard-queralt, line 256: Probably unnecessary, but maybe we should save all three axis? I guess we won't use the other two, but worth asking. And if we only use one axis, maybe we could store the gravity as a float? But that's a separate topic.

gerard-queralt (approved):

Source/DataModels/Components/ComponentRigidBody.cpp
OriolSors: For now we are not going to use the X and Z axis for gravity, but it should be clear to store all the values. In any case, we will adapt this when adapting the bullet to rigidbodies

OriolSors (approved):

Issue #347, opened 2023-05-05
Problems loading a material in GameMode


PaablooCH:
The bug occurs when You put a gameobject in the scene and change the value of smoothness and metallic in "Default" type shader and save the scene (probably just applying the material would be the same). The next step is loading the scene in game mode. As the images shown, we get different results
![image](https://user-images.githubusercontent.com/62140600/236360638-dbf0e500-c0ac-491e-bc7d-f037bb8ba30b.png)*1
![image](https://user-images.githubusercontent.com/62140600/236360649-00477559-7a9b-4770-864b-4909a7048789.png)*2
Its seems that we load the default values of smoothness and metallic variables at some point (or we are not setting the ones that meta gives us) 
![image](https://user-images.githubusercontent.com/62140600/236360849-a5402a79-5edc-4bb2-ab51-400285b1dd75.png)*1
![image](https://user-images.githubusercontent.com/62140600/236360814-09033966-34b6-49b6-ae06-6793ce069ff1.png)*2
Note: the meta is stored well 
![image](https://user-images.githubusercontent.com/62140600/236361411-e8ba5ed8-41d5-4ca5-96d6-792aeeafca65.png)
Note2: i dunno if this happens with other variables

PR #348, opened 2023-05-05
feat: IBL implementation 

PaablooCH:
**What is done:**

- Cubemap class (with resource and importer): this class will create and store the differents maps textures for IrradianceMap and PrefilteredMap, also can transform and image to a cube image (preferable use HDR image, but any other type works too), it creates too a 2DTexture for EnvironmentBRDF. I create two cubemap files to test.
- We can use Includes in shaders: most of the fragments used to create the textures share the same functions, so in order to maintain clean the shaders I put all the common functions in a file inside a folder in shaders folder, i did the same with the common functions in pbr shaders. Note: i move the shader folder to Game folder.
- Texture Importer: i added the option to load an HDR image and i cleaned the code.
- Delete AmbientLight: with this implementation we don't longer need this type of light.
- Update scenes: i delete the ambient light and added a cubemap to all the scene that this branch have (if you have a scene in your branch, you will need to do the same).

Note: Like skybox there is no proper way to create a cubemap inside the engine.

**Things to test**:

- IBL with default and specular shaders, if you open the project in engine mode, you can export any model and check the results (with roughness and metallic 1 you could see the image). You can change between two cubemap in Scene.cpp line 525, the cubemaps are in Assets/Cubemaps.
image
image
- Save, load scene.

Note: If you load and scene in gameMode see this issue #347


sncarbo (dismissed):
Once I modify the outer angle value of a spotlight it stops working, I understand it is a bug:
error ibl

Source/DataModels/Components/ComponentMeshRenderer.cpp
```diff
@@ -223,6 +226,13 @@ void ComponentMeshRenderer::DrawMaterial(Program* program) const
 
 		float3 viewPos = App->GetModule<ModuleCamera>()->GetCamera()->GetPosition();
 		program->BindUniformFloat3("viewPos", viewPos);
+		glActiveTexture(GL_TEXTURE8);
+		glBindTexture(GL_TEXTURE_CUBE_MAP, App->GetModule<ModuleScene>()->GetLoadedScene()->GetCubemap()->GetIrradiance());
```
gerard-queralt, line 230: We can call `GetCubemap` only once and store the value.
```suggestion
		Cubemap* cubemap = App->GetModule<ModuleScene>()->GetLoadedScene()->GetCubemap();
		glBindTexture(GL_TEXTURE_CUBE_MAP, cubemap->GetIrradiance());
```

Source/DataModels/Cubemap/Cubemap.cpp
```diff
@@ -0,0 +1,290 @@
...
+#include "FileSystem/Json.h"
+
+#include "DataModels/Program/Program.h"
+#include "DataModels/Resources/ResourceCubemap.h"
+
+#include "Resources/ResourceTexture.h"
+
+#include "Geometry/Frustum.h"
+
+#define CUBEMAP_RESOLUTION 512
```
gerard-queralt, line 16: Should this values be modifiable by the user? If they are, maybe they belong in the resource's import options. @AuxPort66

Source/DataModels/Program/Program.h
```diff
@@ -75,34 +83,69 @@ inline void Program::BindUniformFloat4x4(const std::string& name, const float* d
 	glUniformMatrix4fv(glGetUniformLocation(id, name.c_str()), 1, transpose, data);
```
gerard-queralt, line 83: Maybe this method can reuse the new one you added to avoid repeated code. I guess something like
```suggestion
	BindUniformFloat4x4(glGetUniformLocation(id, name.c_str()), data, transpose);
```
Same for the rest.

Source/DataModels/Resources/ResourceCubemap.h
```diff
@@ -0,0 +1,53 @@
...
+
+	ResourceType GetType() const override;
+
+	void SaveImporterOptions(Json& meta) override {};
+	void LoadImporterOptions(Json& meta) override {};
+
+	void SaveLoadOptions(Json& meta) override {};
+	void LoadLoadOptions(Json& meta) override {};
+
+	const std::shared_ptr<ResourceTexture>& GetHDRTexture() const;
```
gerard-queralt, line 25: I don't know if this is done somewhere else, but we should not return smart pointers by reference. By doing so, we don't increase the reference count, which defeats the purpose of a shared pointer.
```suggestion
	std::shared_ptr<ResourceTexture> GetHDRTexture() const;
```
[This StackOverflow post](https://stackoverflow.com/questions/10643563/how-to-return-smart-pointers-shared-ptr-by-reference-or-by-value) addresses the concerns regarding the cost.

Source/DataModels/Scene/Scene.h
```diff
@@ -200,6 +190,16 @@ inline Skybox* Scene::GetSkybox() const
 	return skybox.get();
 }
 
+inline Cubemap* Scene::GetCubemap() const
+{
+	return cubemap.get();
+}
+
+inline void Scene::SetCubemap(std::unique_ptr<Cubemap> cubemap)
```
gerard-queralt, line 198: If don't inline this method and move it to the source file, we can use a forward declaration and not include `Cubemap.h`. Just pointing it out, I'm not sure which option is best.

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
```diff
@@ -6,7 +6,8 @@
 #include "ModulePlayer.h"
 #include "Scene/Scene.h"
 #include "ModuleInput.h"
-#include "GameObject/GameObject.h"
+
+#include "DataStructures/Quadtree.h"
```
gerard-queralt, line 10: Looks like this include is not used.

Source/FileSystem/Importers/TextureImporter.cpp
```diff
@@ -34,90 +34,60 @@ void TextureImporter::Import(const char* filePath, std::shared_ptr<ResourceTextu
...
+			if (FAILED(result)) // WIC failed try with HDR
 			{
-				result = DirectX::FlipRotate(img.GetImages(), img.GetImageCount(), img.GetMetadata(),
-					DirectX::TEX_FR_FLAGS::TEX_FR_FLIP_HORIZONTAL, flippedImg);
-
-				if (!FAILED(result)) imgResult = &flippedImg;
+				result = DirectX::LoadFromHDRFile(path, &md, img);
+				if (FAILED(result))
+				{
+					ENGINE_LOG("Can not load the texture.");
```
gerard-queralt, line 49:
```suggestion
					ENGINE_LOG("Cannot load the texture.");
```

Source/Modules/ModuleEditor.cpp
```diff
@@ -7,6 +7,7 @@
 #include "ModuleInput.h"
 
 #include "DataModels/Scene/Scene.h"
+#include "DataStructures/Quadtree.h"
```
gerard-queralt, line 10: Seems unused.

gerard-queralt (dismissed):
Thank you Pablo, left a few comments.

PaablooCH:
> Once I modify the outer angle value of a spotlight it stops working, I understand it is a bug: error ibl


I can't replicate it, pls give me more information. The cube was in specular or default shader? what values did you give to roughness and metallic?


https://user-images.githubusercontent.com/62140600/236594676-a5ebb4c3-c6fa-47c9-a447-a4784c8e2428.mp4

Source/DataModels/Cubemap/Cubemap.cpp
PaablooCH: No, they are constant values and because they are repeated and need to be passed through functions I defined in that way. I considered to make them a function variable, but like this I think is more readable.

Source/DataModels/Program/Program.h
PaablooCH: i don't see the point to do it, if bindUniformFloat4x4 would be more extens i would do it, but calling a function just to do a line i don't like it.
PS: just because the compiler can ignore the inline (as can make function not inlined inline) in terms of what he thinks, the thing could change if we can ensure that inline works for gl functions (that now i'm not sure)

Source/DataModels/Resources/ResourceCubemap.h
PaablooCH: i took from reference the resource Skybox that has a vector, i have adapted from there and forgot to delete const and & 👍

Source/DataModels/Scene/Scene.h
PaablooCH: i will search and respond when find something

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
PaablooCH: it is.
image

PaablooCH:
> Thank you Pablo, left a few comments.

Thanks i left you comments, did you test something? and if yes did you find something strange?

Source/DataModels/Program/Program.h
gerard-queralt: Inline works for functions that call other functions. It's not inlining the entire second function iside the first one, just the redirection to its memory address. So shouldn't be an issue in that regard. Anyway, it's fine if you want to keep it like it is, it was just a suggestion, not a big deal.

gerard-queralt:
> Thanks i left you comments, did you test something? and if yes did you find something strange?

I didn't, I'll take a look later today. Thank you for addressing my comments.

sncarbo:
> > Once I modify the outer angle value of a spotlight it stops working, I understand it is a bug: error ibl
> 
> Thanks for the review Santi, but unfortunately I couldn't replicate it, pls give me more information. The cube was in specular or default shader? what values did you give to roughness and metallic?
> 
>  spotlight.mp4

I'm looking for how I got the error because it seems to come up in very specific cases.

sncarbo:
@PaablooCH There is another bug when you create a spotlight and try to change the "Type" from "Spot" to "Point" in the "Spot Light" panel the program crashes. This is the call stack:
error ibl

sncarbo:
> > Once I modify the outer angle value of a spotlight it stops working, I understand it is a bug: error ibl
> 
> Thanks for the review Santi, but unfortunately I couldn't replicate it, pls give me more information. The cube was in specular or default shader? what values did you give to roughness and metallic?
> 
>  spotlight.mp4

Ok, I just saw that the error is not about changing the outer angle, I don't know why I thought it was associated to that. The error occurs when you have for example two spotlights and you delete one of the two, the other one stops working:

https://user-images.githubusercontent.com/116657602/236622160-36582208-edd9-48e5-b8ba-b1c533cb533d.mp4

PaablooCH:
> > > Once I modify the outer angle value of a spotlight it stops working, I understand it is a bug: error ibl
> > 
> > 
> > Thanks for the review Santi, but unfortunately I couldn't replicate it, pls give me more information. The cube was in specular or default shader? what values did you give to roughness and metallic?
> > spotlight.mp4
> 
> Ok, I just saw that the error is not about changing the outer angle, I don't know why I thought it was associated to that. The error occurs when you have for example two spotlights and you delete one of the two, the other one stops working:
> 
>  error.ibl.mp4

fixed, the problem was that were defined 2 variables with the same name and compared between them, doing that the comparation was always true

PaablooCH:
> @PaablooCH There is another bug when you create a spotlight and try to change the "Type" from "Spot" to "Point" in the "Spot Light" panel the program crashes. This is the call stack: error ibl

this bug is also in develop, i will solve it but is out of this pr scope

sncarbo:
@PaablooCH Should this happen?
error ibl

PaablooCH:
> @PaablooCH Should this happen? error ibl

it happens in develop also, so yes, this could be different when we implement shadows
image

sncarbo:
> > @PaablooCH Should this happen? error ibl
> 
> it happens in develop also, so yes, this could be different when we implement shadows image

Perfect, once the conflicts are solved I approve the pr.

sncarbo (approved):

AuxPort66 (commented):
I checked all the filesystem changes (Importers, ModuleResource, New Resource) and it looks ok.

Good job. Only one question, why you move the Shaders folder?

PaablooCH:
> I checked all the filesystem changes (Importers, ModuleResource, New Resource, etc) and it looks ok.
> 
> Good job. Only one question, why you move the Shaders folder?

thanks :)
as i told you i moved because it was more easier to implement the use of imports inside the Game folder than outside.

AuxPort66 (dismissed):

AuxPort66 (approved):

MacXxs (approved):
Great work Pablo as alwaaays!! 👍🏼

AuxPort66 (approved):

PR #352, opened 2023-05-06
feat [Game play]: separate mobility scripts into rotation and move

DADSILVER:
- The mobility script is divided into two scripts.
- Added a collision function for rotation (We seek to avoid colliding with the ground).
- updated the scene.


Game/Scripts/PlayerMoveScript.h
```diff
@@ -0,0 +1,58 @@
...
+    public Script
+{
+public:
+	PlayerMoveScript();
+	~PlayerMoveScript() override = default;
+
+	void Start() override;
+	void PreUpdate(float deltaTime) override;
+
+	float GetSpeed() const;
```
gerard-queralt, line 26: As a reminder, all these getters and setters are not needed if you're not using the `REGISTER_FIELD_WITH_ACCESSORS` macro.

markmeliaa (dismissed):
Everything works and looks correct, thank you David!

gerard-queralt (changes requested):

Game/Scripts/PlayerMobilityScript.cpp
```diff
@@ -39,408 +35,42 @@ PlayerMobilityScript::~PlayerMobilityScript()
...
-	ModuleInput* input = App->GetModule<ModuleInput>();
-	if (input->GetKey(SDL_SCANCODE_E) != KeyState::IDLE && GetCanDash())
-	{
-		sizeForce = deltaTime * dashForce;
-		if (nextDash == 0)
+		if (gameObjectScripts[i]->GetConstructName() == "PlayerMoveScript")
 		{
-			nextDash = static_cast<float>(SDL_GetTicks()) + 200.0f;
+			moveScript = gameObjectScripts[i];
+			moveScript->GetScript()->Start();
```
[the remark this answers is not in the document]
DADSILVER: The scrip is disable, so i need call start, preupdate...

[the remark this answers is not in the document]
gerard-queralt: Do they need to be disabled though? I feel like this might create issues down the line if someone forgets about disabling them, but could be wrong.

Game/Scripts/PlayerRotationScript.cpp
```diff
@@ -0,0 +1,53 @@
...
+	trans->UpdateTransformMatrices();
+
+
+	//Corroborate that you don't fuse with a owner
+	ComponentMeshCollider* collider = static_cast<ComponentMeshCollider*>(owner->GetComponent(ComponentType::MESHCOLLIDER));
+	math::vec points[8];
+	trans->GetObjectOBB().GetCornerPoints(points);
+	std::vector<float3> frontPoints = { points[1], points[3], points[5], points[7] };
+	float3 direction = (points[1] - points[0]).Normalized();
+	if (collider->IsRotationColliding(frontPoints, -direction, trans->GetLocalAABB().Size().z * 0.00f))
```
OriolSors, line 44: Are you passing a 0 value in the size? How are you doing the raycast then?

Source/DataModels/Components/ComponentMeshCollider.cpp
```diff
@@ -107,6 +107,31 @@ bool ComponentMeshCollider::IsColliding(std::vector<float3>& startingPoints, flo
 	return false;
 }
 
+bool ComponentMeshCollider::IsRotationColliding(std::vector<float3>& startingPoints, float3 direction, float size, float stepSize) const
+{
+	std::vector<float3> points;
+	startingPoints[0].y += 0.25f;
+	startingPoints[1].y -= 0.25f;
+	startingPoints[2].y += 0.25f;
+	startingPoints[3].y -= 0.25f;
```
OriolSors, line 116: What is the purpose of these offsets?

Game/Scripts/PlayerRotationScript.cpp
DADSILVER: I changed and nothing happened. So I forgot put it back again this vale.

Game/Scripts/PlayerMobilityScript.cpp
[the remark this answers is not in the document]
DADSILVER: For now, we believe this is the best way to do it. It may be changed at a later date.

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -0,0 +1,349 @@
...
+		{
+			componentAudio->PostEvent(audio::SFX_PLAYER_FOOTSTEPS_WALK_STOP);
+			playerState = PlayerActions::IDLE;
+		}
+	}
+
+	//rigidBody->AddForce(forceVector * forceParameter);
+
+	// Jump
+	if (input->GetKey(SDL_SCANCODE_SPACE) == KeyState::DOWN && jumps > 0)
```
UriMurcia, line 306: same for jump. I think would be better to separate, jump, crouch and dash in different scripts, but not the movement and rotation

Game/Scripts/PlayerMobilityScript.cpp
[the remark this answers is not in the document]
markmeliaa: If we keep the functionality of the gameplay in small scripts, we will need a general script for each entity that manages the scripts that the enitity will support (as for example the player with move/rotate or the drone with patrol/seek). That said, I agree that disabling every other script and only keeping active the manager one is not a good solution, but for now we are working like that until we have a better solution.

Game/Scripts/PlayerMobilityScript.cpp
```diff
@@ -39,408 +35,42 @@ PlayerMobilityScript::~PlayerMobilityScript()
...
+	if (!componentPlayer->IsStatic() && App->GetModule<ModuleCamera>()->GetSelectedPosition() == 0
+		&& !SDL_ShowCursor(SDL_QUERY))
 	{
-		sizeJump = deltaTime * jumpParameter;
-		jumps -= 1;
-		if (rigidBody->IsOnGround() || jumps > 0)
+		if (rotateScript)
 		{
-			rigidBody->AddForce(jumpVector * jumpParameter, ForceMode::Acceleration);
+			rotateScript->GetScript()->PreUpdate(deltaTime);
```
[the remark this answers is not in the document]
markmeliaa: Already adressed here https://github.com/Horizons-Games/Axolotl-Engine/pull/352#discussion_r1187658111

Game/Scripts/PlayerMoveScript.cpp
markmeliaa: This is related to what we talk in the discord once, I find separating behaviours in different scripts a good idea (one for attack, one for movement, one for rotation, one for health...), but maybe having one for jump, another one for crouch and another one for the dash is way too much (we are working on creating a Jump(), Crouch(), Dash() functions but for now David told me that we cannot decouple them so easily, so we'll do it in the future).

I don't find the usefulness of having scripts as divided as you say because no single entity in our game will dash, crouch or jump apart from the two main characters (who will also move, that's why all those actions are together). I mean, it is good to keep in mind the "one script should do one thing" principle, but only if that one thing is or would be reused by any other entity in our game (which is why the patrol and seek behaviours are separated into different scripts but the jump and crouch are not).

Game/Scripts/PlayerMobilityScript.cpp
```diff
@@ -39,408 +35,42 @@ PlayerMobilityScript::~PlayerMobilityScript()
...
-	{
-		if (playerState == PlayerActions::WALKING)
-		{
-			componentAudio->PostEvent(audio::SFX_PLAYER_FOOTSTEPS_WALK_STOP);
-			playerState = PlayerActions::IDLE;
-		}
-	}
-
-	//rigidBody->AddForce(forceVector * forceParameter);
+	componentPlayer = static_cast<ComponentPlayer*>(owner->GetComponent(ComponentType::PLAYER));
```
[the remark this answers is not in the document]
markmeliaa: There is a reason why ComponentPlayer is not yet dead, @DADSILVER explained it to me but I can't remember now, so he'll give you an answer.

Game/Scripts/PlayerRotationScript.cpp
```diff
@@ -0,0 +1,53 @@
...
+	{
+		float deltaTime = App->GetDeltaTime();
+		ComponentTransform* trans = static_cast<ComponentTransform*>(owner->GetComponent(ComponentType::TRANSFORM));
+		float3 newRot = trans->GetRotationXYZ();
+		newRot.y += input->GetMouseMotion().x * deltaTime;
+		trans->SetRotation(newRot);
+		trans->UpdateTransformMatrices();
```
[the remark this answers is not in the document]
markmeliaa: It could in fact, yep.

Game/Scripts/PlayerMoveScript.cpp
diegusmejia117: We havent discussed it but maybe the mini bosses or the boss migh be able to jump, dash or crouch, so i think is valuable to have it separated, but as you said Mark, its not needed right now. So for now we can leave it like this but in the future we would have to separate it we we need it.

gerard-queralt: Another thing to consider is what both characters can do, which I don't know so maybe my point is moot. But if we want to make it so, say, only Bix can dash because she is more agile or something, it'd be better to have them separated. Again, maybe this doesn't apply to our game, I'm not up to date with the mechanics.

markmeliaa: The basic moveset will be the same for both Bix and Allura, the only thing that will differentiate them is that one of them has melee attacks and the other one ranged attacks (that's why the attack will go in a separate file).

DADSILVER:
> Commenting here because after the merge with develop I can't view the files changed. I noticed that the destructor of `PlayerMobilityScript` calls `delete` on one of the components of the owner of the script. You shouldn't do this, this memory is owned by the game object.

Changed, thx.

Game/Scripts/PlayerMoveScript.cpp
DADSILVER: I see your point, but actually the problem we have is that the gravity calculation is in the rigbody but the calculation of not falling into the void is in the ScriptMove. And all the functions of the movement have this values in common. That's why I decided to keep them in the same script until we can separate them better.

markmeliaa (approved):

Game/Scripts/PlayerMoveScript.h
```diff
@@ -0,0 +1,43 @@
+#pragma once
+
+#include "Scripting\Script.h"
+
+// Little fix until we could check if an audio is being reproduced
```
MarcGonzalezMoratona, line 5:
```suggestion
// Small fix until we can check if an audio is being played
```

Game/Scripts/PlayerRotationScript.cpp
```diff
@@ -0,0 +1,53 @@
...
+	float deltaTime = App->GetDeltaTime();
+	ComponentTransform* trans = static_cast<ComponentTransform*>(owner->GetComponent(ComponentType::TRANSFORM));
+	float3 newRot = trans->GetRotationXYZ();
+	ModuleInput* input = App->GetModule<ModuleInput>();
+	newRot.y += -input->GetMouseMotion().x * deltaTime;
+	trans->SetRotation(newRot);
+	trans->UpdateTransformMatrices();
+
+
+	//Corroborate that you don't fuse with a owner
```
MarcGonzalezMoratona, line 38: What do you mean with "fuse"?
```suggestion
	//Corroborate that you don't fuse with an owner
```

PR #353, opened 2023-05-08
feat: Bullet library integration

OriolSors:
- Bullet library to our project
- New ModulePhysics that will manage all the collisions between rigidbodies
- Same PR than #350 and #341


Source/ModulePhysics.h
gerard-queralt, line 1: Both this file and the source should be inside the `Modules/` folder.

gerard-queralt (changes requested):

gerard-queralt (approved):

totisama (approved):

PR #354, opened 2023-05-08
feat: Remove/Load new scripts from the graphical interface improved

markmeliaa:
Not improved in a performance way, but in a visual way:

- The reset script button has been changed to a remove button.
- The list of scripts is not shown if a script is already loaded.
- Fixed a couple warnings (C-style casts and const variables)

### DIFF
--- a/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
+++ b/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
@@ -100,7 +100,7 @@ void WindowComponentScript::DrawWindowContents()
 				
 					label = stringField.name;
 					finalLabel = label + separator + thisID;
-					if (ImGui::InputText(finalLabel.c_str(), const_cast<char*>(value.c_str()), 24))
+					if (ImGui::InputText(finalLabel.c_str(), value.data(), 24))
 					{
 						stringField.setter(value);
 					}
--- a/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
+++ b/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
@@ -35,17 +35,21 @@ void WindowComponentScript::DrawWindowContents()
 	std::string thisID = std::to_string(windowUID);
 	std::string finalLabel = label + separator + thisID;
 
-	if (ImGui::ListBox(finalLabel.c_str(), &current_item, constructors.data(), (int)(constructors.size()), 5))
+	const IScript* scriptObject = script->GetScript();
+
+	if (!scriptObject)
 	{
-		if (script->GetConstructName() != constructors[current_item])
+		if (ImGui::ListBox(finalLabel.c_str(), &current_item, constructors.data(), static_cast<int>(constructors.size()), 5))
 		{
-			ChangeScript(script, constructors[current_item]);
-			ENGINE_LOG("%s SELECTED, drawing its contents.", script->GetConstructName().c_str());
+			if (script->GetConstructName() != constructors[current_item])
+			{
+				ChangeScript(script, constructors[current_item]);
+				ENGINE_LOG("%s SELECTED, drawing its contents.", script->GetConstructName().c_str());
+			}
 		}
 	}
 
-	const IScript* scriptObject = script->GetScript();
-	if (scriptObject)
+	else
 	{
 		ImGui::Separator();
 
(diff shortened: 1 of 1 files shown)


Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
```diff
@@ -63,14 +64,14 @@ void WindowComponentScript::DrawWindowContents()
...
 		finalLabel = label + thisID;
 		if (ImGui::Button(finalLabel.c_str()))
 		{
-			ComponentScript* newScript = static_cast<ComponentScript*>(component);
+			ENGINE_LOG("%s REMOVED, showing list of available scripts.", script->GetConstructName().c_str());
 
-			ChangeScript(newScript, newScript->GetConstructName().c_str());
-			ENGINE_LOG("%s RESET, drawing its contents again.", newScript->GetConstructName().c_str());
+			script->SetScript(nullptr); // This deletes the script itself
+			script->SetConstuctor("");	// And this makes that it is also deleted from the serialization
```
gerard-queralt, line 74: Since it looks like both of these things are necessary in order to remove a script object, we could consider creating a `ComponentScript::RemoveScript` function that encapsulates them. Though not sure is this behaviour will be needed elsewhere in the code, might not be necessary.

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
```diff
@@ -99,7 +100,7 @@ void WindowComponentScript::DrawWindowContents()
 				
 					label = stringField.name;
 					finalLabel = label + separator + thisID;
-					if (ImGui::InputText(finalLabel.c_str(), (char*)(value.c_str()), 24))
+					if (ImGui::InputText(finalLabel.c_str(), const_cast<char*>(value.c_str()), 24))
```
gerard-queralt, line 103: `const_cast` are generally avoidable (and it's a bad sign if they can't be avoided), I think this can be replaced with
```suggestion
					if (ImGui::InputText(finalLabel.c_str(), value.data(), 24))
```
Though this is only available since C++17. In that version, `data` returns a non-const `char*` for a non-const `std::string`, whereas in previous versions it always returned a `const char*`. More information in [the documentation](https://en.cppreference.com/w/cpp/string/basic_string/data).

gerard-queralt (commented):
Thank you for your work Mark, everything worked when I tested. Left a few suggestion but nothing that important, feel free to discuss them ;)

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
```diff
@@ -35,17 +35,18 @@ void WindowComponentScript::DrawWindowContents()
...
+		if (ImGui::ListBox(finalLabel.c_str(), &current_item, constructors.data(), static_cast<int>(constructors.size()), 5))
 		{
 			ChangeScript(script, constructors[current_item]);
 			ENGINE_LOG("%s SELECTED, drawing its contents.", script->GetConstructName().c_str());
 		}
 	}
 
-	IScript* scriptObject = script->GetScript();
-	if (scriptObject)
+	else
```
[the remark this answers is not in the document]
markmeliaa: Good catch!

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
markmeliaa: I don't think that we will use it elsewhere either, so I did not found the need to create a function to encapsulate them. What I don't get is, why in the script serialization its is only saved the name and not the fact that the script is nullptr?

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
markmeliaa: Felt wrong doing a `const_cast` though, thanks for pointing out a better solution!

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
gerard-queralt: Just to make sure, you are talking about this line in `ComponentScript::SaveOptions`, right? We have both a `ComponentScript` and a plain `Script`, and both have serialization methods, so it gets a bit confusing, sorry ^^'
```c++
meta["constructName"] = this->constructName.c_str();
```
I think the idea was that this was enought information to load the script. When de-serializing (LoadOptions) we don't check the contents of the string, we just ask the factory for the corresponding `Script` instance. If the string is empty we obviously can't find it, but to us who call the function, the important thing is that there's no script with this name. Basically, I don't think we need to store in the `.meta` whether we have/had a script or not. Let me know if you see my point or you're still not convinced, it's very possible I'm tunnel-visioning and missing something :P

gerard-queralt (approved):

DADSILVER (approved):
Goob job

Issue #217, opened 2023-03-20
Quadtree wants to render an object more than one time


PaablooCH:
Implementing the efficient rendering we encounter that the function FillRenderList process more objects than the actual number of objects inside the scene.
For example in the actual scene there are almost 220 objects but what the quadtree return is more than that:
![image](https://user-images.githubusercontent.com/62140600/226432455-7ac55468-a0b9-4388-be63-590b83991e68.png)

This means that actually we are rendering the same object more than one time and implementing efficient rendering cause that the scene doesn't load properly.

PR #355, opened 2023-05-08
perf: Quadtree refactor

Dedado3000:
Full rework of quadtree based on Task #865c2x27z
- No more duplicated render
- Remove "Shared GameObject", only object at leaves
- More eficient rendering selected game object


gerard-queralt (commented):
Do we know the cause of #217? Because a GameObject shouldn't appear in the quadtree twice, even before this refactor, or am I wrong? I thought we fixed that.
Also, I don't know if the vector `drawnGameObjects` can be removed, since it was used to prevent a game object from being drawn twice, even if it appeared multiple times in the vector to drawn.

Dedado3000:
> added

In fact #217 was one of the focus of this refactor and is solved with this pr

Source/DataStructures/Quadtree.h
```diff
@@ -60,6 +60,8 @@ class Quadtree
 
 	std::list<GameObject*> GetAllGameObjects(GameObject* gameObject);
 
+	void RecalculateQuadtree();
```
PaablooCH, line 63: is this function called somewhere?

PaablooCH (changes requested):
Plus some code refactor and questions, i loaded the first scene and count the gameobject that has (i delete some like the whole blockout father and cameras) and compared them with the size of opaquesToDraw and is like 320~ vs 407, can you check it too?

Source/Modules/ModuleRender.h
```diff
@@ -54,7 +55,7 @@ class ModuleRender : public Module
 
 	unsigned vbo;
 	
-	std::vector<const GameObject*> opaqueGOToDraw;
+	std::set<const GameObject*> opaqueGOToDraw;
```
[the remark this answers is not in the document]
Dedado3000: Im looking for efficiency

Source/DataStructures/Quadtree.cpp
```diff
@@ -516,4 +465,31 @@ void Quadtree::LoadOptions(Json& meta) {
...
+
+/*
+Recalculate Quadtree
+Info: the Quadtree in which this function is called get all objects 
+from his structure, delete it childrens and redone the structure
+*/
+void Quadtree::RecalculateQuadtree()
+{
+	std::set<GameObject*> familyGameObjects;
+	this->GetFamilyObjects(familyGameObjects);
```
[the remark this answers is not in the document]
Dedado3000: well, this was deleted bc was unused.

Source/DataStructures/Quadtree.h
Dedado3000: I think that was something that was dropped after some conversation at physics meetings

Source/Modules/ModuleRender.h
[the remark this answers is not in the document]
Dedado3000: Okey, as we are deleting every frame and rebuilding is more eficiency using a normal Set, unordered_set is better in a case we will mantain the "vector" across the ticks and we will not rebuild it so frequently

Source/DataStructures/Quadtree.cpp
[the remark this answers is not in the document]
PaablooCH: > we don't have any standard about this, should we?

I think that we already have it, but i could be wrong, plus we have a task https://app.clickup.com/t/865bph92u where we deleted unnecessary `this->`

Source/Modules/ModuleRender.cpp
```diff
@@ -461,6 +461,40 @@ void ModuleRender::AddToRenderList(GameObject* gameObject)
...
+			continue;
+		}
+		if (gameObject->IsActive())
+		{
+			if (!CheckIfTransparent(gameObject))
+				opaqueGOToDraw.insert(gameObject);
+			else
+			{
+				const ComponentTransform* transform =
+					static_cast<ComponentTransform*>(gameObject->GetComponent(ComponentType::TRANSFORM));
```
OriolSors, line 485: Why are you retrieving the transform again?

Source/Modules/ModuleRender.cpp
```diff
@@ -461,6 +461,40 @@ void ModuleRender::AddToRenderList(GameObject* gameObject)
...
+				float dist = Length(cameraPos - transform->GetGlobalPosition());
+				while (transparentGOToDraw[dist] != nullptr)
+				{
+					float addDistance = 0.0001f;
```
OriolSors, line 489: Maybe we can start putting epsilon offsets like this in a Local configuration or something? Abro debate

Source/Modules/ModuleRender.h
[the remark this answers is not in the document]
PaablooCH: are u sure about that? for what i found it's the same as map/unordered_map and if we are rebuilding the structure every frame an unordered_set could fit better.

PaablooCH (dismissed):

Source/Modules/ModuleRender.cpp
Dedado3000: this is someone from graphics, you can speak with them but seems ok

Source/Modules/ModuleRender.cpp
Dedado3000: Mental lag, solved

PaablooCH (approved):

OriolSors (approved):

PR #356, opened 2023-05-08
feat: Game can be built throught menu

gerard-queralt:
New features:

- "Build" button in the Main Menu that allows you to choose a configuration (`Debug`/`Release`) and builds our game with it, including creating the Zip file.
- Scenes and shaders moved to Lib and included in the Zip automatically.
- WindowLoading can have multiple loads with the corresponding text and spinner.

Possible improvements:

- Instead of building directly, open a menu where you can choose the starting scene and whether to re-generate the Zip file or not. Currently, we err on the side of caution and **always.** regenerate the Zip. We could only do it if it doesn't exist, but I think it's better this way.
- ~Block all input from the user in the rest of the UI. Should be always done when `WindowLoading` is being drawn.~
- Allow both threads to be cancellable. Currently, we must wait for them to finish.
- Redirect console output from the threads to the main thread's output, so it can be seen through the graphical interface.
- ~Center the loading window. It can't be that hard, but I haven't managed to do it.~


markmeliaa (dismissed):
Everything looks and works correctly (both debug and release game built without a problem). Would it be possible to add a "Launch" button to execute either the debug or the release game after building them?

And about the possible improvements:
- Block all input from the user in the rest of the UI. Should be always done when `WindowLoading` is being drawn.
- Center the loading window. It can't be that hard, but I haven't managed to do it.

Both the save/load scene and the import scene dialogues appear centered and block the input of the user meanwhile they are opened, so maybe mimmicking what it's done there, those two improvements could be acomplished?

gerard-queralt:
> Would it be possible to add a "Launch" button to execute either the debug or the release game after building them?

It should be possible, I'll look into it. Thanks for the suggestion!

> Both the save/load scene and the import scene dialogues appear centered and block the input of the user meanwhile they are opened, so maybe mimmicking what it's done there, those two improvements could be acomplished?

The issue is that window is managed by an ImGui extension, it's not something we've done. But yes, we could try checking how that window works.

markmeliaa:
> The issue is that window is managed by an ImGui extension, it's not something we've done. But yes, we could try checking how that window works.

And what about a `PopUpModal`? It also disables the user inputs when open and appears on the center of the screen.

gerard-queralt:
> And what about a `PopUpModal`? It also disables the user inputs when open and appears on the center of the screen.

Huh, didn't think about that. I'll have a look, see if I can adapt the current window. Thank you!

markmeliaa (approved):
Looks great now!

AuxPort66 (approved):
Works well. We can have some improvements later like putting a window to select scenes and then load the list so we create the binarys instead of zipping what we have on the folder at that moment

PR #364, opened 2023-05-10
chore: Efficient rendering first merge with develop

MacXxs:
* Added batching for specular and metallic shaders
* Added highlightning in batching
* Added transparent sorting in batch
* Mantained drawing of components (other than MeshRenderer)
* Modified specular shader to use bindlessTextures

Due to this being a merge with a very old branch, the files changed has gone crazy and it states that are infinit files. The important files to check are:
* GeometryBatch.h/.cpp
* BatchManager.h/,cpp
* ModuleRender.h/.cpp

And if you want to also, you can look at the shaders for default and specular materials

TODO: Ones this PR is aproved we will do another merge with develop, and then we are going to merge Santi's work with animation bones and weights with batching


AlexisDieu (dismissed):

sncarbo (changes requested):
When I try to create a Cube and another 3D object the program crashes. This is the call stack:
ef1

AlexisDieu:
> When I try to create a Cube or another 3D object the program crashes. This is the call stack: ef1

Is the command.size() is equal to 0 when it crash ?

MacXxs:
> When I try to create a Cube or another 3D object the program crashes. This is the call stack: ef1

Hi Santi! Have you tried deleting Lib? For us is not crashing so maybe it's that. If not we are going to keep looking into it!

sncarbo:
> > When I try to create a Cube or another 3D object the program crashes. This is the call stack: ef1
> 
> Hi Santi! Have you tried deleting Lib? For us is not crashing so maybe it's that. If not we are going to keep looking into it!

I deleted `Lib` but still not working. I will keep looking.

AlexisDieu:
> > > When I try to create a Cube or another 3D object the program crashes. This is the call stack: ef1
> > 
> > 
> > Is the command.size() is equal to 0 when it crash ?
> 
> Yes, `command.size()` is 0.

Can you test with the new commit

sncarbo:
@MacXxs @AlexisDieu It no longer crashes, but I have found a weird bug. I created a box inside a loaded scene and, when I change the parameters of the box material, it seems that they are also changed in the environment. It is possible that the parameters of several mesh renderers in the same batch are changed at the same time.

https://github.com/Horizons-Games/Axolotl-Engine/assets/116657602/e3a73fb3-f41a-47e3-8508-8e9a335e1ed3

MacXxs:
> @MacXxs @AlexisDieu It no longer crashes, but I have found a weird bug. I created a box inside a loaded scene and, when I change the parameters of the box material, it seems that they are also changed in the environment. It is possible that the parameters of several mesh renderers in the same batch are changed at the same time.
> 
>  error.ef.mp4

Hi Santi! That's normal, the cube is using the same default material as the environment, that's why they change.

sncarbo:
@MacXxs You are right, the material is common to all GOs. There is another bug if you create a spot light and try to change it to point light in the spot light panel (or vice versa) it crashes. It's similar to the error that Pablo got in IBL, and it's solved in develop, but I don't know if it's the same. The call stack is as follows:

ef1

Source/DataModels/Batch/GeometryBatch.cpp
```diff
@@ -275,31 +400,98 @@ void GeometryBatch::DeleteComponent(ComponentMeshRenderer* componentToDelete)
...
 		App->resources->FillResourceBin(componentToDelete->GetMaterial());
 	}
 #endif //ENGINE
 }
 
+std::vector<ComponentMeshRenderer*> GeometryBatch::ChangeBatch(const ComponentMeshRenderer* componentToDelete)
+{
+	componentToMove.clear();
+
+	bool findMaterial = false;
```
PaablooCH, line 429: this boolean can be inside of the loop and then we don't need to set it to false each loop, or we can also delete this variable and put the if(findMaterial) inside of the first if

Source/DataModels/Batch/GeometryBatch.cpp
```diff
@@ -275,31 +400,98 @@ void GeometryBatch::DeleteComponent(ComponentMeshRenderer* componentToDelete)
...
+	dirtyBatch = true;
 #else
 		App->resources->FillResourceBin(componentToDelete->GetMaterial());
 	}
 #endif //ENGINE
 }
 
+std::vector<ComponentMeshRenderer*> GeometryBatch::ChangeBatch(const ComponentMeshRenderer* componentToDelete)
+{
+	componentToMove.clear();
```
PaablooCH, line 427: this class variable is used in other place? we can just create the vector here

Source/DataModels/Batch/GeometryBatch.cpp
```diff
@@ -319,107 +511,147 @@ void GeometryBatch::BindBatch(const std::vector<ComponentMeshRenderer*>& compone
...
+			if (!isRoot)
+			{
+				if (!selected) //(selected && (owner == selectedGo || selectedGo->IsADescendant(owner))))
+				{
+					if (owner != selectedGo && !selectedGo->IsADescendant(owner))
+					{
+						draw = true;
+					}
+				}
+				else if (selected)
```
PaablooCH, line 566:
```suggestion
				else
```

MacXxs:
> @MacXxs You are right, the material is common to all GOs. There is another bug if you create a spot light and try to change it to point light in the spot light panel (or vice versa) it crashes. It's similar to the error that Pablo got in IBL, and it's solved in develop, but I don't know if it's the same. The call stack is as follows:
> 
> ef1

Yeah this will be fixed with the second merge with develop. This first merge was with a version of develop of three/four weeks ago, so with the second merge after this PR is approved, it will be fixed!

Source/DataModels/Batch/GeometryBatch.h
```diff
@@ -1,11 +1,16 @@
 #pragma once
 
 #include "Math/float3.h"
+#include "Math/float2.h"
```
PaablooCH, line 4: this include is already in .cpp

Source/DataModels/Batch/GeometryBatch.h
```diff
@@ -1,11 +1,16 @@
...
 
 #include <vector>
+#include <map>
```
PaablooCH, line 7: this is not used

Source/DataModels/Batch/GeometryBatch.h
```diff
@@ -45,40 +64,53 @@ struct ResourceInfo
...
 public:
-	GeometryBatch();
+	GeometryBatch(int flag);
 	~GeometryBatch();
 
-	const int GetFlags() const;
-
 	void CreateVAO();
+	void ClearBuffer();
+	bool CleanUp();
```
PaablooCH, line 72: make it private

Source/DataModels/Batch/GeometryBatch.h
```diff
@@ -45,40 +64,53 @@ struct ResourceInfo
...
+	std::vector< ComponentMeshRenderer*> ChangeBatch(const ComponentMeshRenderer* componentToDelete);
 
+	void BindBatch(bool selected);
 	void FillMaterial();
-
 	void ReserveModelSpace();
 
-	bool CleanUp();
+	bool IsEmpty() const;
+	bool IsFillMaterials() const;
```
PaablooCH, line 87: not used

Source/DataModels/Batch/GeometryBatch.h
```diff
@@ -45,40 +64,53 @@ struct ResourceInfo
...
 
-private:
+	void SetFillMaterials(const bool fillMaterials);
+	void SortByDistance();
+
+	const int GetFlags() const;
+
+	bool dirtyBatch;
```
PaablooCH, line 94: make it private?

PaablooCH (changes requested):
well done, it costs a lot but we are close, a part of the changes  remember to delete comments and check the includes 👍

AlexisDieu (commented):
I fix the last issue I said with this

AlexisDieu:
image
Changing the scene make the engine crash (not happening before the commit again i verified)

MacXxs:
> image Changing the scene make the engine crash (not happening before the commit again i verified) It seems to enter in deleteComponent but the vector of component it's empty (It occur because i but back the batch->DeleteComponent(this); i think but else there was another issue and i just revert the change

Hi! I told you again from here so everybody knows. This wasn't happening before 'cause the batches were never destroyed because the batchmanager cleanup was inlined, and when doing deletes inside inline functions don't work, so the destructor of the objects to delete are never called. That's why you could change, remove and add stuff from a scene, 'cause the pointers to that batch in the componentMeshRenderer were pointing to batches that were never truly destroyed. I've fixed this so now everything is propertly being deleted.

AlexisDieu (approved):

sncarbo (approved):

PaablooCH (approved):

PR #366, opened 2023-05-11
feat: Create scripts into the Script folder using the engine's UI

markmeliaa:
Now, under the list of available scripts that you can choose in a `ComponentScript`, there is a button to create new scripts with any given name. Those scripts will go directly to your local Axolotl-Engine/Game/Scripts folder, and from there you'll only have to include them in the project through VS to be able to use them (we are working on doing that automatically, but for now we haven't found a way, so you'll have to do it manually for the time being).

### DIFF
--- a/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
+++ b/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
@@ -233,14 +233,25 @@ void WindowComponentScript::AddNewScriptToProject(const std::string& scriptName)
 	char* headerBuffer = nullptr;
 	char* sourceBuffer = nullptr;
 
-	unsigned int headerBufferSize = App->GetModule<ModuleFileSystem>()->Load("Source/PreMades/TemplateHeaderScript", headerBuffer);
-	unsigned int sourceBufferSize = App->GetModule<ModuleFileSystem>()->Load("Source/PreMades/TemplateSourceScript", sourceBuffer);
+	unsigned int headerBufferSize = 
+		App->GetModule<ModuleFileSystem>()->Load("Source/PreMades/TemplateHeaderScript", headerBuffer);
+	unsigned int sourceBufferSize = 
+		App->GetModule<ModuleFileSystem>()->Load("Source/PreMades/TemplateSourceScript", sourceBuffer);
 
 	std::string scriptsPath = "Scripts/";
 
 	std::string scriptHeaderPath = scriptsPath + scriptName + ".h";
 	std::string scriptSourcePath = scriptsPath + scriptName + ".cpp";
 
-	App->GetModule<ModuleFileSystem>()->Save(scriptHeaderPath.c_str(), headerBuffer, headerBufferSize);
-	App->GetModule<ModuleFileSystem>()->Save(scriptSourcePath.c_str(), sourceBuffer, sourceBufferSize);
+	// Both header and source have the same name, so only checking the header is enough
+	if (!App->GetModule<ModuleFileSystem>()->Exists(scriptHeaderPath.c_str()))
+	{
+		App->GetModule<ModuleFileSystem>()->Save(scriptHeaderPath.c_str(), headerBuffer, headerBufferSize);
+		App->GetModule<ModuleFileSystem>()->Save(scriptSourcePath.c_str(), sourceBuffer, sourceBufferSize);
+	}
+
+	else
+	{
+		ENGINE_LOG("That name is already in use, please use a different one");
+	}
 }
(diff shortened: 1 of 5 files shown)


Piterga12 (approved):
Tested and everything works fine! Really good job

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
```diff
@@ -183,4 +203,75 @@ void WindowComponentScript::ChangeScript(ComponentScript* newScript, const char*
 	Iscript->SetGameObject(component->GetOwner());
 	Iscript->SetApplication(App.get());
 	newScript->SetScript(Iscript);
+}
+
+void WindowComponentScript::OpenCreateNewScriptPopUp()
+{
+	static char name[FILENAME_MAX] = "NewScript";
```
gerard-queralt, line 210: Is there a reason for this variable to be `static`? Now, if you create a new script and open the window to create another one the name of the previous script will remain. Is that what we want?

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
```diff
@@ -183,4 +203,75 @@ void WindowComponentScript::ChangeScript(ComponentScript* newScript, const char*
...
+	std::string scriptSourcePath = scriptsPath + scriptName + ".cpp";
+
+	// Both header and source have the same name, so only checking the header is enough
+	if (App->GetModule<ModuleFileSystem>()->Exists(scriptHeaderPath.c_str()))
+	{
+		ENGINE_LOG("That name is already in use, please use a different one");
+		return;
+	}
+
+	ENGINE_LOG("New script %s created", scriptName.c_str());
```
gerard-queralt, line 242: I'd move this log after we save the file. This is just being picky, but we are logging this before the script is actually created :P

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
```diff
@@ -183,4 +203,75 @@ void WindowComponentScript::ChangeScript(ComponentScript* newScript, const char*
...
+
+void WindowComponentScript::AddNewScriptToProject(const std::string& scriptName)
+{
+	std::string scriptsPath = "Scripts/";
+
+	std::string scriptHeaderPath = scriptsPath + scriptName + ".h";
+	std::string scriptSourcePath = scriptsPath + scriptName + ".cpp";
+
+	// Both header and source have the same name, so only checking the header is enough
+	if (App->GetModule<ModuleFileSystem>()->Exists(scriptHeaderPath.c_str()))
```
gerard-queralt, line 236: We could store the result of `App->GetModule<ModuleFileSystem>()` to avoid repeated calls, but it's a really cheap call so it won't really affect performance to be fair.

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
```diff
@@ -183,4 +203,75 @@ void WindowComponentScript::ChangeScript(ComponentScript* newScript, const char*
...
+	ENGINE_LOG("New script %s created", scriptName.c_str());
+
+	char* headerBuffer = nullptr;
+	char* sourceBuffer = nullptr;
+
+	App->GetModule<ModuleFileSystem>()->Load("Source/PreMades/TemplateHeaderScript", headerBuffer);
+	App->GetModule<ModuleFileSystem>()->Load("Source/PreMades/TemplateSourceScript", sourceBuffer);
+
+	std::string headerBufferAsString = headerBuffer;
+	std::string sourceBufferAsString = sourceBuffer;
```
gerard-queralt, line 251: I think we need to `delete` both buffers after storing the value, since `Load` allocates memory so we need to free it. I could be wrong @AuxPort66

gerard-queralt (commented):
Thanks Mark, just a few suggestions.

markmeliaa:
> Whoops, too slow xD

My bad!! I can push the required changes directly into scripting-main if you'd like.

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
markmeliaa: Not it is not indeed, I think i was tweaking around trying some things and forgot to change this back, thanks for noticing!

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
markmeliaa: But these buffers are allocated in the stack, shouldn't they get deleted automatically?

gerard-queralt:
> My bad!! I can push the required changes directly into scripting-main if you'd like.

Don't worry, I'll do it myself when fixing merge conflicts, let me know if you agree with them ;)

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
gerard-queralt: Sorry, I meant the `*Buffer`, not the `*BufferAsString`. The latter does get removed automatically, but I'm not sure about the former.

markmeliaa:
> Don't worry, I'll do it myself when fixing merge conflicts, let me know if you agree with them ;)

I do!

PR #367, opened 2023-05-11
Feat: Alpha discard

AuxPort66:
I've added alpha discard to be able to have grids and stuff per texture. The art department requested it asap.

So here it is


markmeliaa (approved):

AuxPort66:
Well I was waiting to the lead of graphics to see it because yesterday he commented that maybe there is a better way than the discard...

markmeliaa:
Sorry!! I though that as there were no pending reviewers it was good to go.

PR #369, opened 2023-05-12
feat: Collisions, forces and gravity with Bullet

OriolSors:
- Implementation of the features described
- New test script as example of the functionalities

Functions for movement tested:

1. AngularFactor
2. AngularVelocity
3. LinearVelocity
4. ApplyCentralImpulse


Dedado3000 (approved):

xavi1111 (commented):
Changes seem great. I tested loading the scene and starting play mode everything works fine here, physics seem to work great. Then I try to delete objects from hierarchy and also works but when I try to put the cube inside the charachter or otherwise it crashes I don't know if it's only happening to me or is a bug.

Source/NewPlayerMobilityTest.cpp
markmeliaa, line 1: The scripts should go into the Game/Scripts folder, could you move it there please. It's just for the sake of organization.

markmeliaa:
> Changes seem great. I tested loading the scene and starting play mode everything works fine here, physics seem to work great. Then I try to delete objects from hierarchy and also works but when I try to put the cube inside the charachter or otherwise it crashes I don't know if it's only happening to me or is a bug.

It happens to me too, I've checked the stack and in theory it has nothing to do with physics, so it seems out of this scope (?)

markmeliaa:
Thank you @totisama!

Dedado3000 (approved):

PR #370, opened 2023-05-13
Fix: Join Save/Load Scene functions

AuxPort66:
This PR is to centralize in one function the Save Scene code and the Load Scene code. 

Also solved a problem with on the OnPlay and the ImportScene because the code wasn't updated with the last changes on the scene, like the CubeMap or the disappearance of the ambient light. I hope that using all the same functions does not happen again

### DIFF
--- a/Source/Modules/ModuleScene.h
+++ b/Source/Modules/ModuleScene.h
@@ -33,7 +33,7 @@ public:
 	bool hasNewUID(UID oldUID, UID& newUID);
 	void SetSceneRootAnimObjects(std::vector<GameObject*> gameObjects);
 
-	void SaveSceneToJson(const std::string& name);
+	void SaveScene(const std::string& name);
 	void LoadSceneFromJson(const std::string& name);
 	void ImportFromJson(const std::string& name);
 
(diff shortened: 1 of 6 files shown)


Source/Modules/ModuleScene.h
```diff
@@ -33,9 +33,8 @@ class ModuleScene : public Module
 	bool hasNewUID(UID oldUID, UID& newUID);
 	void SetSceneRootAnimObjects(std::vector<GameObject*> gameObjects);
 
-	void SaveSceneToJson(const std::string& name);
-	void LoadSceneFromJson(const std::string& name);
-	void ImportFromJson(const std::string& name);
+	void SaveScene(const std::string& name);
+	void LoadScene(const std::string& name, bool mantainActualScene);
```
gerard-queralt, line 37: Maybe we could default the bool to false? Since I imagine is the most common use case for this method.
```suggestion
	void LoadScene(const std::string& name, bool mantainActualScene = false);
```
Just being picky though.

gerard-queralt (dismissed):

Dedado3000 (approved):

MarcGonzalezMoratona (approved):

gerard-queralt (approved):

PR #371, opened 2023-05-13
feat[Gameplay]: Health system

csandra24:


Game/Scripts/HealthSystem.h
```diff
@@ -0,0 +1,24 @@
...
+	float GetHealth() const;
+	void SetHealth(float health);
+
+	void TakeDamage(float damage);
+	void HealLife(float amountHealed);
+
+
+private:
+	float health;
+	float maxHealth;
```
DADSILVER, line 23: Why are we using float instead of int?

DADSILVER (commented):
Great job!! Only one question.

PaablooCH (approved):

csandra24 (commented):
Good!

Game/Scripts/HealthSystem.cpp
```diff
@@ -0,0 +1,51 @@
...
+void HealthSystem::TakeDamage(float damage)
+{
+	health = health - damage;
+}
+
+void HealthSystem::HealLife(float amountHealed)
+{
+	health = health + amountHealed;
+
+	if (health > maxHealth)
```
gerard-queralt, line 37: Just a tip, but instead of doing this `if`, you could change this block to
```c++
health = std::max(health, maxHealth);
```
If you want to have it even more condensed, it could even be
```c++
health = std::max(health + amountHealed, maxHealth);
```

DADSILVER: I think this is a good idea, but you want to use Min instead of Max. 

`
void HealthSystem::HealLife(float amountHealed)
{
    health = std::min(health + amountHealed, maxHealth);
}
`
And then, as Gerard said, you don't need to use the "if".

Game/Scripts/HealthSystem.cpp
```diff
@@ -0,0 +1,51 @@
...
+REGISTERCLASS(HealthSystem);
+
+HealthSystem::HealthSystem() : Script(), health(100), maxHealth(100)
+{
+	REGISTER_FIELD(health, float);
+}
+
+void HealthSystem::Start()
+{
+	maxHealth = health;
```
UriMurcia, line 16: maxHealth may not be always the health? I would just delete this line and leave maxHealth as 100

markmeliaa: At the time the `Start()` is executed (which is at the beginning of everything), maxHealth will always be the health (and health can be modified through the inspector, so we can't set maxHealth to a fixed value because it will not always be the same value for every entity that we attach the HealthSystem script).

UriMurcia: Ah I see, well okay, so then it should be like: health = maxHealth;

markmeliaa: Not at all, because `health` is the parameter that the user will change through ImGui, not `maxHealth`, so we have to keep track of the health that the user wrote in the inspector and set it as `maxHealth`, so later in the healing function no entity could be healed above that `maxHealth`.

gerard-queralt: But shouldn't the user set the max health, rather than the health itself?

markmeliaa: The user is indeed deciding the maxHealth, that is the point of this line, but I find that setting the health as public (a value that will change and that may be useful to be visible because it will help to see if the health is being updated correctly) is better than setting the maxHealth as public (a value that will not change and that will not help in any way by having it visible).

Also, both setting maxHealth as public or health as public will require to create another private variable to monitorize the health or the maxHealth respectively, so we will have this same line of code but inverted, as Uri suggested before.

UriMurcia: I didn't see that you only Registered_Field the var Health, I think we should Register both Max and Health to have the control of both as I said in the last comment. Indeed, it will help having the MaxHealth value Registered, as programmers we should give the maximum information to designers, and is both important to give them the freedom to control health and maxHealth. MaxHealth shouldn't be restricted by how much health the entity has, but yes the other way (in the start, as you did).
And you say MaxHealth will not change, but actually you are changing it from the variable health, which is not an intuitive praxis.

markmeliaa: Fine, let's agree to disagree, I will perform the required changes. 
`maxHealth`, once assigned in the `Start()`, never changes, were do you see it changing?

UriMurcia: from the editor

markmeliaa: What do you mean from the editor?

markmeliaa: Well, whatever, the changes are done, take a look and lmk what you think now.

PaablooCH (approved):

DADSILVER (approved):

PR #378, opened 2023-05-15
feat[Level design]: new level 1

Dedado3000:
The base of Finished Sewers and New Bock-Out of Cantina ready to gameplay use


AuxPort66 (approved):

markmeliaa (commented):
When I hit play, the selected camera is not Bix's camera, shouldn't it be the default camera, when you hit play, the camera from the player? Also, Bix has gravity and collides with the floor? I can't try it, but we had some problems with the old physics regarding that.

Dedado3000:
> When I hit play, the selected camera is not Bix's camera, shouldn't it be the default camera, when you hit play, the camera from the player? Also, Bix has gravity and collides with the floor? I can't try it, but we had some problems with the old physics regarding that.

Yeah, I only added a basic camera into bix only to make sure that all there wasn't a crash in debugGame, no playable character was done because it will cause problems with physics changes that will come at the same time with develop

AuxPort66:
Moved to the physics PR to speed up the process of finishing the level with the rigidbodys from the new library, it will merge to develop with to #385

PR #382, opened 2023-05-16
feat: Modify the size of a collider in the ImGui

totisama:


Source/Modules/ModulePhysics.h
```diff
@@ -73,4 +78,18 @@ class ModulePhysics : public Module
...
\ No newline at end of file
+
+	bool drawWorld = false;
+	int drawableRigidBodies = 0;
+
+};
+
+inline bool ModulePhysics::GetDrawWorld() const
+{
+	return drawWorld;
```
OriolSors, line 89: Where is this used?

totisama: For now it's not used, but we should add this flag on the modulePhysics PreUpdate instead of the App->IsOnPlayMode() to know when to debug or not. It will be so we can set this variable when we are on GameMode and want to debug the physics from the debug Window.

PR #383, opened 2023-05-17
Feat/Fix: Animations and StateMachine Updates

AuxPort66:
**Animation changes**
- Fixed the bug that caused a animated gameObject to steal pieces of other animated gameObject (When they are the same fbx), the palette used in the shader was not initialized correctly.
- This bug also do that when you load a mesh after a animated mesh the animated mesh stop rendering and desappears.
- Animation component is now removable.
- Fixed a bug in Play (ControllerAnimation), the time did not initialize correctly when you play new animations.
--------------------------------
**State Machines Changes**
- Added to the states machines the possibility to decide if a resource is looped or not and to wait for a resource to finish before executing a transition.
- The StateMachine Default and the Save/Load have been updated, so the created statemachines will have to be redone. If any of them are very complex, let me know and I can do it by code (Only if is really necessary)
- Added the name of the state machine to the WindowStateMachineEditor to make it clearer what you are editing.


markmeliaa (approved):
Nice changes! We didn't have many state machines done, so it will be easy to redo them!

MacXxs (approved):
Works perfectly! Nice catch with the palette binding👍🏼

Issue #397, opened 2023-05-18
Can load a new scene when play is runnning


Vlukkee:
When you are in play mode you can use tabs bars and others options, and thats make breake the engine.

Should be block the tabs or the functions when you are in play mode.

https://github.com/Horizons-Games/Axolotl-Engine/assets/25518155/d15165d1-a5be-408d-9a45-93096d593935

Issue #396, opened 2023-05-18
Ctrl + Q while the framerate screen is open crashes


OriolSors:
![PlayModeFramerateError](https://github.com/Horizons-Games/Axolotl-Engine/assets/31882514/ffec2a20-04fb-487e-bbc1-0095b66c3579)
![PlayModeFramerateError_VS](https://github.com/Horizons-Games/Axolotl-Engine/assets/31882514/f6b8c22b-e24f-42d0-a9d9-9920f19c1f8d)

Issue #394, opened 2023-05-18
Selecting an object in the scene is not displayed in the hierarchy


Vlukkee:
Selecting an object in the scene is not displayed in the hierarchy, and in the hierarchy is really hard to find the object that you want to work if you dont know it.
Here an example of the hierarchy of the cantina sewers, and below a  video.

The ideal is select an object in the scene and the hierarchy window show it in blue.

![2023-05-18_05h15_59](https://github.com/Horizons-Games/Axolotl-Engine/assets/25518155/d6625cc2-ff4d-4225-ad08-82b3b1304445)
![2023-05-18_05h16_12](https://github.com/Horizons-Games/Axolotl-Engine/assets/25518155/f64389d9-5e16-4344-875e-77702aca57ac)
![2023-05-18_05h16_23](https://github.com/Horizons-Games/Axolotl-Engine/assets/25518155/97d56cb5-43ec-4303-b581-1e08c41b4a0a)

https://github.com/Horizons-Games/Axolotl-Engine/assets/25518155/e6c0b62a-6330-4b94-943a-29c81ce5fbd9

Issue #392, opened 2023-05-18
New scene windows advice cant be canceled.


Vlukkee:
When you click a new scene, the prompt window cannot be canceled and if you try to save it and cancel the save, the old scene will be deleted.

![2023-05-18_02h30_44](https://github.com/Horizons-Games/Axolotl-Engine/assets/25518155/8ca97831-84d3-4fdc-b920-3af3aadb7d87)

![2023-05-18_04h52_37](https://github.com/Horizons-Games/Axolotl-Engine/assets/25518155/11d431a2-4e1f-478a-ad49-3948f1f7fd8d)

PR #400, opened 2023-05-18
fix: various fixes to reported bugs

gerard-queralt:
- https://github.com/Horizons-Games/Axolotl-Engine/issues/392
- https://github.com/Horizons-Games/Axolotl-Engine/issues/389
- https://github.com/Horizons-Games/Axolotl-Engine/issues/394
- https://github.com/Horizons-Games/Axolotl-Engine/issues/397
- https://github.com/Horizons-Games/Axolotl-Engine/issues/396


Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
Vlukkee, line 1: I check this one and works fine, the problem now is i can´t close any parent or anything in the herarchy


Uploading 2023-05-18_21h01_07.mp4…

gerard-queralt: Huh, I thought I fixed that. Thanks, I'll take another look.

Source/DataModels/Windows/SubWindows/WindowFPS.cpp
```diff
@@ -20,8 +20,23 @@ void WindowFPS::DrawWindowContents()
...
 	float deltaTime = App->GetDeltaTime();
-	fpsHist[currentFpsIndex] = 1.f / deltaTime;
-	timeHist[currentTimeIndex] = deltaTime * 1000.f;
+
+	{
+		float actualValue = 1.f / deltaTime;
+		fpsHist[currentFpsIndex] = std::max(actualValue, 0.0001f);
+		if (actualValue != fpsHist[currentFpsIndex])
+		{
+			ENGINE_LOG("Actual framerate was less than 0.0001");
```
MarcGonzalezMoratona, line 29:
```suggestion
			ENGINE_LOG("Current framerate was less than 0.0001");
```

Source/DataModels/Windows/SubWindows/WindowFPS.cpp
```diff
@@ -20,8 +20,23 @@ void WindowFPS::DrawWindowContents()
...
+		}
+	}
+	{
+		float actualValue = deltaTime * 1000.f;
+		timeHist[currentTimeIndex] = std::min(actualValue, 1000.f);
+		if (actualValue != timeHist[currentTimeIndex])
+		{
+			ENGINE_LOG("Actual frame time was more than 1000 seconds");
```
MarcGonzalezMoratona, line 37:
```suggestion
			ENGINE_LOG("Current frame time is more than 1000 seconds");
```

Source/DataModels/GameObject/GameObject.h
```diff
@@ -148,14 +146,21 @@ inline UID GameObject::GetUID() const
...
-			parent->SetStateOfSelection(StateOfSelection::NO_SELECTED);
-		}
+		return;
+	}
+	switch (stateOfSelection)
+	{
+	case StateOfSelection::NO_SELECTED:
+		parent->SetStateOfSelection(StateOfSelection::NO_SELECTED);
+		break;
+	case StateOfSelection::SELECTED:
```
MarcGonzalezMoratona, line 159: why this condition is empty?

markmeliaa (dismissed):
Tried the bugs a couple times and everything seems to be fixed, good job!

Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
Vlukkee: Thanks! Works fine, but you need to click two times to close the hierarchy. We can live with that but i will report it for dont forget it.

Vlukkee (dismissed):
Works fine, but you need to click two times to close the hierarchy. We can live with that but i will report it for dont forget it.

Source/DataModels/Windows/SubWindows/WindowFPS.cpp
gerard-queralt: I'm not sure I agree. This log is meant to indicate that even if we've displayed that the framerate was 0.0001, that was not true, and the real/actual value was lower. Let me know why you prefer it that way, I'm open to change it :)

Source/DataModels/GameObject/GameObject.h
gerard-queralt: Switch statements work by jumping the execution to the line that matches the value being switched on. In other words, if the value of the selection evaluates to `StateOfSelection::SELECTED`, we jump to line 159. We usually add a break/return at the end of each case because, otherwise, the execution would continue, which is generally not what we want. In this case, however, we want the code executed for both cases to be the same, so even if the case is empty, it will still run the code in the other case. If we were to use `if` statements, this would be equivalent to doing
```c++
if (stateOfSelection == StateOfSelection::SELECTED || stateOfSelection == StateOfSelection::CHILD_SELECTED)
```
Hope that clears things up! As always, if you want a more in-depth explanation, I encourage you to check out the [official documentation](https://en.cppreference.com/w/cpp/language/switch) ;)

gerard-queralt:
> you need to click two times to close the hierarchy

Should be fixed now, take another look when you can.

Source/DataModels/Windows/SubWindows/WindowFPS.cpp
MarcGonzalezMoratona: It's ok, don't worry

Source/DataModels/GameObject/GameObject.h
MarcGonzalezMoratona: This is really useful. Thank you!

markmeliaa (dismissed):

AuxPort66:
Selecting a gameObject on the hierarchy isn't working properly. Check it Please. The rest seems fine. Thank you.

https://github.com/Horizons-Games/Axolotl-Engine/assets/93783740/14654053-bb4d-487b-b718-53640f5d1b46

markmeliaa (approved):

AuxPort66 (approved):

PR #402, opened 2023-05-19
fix: UI Draw Hierarchy Recursive

AuxPort66:
- UI draw hierarchy recursively
- Fix Enable bug: UI only check the canvas enable not the rest of the gameObjects, and ignore the Component enable.


Vlukkee (approved):
Very nice! Very Guapo!

Dedado3000 (approved):

PR #403, opened 2023-05-20
fix: Detect disabled Interaction UI element

AuxPort66:
- Disable gameObjects put the hover and clicked booleans to false of any Component Button on that gameObject or children gameObject.
- Check Disable/Enable when check Interaction between mouse and Buttons


Vlukkee (dismissed):
Very good! Very guapo!

AuxPort66:
Fixed

markmeliaa (approved):

Vlukkee (approved):
Very good! Very guapo! x2

PR #405, opened 2023-05-20
fix: diagonal solved

Dedado3000:
This was the bug that was making infinite loop if height was superior to minquadrant size (this bug was since start but we didn't notice)
Sin título


Vlukkee (approved):
Very good! Very Guapo!

AuxPort66 (dismissed):

AuxPort66 (approved):

Vlukkee (approved):

PR #407, opened 2023-05-22
feat: improve logging

gerard-queralt:
- Use generic format instead of C format for logging
    - Allows us to define what we can log, like Resources or GameObjects
- Define severity of log message, console line colored accordingly
- Write all log output to a file (`Game/Documents/Axolotl.log`)
    - Easier to detect errors, we can send each other this file when encountering one
-  Console window can be filtered and cleared

Important classes to review:
- `AxoLog` (renamed from `EngineLog`)
- `WindowConsole`

EDIT: Also made logging thread-safe.


Dedado3000 (approved):
Good working in engine and play mode.

markmeliaa (approved):
Nice change! Only one question, what should be the difference between `INFO_LOG`, `VERBOSE_LOG` and `DEBUG_LOG`?

PR #410, opened 2023-05-22
fix: disable interaction with scene if it's not focused AND hovered

gerard-queralt:
Guizmo is still highlighted, but it can't be interacted with. Ideally it wouldn't even highlight.


Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
```diff
@@ -240,6 +218,38 @@ bool WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject)
...
+		{
+			CutAnObject();
+		}
+		if (input->GetKey(SDL_SCANCODE_D) == KeyState::DOWN)
+		{
+			DuplicateAnObject();
+		}
+	}
+
+	// Delete a GameObject with the SUPR key
```
MarcGonzalezMoratona, line 246: I think we can remove this comment, I know it's from another PR but the code is self-exaplanatory

MarcGonzalezMoratona (commented):
I haven't tested the changes yet, but the code looks good. I will give the final approve once everything is tested. I added some doubts and review comments, but not related to functionality.

Source/DataModels/Windows/EditorWindows/WindowScene.cpp
```diff
@@ -34,7 +34,7 @@ void WindowScene::DrawWindowContents()
 {
 	ManageResize();
 
-	ImGui::Image((void*) App->GetModule<ModuleRender>()->GetRenderedTexture(),
+	ImGui::Image(reinterpret_cast<void*>(static_cast<intptr_t>(App->GetModule<ModuleRender>()->GetRenderedTexture())),
```
MarcGonzalezMoratona, line 37: Could you explain this change? What is this cast doing?

gerard-queralt: In C++, everything can be cast to a `void*`. A pointer of this type represents a pointer to an address of memory, just like the other pointers, but unlike them these are not strongly typed. `ImGui::Image` requires as a parameter a `void*`, so we need to cast the `unsigned int` to one. It's recommended that, instead of doing a direct cast to `void*`, you first cast it to `intptr_t`/`uintptr_t`, which we previously didn't do. As far as I understood, this is safer because it doesn't rely on the size of the types being the same, which might not be the case with different architectures.
So, what we do is first do a cheap `static_cast` to convert our `unsigned int` to a `intptr_t` (which I just noticed should be an `uintptr_t`). Now it is safe to convert to a `void*`. But we cannot do a `static_cast`. What we want is the compiler to think that this value is a `void*`; in other words, reinterpret it. `reinterpret_cast` is the most dangerous type of cast and should generally be avoided, because we are telling the compiler to consider a given memory address as if it were a different type, which is obviously a dangerous thing to do. However, it's perfectly fine to do it in this case, because as I said in the beginning, everything can be interpreted as a void pointer.
Hopefully I explained myself! Here's some links in case you want a better explanation of this whole ordeal:
- [StackOverflow: Casting int to void pointer](https://stackoverflow.com/questions/19818732/casting-int-to-void-pointer)
  - There are many SO questions covering this topic, this is just the first result
- [ImGui wiki: how to display an image for OpenGL users](https://github.com/ocornut/imgui/wiki/Image-Loading-and-Displaying-Examples#Example-for-OpenGL-users) (bottom code block)
- [C++ documentation: static_cast](https://en.cppreference.com/w/cpp/language/static_cast)
- [C++ documentation: reinterpret_cast](https://en.cppreference.com/w/cpp/language/reinterpret_cast)
- [StackOverflow: when to use each of the four C++ casts](https://stackoverflow.com/questions/332030/when-should-static-cast-dynamic-cast-const-cast-and-reinterpret-cast-be-used)

markmeliaa (dismissed):
Everything looks and works correctly, gj once again.

Dedado3000 (dismissed):

Dedado3000 (approved):

markmeliaa (approved):

PR #412, opened 2023-05-22
perf: drone behaviour scripts improved

markmeliaa:
This branch was originally intended to adapt the drone scripts into Bullet, but that was already done by the physics team (thanks physics team).

Instead, the drone scripts were reworked and now they do not need to call the Update() of each behaviour's script, but rather a function inside each one of those behaviour's scripts.

Also, you can check the drone working (it is a capsule now because I couldn't find the drone fbx in the Assets folder) in the DroneTrial.axolotl scene.

---------

**CURRENTLY WORKING ON APPLYING THE ROTATION TO THE DRONE CORRECTLY**


diegusmejia117 (dismissed):
Just tested it, it works as intended, just a question, if we want to use more waypoint for example we have to register them as field because we dont have vectors still?

markmeliaa:
> Just tested it, it works as intended, just a question, if we want to use more waypoint for example we have to register them as field because we dont have vectors still?

Exactly, once we have vectors the waypoints will be a vector of transforms instead of individual variables, so we could add as many as we want.

AuxPort66 (approved):

Game/Scripts/PatrolBehaviourScript.cpp
```diff
@@ -25,50 +28,59 @@ void PatrolBehaviourScript::Start()
...
-
-GameObject* PatrolBehaviourScript::GetWayPointTwo() const
-{
-	return wayPointTwo;
-}
-
-void PatrolBehaviourScript::SetWayPointTwo(GameObject* wayPointTwo)
-{
-	this->wayPointTwo = wayPointTwo;
+		dd::arrow(ownerTransform->GetGlobalPosition(),
```
UriMurcia, line 80: ifdef debug to avoid drawing arrows in game

UriMurcia (commented):
Nice work!

UriMurcia (approved):

PR #413, opened 2023-05-23
Fix: Inspector Fixes and Copy RigidBody

AuxPort66:
- Inspector thought that the InputText of the gameObject name was the same as the InputText of the last selected gameObject, so when you have the InputText selected and you delete the gameObject. The newly selected GameObject (the parent) is renamed to the name of the removed gameObject. ImGui is special
- Bug that does not allow to put a gameObject name longer than the one that already existed even if it did not exceed the maximum characters
- RigidBody CopyConstructor finished

### DIFF
--- a/Source/DataModels/Components/ComponentRigidBody.cpp
+++ b/Source/DataModels/Components/ComponentRigidBody.cpp
@@ -44,36 +44,27 @@ ComponentRigidBody::ComponentRigidBody(bool active, GameObject* owner) :
 }
 
 ComponentRigidBody::ComponentRigidBody(const ComponentRigidBody& toCopy) :
-	Component(ComponentType::RIGIDBODY, toCopy.active, toCopy.owner, true)
+	Component(ComponentType::RIGIDBODY, toCopy.active, toCopy.owner, true),
+	isKinematic(toCopy.isKinematic),
+	isStatic(toCopy.isStatic),
+	currentShape(toCopy.currentShape),
+	boxSize(toCopy.boxSize),
+	radius(toCopy.radius),
+	factor(toCopy.factor),
+	height(toCopy.height),
+	usePositionController(toCopy.usePositionController),
+	useRotationController(toCopy.useRotationController),
+	KpForce(toCopy.KpForce),
+	KpTorque(toCopy.KpTorque),
+	mass(toCopy.mass)
 {
 	id = GenerateId();
 
 	transform = toCopy.transform;
-	boxSize = toCopy.boxSize;
-	radius = toCopy.radius;
-	factor = toCopy.factor;
-	height = toCopy.height;
 
-	currentShape = toCopy.currentShape;
 	motionState = std::unique_ptr<btDefaultMotionState>(new btDefaultMotionState(*toCopy.motionState.get()));
-	switch (static_cast<Shape>(currentShape))
-	{
-		default:
-		case Shape::BOX:
-			shape = std::unique_ptr<btBoxShape>(new btBoxShape(*static_cast<btBoxShape*>(toCopy.shape.get())));
-			break;
-		case Shape::SPHERE:
-			shape = std::unique_ptr<btSphereShape>(new btSphereShape(*static_cast<btSphereShape*>(toCopy.shape.get())));
-			break;
-		case Shape::CAPSULE:
-			shape =
-				std::unique_ptr<btCapsuleShape>(new btCapsuleShape(*static_cast<btCapsuleShape*>(toCopy.shape.get())));
-			break;
-		case Shape::CONE:
-			shape = std::unique_ptr<btConeShape>(new btConeShape(*static_cast<btConeShape*>(toCopy.shape.get())));
-			break;
-	}
-	rigidBody = std::make_unique<btRigidBody>(toCopy.rigidBody->getMass(), motionState.get(), shape.get());
+
+	rigidBody = std::make_unique<btRigidBody>(toCopy.mass, motionState.get(), toCopy.shape.get());
 
 	App->GetModule<ModulePhysics>()->AddRigidBody(this, rigidBody.get());
 	SetUpMobility();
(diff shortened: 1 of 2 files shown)


Source/DataModels/Windows/EditorWindows/WindowInspector.cpp
```diff
@@ -129,19 +129,27 @@ void WindowInspector::DrawWindowContents()
...
+	}
+	
 	if (lastSelectedGameObject)
 	{
 		bool enable = lastSelectedGameObject->IsEnabled();
 		ImGui::Checkbox("Enable", &enable);
 		ImGui::SameLine();
 
 		std::string name = lastSelectedGameObject->GetName();
-		name.resize(24);
```
gerard-queralt, line 143: Why was this changed? Doesn't look related to the issue you mention.

Source/DataModels/Components/ComponentRigidBody.cpp
```diff
@@ -44,47 +44,42 @@ ComponentRigidBody::ComponentRigidBody(bool active, GameObject* owner) :
...
 
 	transform = toCopy.transform;
-	boxSize = toCopy.boxSize;
-	radius = toCopy.radius;
-	factor = toCopy.factor;
-	height = toCopy.height;
 
-	currentShape = toCopy.currentShape;
 	motionState = std::unique_ptr<btDefaultMotionState>(new btDefaultMotionState(*toCopy.motionState.get()));
-	switch (static_cast<Shape>(currentShape))
```
gerard-queralt, line 59: Could be missing something, but we don't seem to be copying the `shape` member anywhere. Won't there be issues, if that's the case? @OriolSors

Source/DataModels/Components/ComponentRigidBody.cpp
```diff
@@ -44,47 +44,42 @@ ComponentRigidBody::ComponentRigidBody(bool active, GameObject* owner) :
...
-			shape =
-				std::unique_ptr<btCapsuleShape>(new btCapsuleShape(*static_cast<btCapsuleShape*>(toCopy.shape.get())));
-			break;
-		case Shape::CONE:
-			shape = std::unique_ptr<btConeShape>(new btConeShape(*static_cast<btConeShape*>(toCopy.shape.get())));
-			break;
-	}
-	rigidBody = std::make_unique<btRigidBody>(toCopy.rigidBody->getMass(), motionState.get(), shape.get());
+
+	rigidBody = std::make_unique<btRigidBody>(toCopy.mass, motionState.get(), toCopy.shape.get());
```
gerard-queralt, line 67: We're using the copy's `shape` to create the rigid body. Could this cause issues if the original rigid body is deleted? I tried testing and it seems fine, but since stuff like this is delicate I want to double check. @OriolSors

gerard-queralt (commented):
Aside from the questions, I noticed a weird error where disabling a rigid body and enabling it again rounds up the gravity to 10.

Source/DataModels/Windows/EditorWindows/WindowInspector.cpp
AuxPort66: Is it necessary to resize? With just c_str() it seems to work correctly. (At least the bug where you can't put more characters than you have at the beginning) the resize was for something else?

AuxPort66:
> Aside from the questions, I noticed a weird error where disabling a rigid body and enabling it again rounds up the gravity to 10.

Fixed! When you add the rigidbody back to the ModulePhysics, it resets its gravity, so I need to rewrite the rigidbody's gravity when it's enabled.

Source/DataModels/Components/ComponentRigidBody.cpp
OriolSors: We first assign the type of shape, boxSize, the radius, factor... and in line 79 we call the SetCollisionShape with these variables properly copied, so the shape is updated as @AuxPort66 explained. In fact, we will clean soon some of these variables as @gerard-queralt suggested in another PR.

OriolSors: We first assign the type of shape, boxSize, the radius, factor... and in line 79 we call the SetCollisionShape with these variables properly copied, so the shape is updated as @AuxPort66 explained. In fact, we will clean soon some of these variables as @gerard-queralt suggested in another PR.

Source/DataModels/Components/ComponentRigidBody.cpp
OriolSors: Same as my answer above

Source/DataModels/Windows/EditorWindows/WindowInspector.cpp
gerard-queralt: It's not necessary, but to my understanding it's more efficient. We avoid having to reallocate the memory when pushing characters into the string, and we avoid converting back to `const char*`. Definetly not a big deal, though.

Source/DataModels/Components/ComponentRigidBody.cpp
gerard-queralt: Thank you both for answering, missed that! I agree that it's better this way.

gerard-queralt (approved):
Thanks again for answering my question. Leaving the first thread unresolved in case you want to apply my suggestion.

Source/DataModels/Windows/EditorWindows/WindowInspector.cpp
AuxPort66: But that string it's created all the time so everytime you are reallocating that memory. I don't think it makes any difference

AuxPort66: I'm going to merge it so that the level design doesn't kill me but I want to look at it more because as I understand the relocation of this part there is no benefit? (I think) but now I'm wondering. I will keep looking at it.

PR #414, opened 2023-05-23
feat: Triggers created

Dedado3000:
Now you can create a Ghost objects that triggers the on collision enter / exit of the scripts of the 2 objects


Dedado3000:
The develop update was for having a real scene where I can test big scenarys

Source/DataModels/Components/ComponentRigidBody.cpp
```diff
@@ -88,9 +90,12 @@ ComponentRigidBody::~ComponentRigidBody()
 void ComponentRigidBody::OnCollisionEnter(ComponentRigidBody* other)
 {
 	assert(other);
-	// delegate to notify other components
```
OriolSors, line 91: We need that for the ComponentBreakable

Source/Modules/ModulePhysics.cpp
```diff
@@ -140,6 +140,7 @@ void ModulePhysics::ManageCollisions()
 					if (collisions.find(key) == collisions.end())
 					{
 						rb->OnCollisionEnter(other);
+						other->OnCollisionEnter(rb);
```
OriolSors, line 143: With the other revert we doesn't need that

OriolSors (changes requested):

OriolSors (approved):

AlexisDieu (approved):

PR #415, opened 2023-05-23
refactor: improved methods to access a GameObject's components

gerard-queralt:
The main idea is to change this recurring piece of code
```c++
static_cast<ComponentTransform*>(object->GetComponent(ComponentType::TRANSFORM))
```
to
```c++
object->GetComponent<ComponentTransform>()
```

Important files to review:
- `GameObject.h` <- renamed and refactored methods, added new ones
- `GameObject.inl` <- file where function templates were implemented, to avoid cluttering the actual header file

Things I'd like to improve:
- `GameObject::GetComponents` template should also return a view, just like the non-templated version. We could do that if we change the signature to return `auto` and do some casts or if we templarize the current `ComponentView`. I could try that if you're interested, but my main goal was improving usability rather than performance (which, to emphasize, should have been improved as well).
- `GameObject::GetAllDescendants` should return a view as well. Right now, we are iterating all the descendants, creating the list that we return, and presumably iterating that list as well. Basically, we iterate all the descendants twice. If we could use a view, we could construct the list "on the fly". Issue is, I'm not sure how we can do that, since the function is recursive, so it's not as simple as applying it to a given vector; we'd need to apply it recursively, somehow, and I haven't figured out a way to do that. Any ideas are welcome :D


Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -731,13 +687,13 @@ bool GameObject::IsADescendant(const GameObject* descendant)
 	return false;
 }
 
-std::list<GameObject*> GameObject::GetGameObjectsInside()
+std::list<GameObject*> GameObject::GetAllDesdendants()
```
markmeliaa, line 690:
```suggestion
std::list<GameObject*> GameObject::GetAllDescendants()
```

markmeliaa:
Nice improvement Gerard, thanks! 

Regarding the `GameObject::GetComponents` if you'd like to perform the change, I have no issue with it, but its not necessary, so do as you please.
Regarding the `GameObject::GetAllDescendants`, I'll think about any possible solutions and I'll let you know if somethings comes to mind.

gerard-queralt:
> Regarding the `GameObject::GetComponents` if you'd like to perform the change, I have no issue with it, but its not necessary, so do as you please.

I've mostly done it, I think. I pushed it into a different branch to try and get this MR merged, you can check the [commit](https://github.com/Horizons-Games/Axolotl-Engine/commit/a5970d4c958d6e7f3838320f1a068bad7159824a). I could create a draft MR too, if you want.
The main work I've done is trying to organize better our aliases for `std::ranges::views` and facilitate their usage. This should, ideally, make it easier to use views for other stuff. But I'm saying I've _mostly_ done it because I want to figure out why `size` and the array subscript operator `[]` don't work. According to the [documentation](https://en.cppreference.com/w/cpp/ranges/view_interface), the view needs to satisfy some constraints in order to have access to those operations, so I'll try to figure it out. In the meantime, seems everything works fine with `foreach` loops.

markmeliaa (approved):

markmeliaa (commented):
The build failed in ReleaseGame @gerard-queralt

UriMurcia (approved):

PR #419, opened 2023-05-25
Refactor [Game play]: new mobility script using bullet

DADSILVER:
- Old mobility scripts are deleted.
- The mobility scrips are divided into 3: Move, rotation, and jump.
- The mobility script also contains the dash. Because you need the direction to know where to dash to.
- A new camera script has been added. With it, you can move the camera vertically.
- A new scene is created to test gameplay.
- The other scenes are updated with the news scripts.


Source/Modules/ModuleInput.cpp
```diff
@@ -129,16 +129,16 @@ update_status ModuleInput::Update()
 				{
 					inFocus = true;
 				}
-
 				break;
-
 			case SDL_KEYDOWN:
-				keysState[sdlEvent.key.keysym.scancode] = KeyState::DOWN;
+				if (keysState[sdlEvent.key.keysym.scancode] != KeyState::REPEAT)
```
gerard-queralt, line 134: I'm not sure this is going to work, even if it should in theory. If I recall correctly, when a key is held, SDL emits alternate SDL_KEYDOWN/UP signals. What we should do to detect such cases is use the `repeat` flag in the SDL event. All of this to say, take a look at #408, it should fix this.

Game/Scripts/PlayerJumpScript.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include "ModuleInput.h"
+
+#include "Components/ComponentTransform.h"
+#include "Components/ComponentRigidBody.h"
+
+REGISTERCLASS(PlayerJumpScript);
+
+PlayerJumpScript::PlayerJumpScript() : Script(), jumpParameter(2000.0f), canDoubleJump(false) , jumpReset(0)
+{
+	REGISTER_FIELD_WITH_ACCESSORS(JumpParameter, float);
```
gerard-queralt, line 13: A suggestion from the scripting side, since you don't seem to be using the accessor methods for anything, you can use
```suggestion
	REGISTER_FIELD(jumpParameter, float);
```
And get rid of them entirely. This macro is intended to allow the game team to add additional logic when getting and setting a given field. For example, if one of the members of the class is computed from another, similar to what you're doing in `Start`.

Game/Scripts/PLAYERROTATIONSCRIPT.CPP
markmeliaa, line 1: Could you change the name of this script to PlayerRotationScript.cpp, please?

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+	// Cooldown Dash
+	if (nextDash != 0 && nextDash < SDL_GetTicks())
+	{
+		canDash = true;
+		nextDash = 0;
+	}
+}
+
+float PlayerMoveScript::GetSpeed() const
```
markmeliaa, line 177: The good thing about what Gerard said is that, when using `REGISTER_FIELD` instead of `REGISTER_FIELD_WITH_ACCESSORS` is that all these getters and setters will not be needed because the MACRO will manage them for us, so only use the `REGISTER_FIELD_WITH_ACCESSORS` only if you have to give some extra logic to any getter or setter.

markmeliaa (changes requested):
Good job to both of you! Left just one small change regarding the name of a script. Also, one question out of curiosity, why did you decide to call the `Move()`, `Rotate()`... functions from the `PreUpdate()` instead of from the `Update()`? I'm not seeing it as a bad thing, just to know why the decision of doing so.

Game/Scripts/PlayerJumpScript.cpp
DADSILVER: Fixed :)

Game/Scripts/PlayerMoveScript.cpp
DADSILVER: Fixed

Source/Modules/ModuleInput.cpp
DADSILVER: Okay, I undo the changes.

markmeliaa (dismissed):

Dedado3000 (dismissed):

markmeliaa (approved):

PaablooCH (approved):

PR #421, opened 2023-05-26
feat : add Breakable with bullet

AlexisDieu:
- Add the possibility of making a rigidBody Breakable
- Be able to change the impulsion from the center to adapt how far you want the fragment to be eject
- Apply random impulsion to make always different explosion
- All the child of the Breakable rigidBody are being rigidBody after it break


Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -19,6 +19,13 @@
 #include "../Components/UI/ComponentCanvas.h"
 #include "../Components/UI/ComponentImage.h"
 #include "../Components/UI/ComponentTransform2D.h"
+#include "../Components/ComponentRigidBody.h"
+#include "../Components/ComponentBreakable.h"
+#include "../Components/ComponentMockState.h"
+#include "../Components/ComponentAudioSource.h"
+#include "../Components/ComponentAudioListener.h"
+#include "../Components/ComponentMeshCollider.h"
+#include "../Components/ComponentScript.h"
```
totisama, line 28: I see that some of this includes are twice

OriolSors (approved):

PR #427, opened 2023-05-28
feat: allowing to change cubemap intensity from inspector

xavi1111:


Source/Shaders/Common/Functions/pbr_functions.glsl
```diff
@@ -11,7 +11,7 @@ samplerCube diffuse_IBL, samplerCube prefiltered_IBL, sampler2D environmentBRDF,
     vec3 radiance = textureLod(prefiltered_IBL, R, roughness * numLevels_IBL).rgb;
     vec2 fab = texture(environmentBRDF, vec2(NdotV, roughness)).rg;
     vec3 diffuse = (diffuseColor * (1 - specularColor));
-    return diffuse * irradiance + radiance * (specularColor * fab.x + fab.y);
+    return (diffuse * irradiance + radiance * (specularColor * fab.x + fab.y));
```
PaablooCH, line 14: why the ()?

Source/Shaders/default_fragment.glsl
```diff
@@ -224,7 +225,7 @@ void main()
 
     vec3 R = reflect(-viewDir, norm);
     float NdotV = max(dot(norm, viewDir), EPSILON);
-    vec3 ambient = GetAmbientLight(norm, R, NdotV, roughness, Cd, f0, diffuse_IBL, prefiltered_IBL, environmentBRDF, numLevels_IBL);
+    vec3 ambient = GetAmbientLight(norm, R, NdotV, roughness, Cd, f0, diffuse_IBL, prefiltered_IBL, environmentBRDF, numLevels_IBL) * cubeMap_intensity;
```
PaablooCH, line 228: can you pass the variable inside the fuction to have all IBL inside one place?

Source/ComponentCubeMap.h
PaablooCH, line 1: is not cubeMap is cubemap, is a only word

Source/ComponentCubeMap.h
```diff
@@ -0,0 +1,27 @@
...
+	void LoadOptions(Json& meta) override;
+
+	bool IsStatic();
+	void SetStatic(bool newStatic);
+	
+	void SetIntensity(float intensity);
+	float GetIntensity();
+
+private:
+	float intensity = 1.0;
```
PaablooCH, line 21: set the variable in the constructor

Source/DataModels/Cubemap/Cubemap.h
```diff
@@ -41,6 +43,8 @@ class Cubemap
 	int numMipMaps;
 
 	std::shared_ptr<ResourceCubemap> cubemapRes;
+
+	float intensity = 1.0f;
```
PaablooCH, line 47: why you create this variable here and in componentCubemap? its better to have it in one place only

PaablooCH: Probably the one that we should preserve is the one in component

Source/WindowComponentCubeMap.cpp
```diff
@@ -0,0 +1,42 @@
...
+{
+	DrawEnableAndDeleteComponent();
+
+	ComponentCubeMap* ascubeMap = static_cast<ComponentCubeMap*>(component);
+
+	if (ascubeMap)
+	{
+		ImGui::Text("");
+		float intensity = ascubeMap->GetIntensity();
+		float max_intensity = 1.0;
```
PaablooCH, line 22: we don't set the max intensity that lower, instead we have already a max_instensity in componentLight at 5000.f use that

PaablooCH (changes requested):
Hi xavi, I see a couple of things

Source/Shaders/default_fragment.glsl
xavi1111: Miqui told me to leave it here I don't mind passing it as a parameter. I 'll go do the other changes.

Source/DataModels/Cubemap/Cubemap.h
xavi1111: I think it's true its unnecessary to have the variable two times but i think the one that should stay is the one in cubemap because it should be accessible for all gameObjects and the component is bound to the cubemap game object.

PaablooCH: if we store the information inside the cubemap, each scene that has it will share the same intensity

xavi1111: Well the component is like a middle man that handles saving and loading the correct value to the cubemap

Source/DataModels/Cubemap/Cubemap.cpp
```diff
@@ -45,6 +46,7 @@ Cubemap::~Cubemap()
 Cubemap::Cubemap(std::shared_ptr<ResourceCubemap> cubemapRes) : cubemapRes(cubemapRes)
```
PaablooCH, line 46:
```suggestion
Cubemap::Cubemap(std::shared_ptr<ResourceCubemap> cubemapRes) : cubemapRes(cubemapRes), intensity(1.f)
```
its more clean like this, but its the same so np

Source/ComponentCubeMap.cpp
PaablooCH, line 1: All the files that you created are in the main folder, move it to its corresponding folders. Don't forget to rename form cubeMap to cubemap

xavi1111: Done, thanks for the feedback.

Source/DataModels/Cubemap/Cubemap.h
MacXxs: Hi Xavi, we should store the value for the instensity in the component better than the cubemap perse. Sorry i didn't tell you before during the call i didn't notice it,
At the end of the day the intensity is a value we pass directly to the fragment shader, and the cubemap doesn't need it directly, just the component so each iteration of a componentCubemap has it's own value for intensity, even though for the moment we just have one cubemap for the whole scene, but later when adding local cubemaps (if we have the time) each component should handle the intensity.

PaablooCH (approved):
works fine well done👍

AuxPort66 (approved):

PR #432, opened 2023-05-29
fix/refactor: delay adding updatables to scene, deleting objects. GameObjects removed from scripts that reference them

gerard-queralt:
As mentioned in the code, the method to remove the GameObject that is being deleted from the scripts turned out relatively complex. Tried my best to explain what it's doing with comments.


markmeliaa (approved):
Looks like a really complex solution but for now it will suffice.

AuxPort66 (approved):

PR #452, opened 2023-05-31
feat: Add Cantina NPCs

AuxPort66:
- Cantina NPCs
- Added a Script to Randomize the activation of the animation so not all the NPCs move at the same time


Vlukkee (approved):
Very Good! Very Guapo!

diegusmejia117 (approved):
Not very good, not very guapo

PR #461, opened 2023-06-01
feat: Build Game window

gerard-queralt:
- Allow the user to select a starting scene for the game
- Allow the user to select the build type
- Allow the user to choose whether to re-generate the whole `Assets.zip`

In order to allow the starting scene to be set dynamically by the user, a `Startup.conf` file will be created in the `Game` directory. For now it only holds the starting scene name, but more things can be added if they are needed. Feel free to discuss alternatives, I'm not particularly fond of it.


Source/Modules/ModuleScene.cpp
```diff
@@ -53,18 +51,23 @@ bool ModuleScene::Init()
...
+		char* buffer;
+		unsigned int fileSize = App->GetModule<ModuleFileSystem>()->Load(GAME_STARTING_CONFIG, buffer);
+		rapidjson::Document doc;
+		Json startConfig(doc, doc);
+		startConfig.fromBuffer(buffer);
+		delete buffer;
+		
+		std::string startingScene = startConfig["StartingScene"];
+		std::string scenePath = LIB_PATH "Scenes/" + startingScene;
+		LoadScene(scenePath, false);
```
MarcGonzalezMoratona, line 68: what happens if the user has no sartingScene defined in the configuration file? is there a default value?

Source/FileSystem/ModuleFileSystem.cpp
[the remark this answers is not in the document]
gerard-queralt: That's right. However, I'm not sure how to read this new file otherwise. Maybe we can try adding it to the zip file, but I don't remember if Physfs allows us to write to it after it's been created. As I said, I'm open to suggestions on alternative approaches.

Source/Modules/ModuleScene.cpp
gerard-queralt: A crash, same as with most things. I guess if we can't find that scene we could default to loading the first scene in the folder, but I'm not sure if we should allow an invalid configuration.

Source/DataModels/Windows/WindowMainMenu.cpp
```diff
@@ -210,16 +212,15 @@ void WindowMainMenu::ShortcutSave()
...
+	ImGui::MenuItem("Build", nullptr, &showBuild);
+	if (showBuild)
 	{
-		if (ImGui::MenuItem("Debug"))
+		build->Draw(showBuild);
+		// if the window was closed, either by actually closing it or by starting the build process, reset its values
+		if (!showBuild || build->StartedBuild())
 		{
-			builder::BuildGame(builder::BuildType::DEBUG_GAME);
+			showBuild = false;
```
MarcGonzalezMoratona, line 222: why do we set the value to false if it was already false?

Source/Modules/ModuleScene.cpp
MarcGonzalezMoratona: maybe we can handle this with a warning or a popup, not loading a default scene but telling the user the config is invalid

Source/DataModels/Windows/PopUpWindows/WindowBuild.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include "FileSystem/ModuleFileSystem.h"
+#include "FileSystem/Json.h"
+
+#include "Auxiliar/GameBuilder.h"
+
+#include <map>
+
+namespace
+{
+const std::string noneSelected = "-";
```
MarcGonzalezMoratona, line 13: why aren't we adding the std namespace here?

MarcGonzalezMoratona: are we defining a custom namespace?

Source/DataModels/Windows/PopUpWindows/WindowBuild.cpp
```diff
@@ -0,0 +1,128 @@
...
+			if (isSelected)
+			{
+				ImGui::SetItemDefaultFocus();
+			}
+		}
+		ImGui::EndCombo();
+	}
+}
+
+void WindowBuild::DrawGenerateZipCheckmark()
```
MarcGonzalezMoratona, line 90:
```suggestion
void WindowBuild::DrawGenerateZipCheckbox()
```

Source/FileSystem/ModuleFileSystem.cpp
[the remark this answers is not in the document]
ericdacal: Maybe a good idea, can be make a folder outside zip for configs, and only allow to Physfs to access this directory.

Source/DataModels/Windows/WindowMainMenu.cpp
gerard-queralt: It's not necessarily `false`, notice that the condition is an `or`. If `build->StartedBuild()` is `true` (meaning the user has clicked the button to start the build) we also want the window to not be drawn anymore.

Source/DataModels/Windows/PopUpWindows/WindowBuild.cpp
gerard-queralt: We are defining an unnamed namespace. Basically, we are defining "static" variables in the source file. In quotes because they are different from actual `static` variables, since those need to be declared in the class body. But they are also shared between all instances of the class, like `static` members. It's a topic with a bit of literature behind, but [this StackOverflow post](https://stackoverflow.com/questions/357404/why-are-unnamed-namespaces-used-and-what-are-their-benefits) is a good start.

Source/FileSystem/ModuleFileSystem.cpp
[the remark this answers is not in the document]
gerard-queralt: OK, turns out you can write to an existing zip using the zip library. Did just that and it seems to work.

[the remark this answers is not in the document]
ericdacal: Nice, that's perfect then.

ericdacal (dismissed):

AuxPort66 (approved):

AuxPort66 (approved):

ericdacal (approved):

PR #464, opened 2023-06-02
fix: Fix memory leaks I

AuxPort66:
First batch fixing memory leaks. I'm working on more, this batch was put in the release so Carlos can see that we're working on it, but you can tell me any change requests anyway


AuxPort66:
This was merged in #467

PR #465, opened 2023-06-03
Fix: Fix copy

AuxPort66:
- Fix Tag Copy
- Fix Animation Copy

### DIFF
--- a/Source/DataModels/GameObject/GameObject.cpp
+++ b/Source/DataModels/GameObject/GameObject.cpp
@@ -365,6 +365,12 @@ void GameObject::CopyComponent(Component* component)
 			break;
 		}
 
+		case ComponentType::ANIMATION:
+		{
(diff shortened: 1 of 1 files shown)


Dedado3000 (dismissed):
Thanks god someonr that take care of level

Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -65,6 +65,8 @@ GameObject::GameObject(const GameObject& gameObject) :
 			   StateOfSelection::NO_SELECTED,
 			   gameObject.staticObject)
 {
+	tag = gameObject.tag;
```
gerard-queralt, line 68: Better to add this to the initializer list.

gerard-queralt (dismissed):

Dedado3000 (approved):

gerard-queralt (approved):

PR #474, opened 2023-06-06
feat[WIP]: AreaLights

PaablooCH:
First pr for areaLights. While i'm trying to solve a visual bug, level design needs to start working with areaLights.

This PR **adds**:

- The componentAreaLight with visual debug like the shape of the light and also the attenuation distance. Look that tube attenuation is a fusion with 2 spheres and a tube.
- Added an option below Spot and Point light to create an Area Light.
- Able to save a scene and load it with the lights 
- Added also a 3D sphere object.

**While this is just a request from level I won't touch anything unless is mandatory (crash or something similar), any change you request, please contact me and i will add it in the final version**


MacXxs (approved):

Dedado3000 (approved):
Functionality checked:
Control C+ V
Control D
Copy Parameter
Serialization

MacXxs (approved):

PR #476, opened 2023-06-07
fix: can load scenes multiple times in Game mode

gerard-queralt:
From the looks of it, the error was caused because the resources map was not properly cleaned in Game mode, unlike in Engine. Did a quick solution, there might be better ways.

Also cleaned up a bit the scene de-serialization process at the beginning, reducing the number of maps.


gerard-queralt:
Seems like UI stuff doesn't work well unless the map we use for de-serialization is an `std::unordered_map`. Any ideas on why this might happen @AuxPort66 ? I'm don't think there's any reason (besides that one) for the map to be unordered. For the time being, I've reverted it so we can merge this.

AuxPort66:
> Seems like UI stuff doesn't work well unless the map we use for de-serialization is an `std::unordered_map`. Any ideas on why this might happen @AuxPort66 ? I'm don't think there's any reason (besides that one) for the map to be unordered. For the time being, I've reverted it so we can merge this.

The UI rendering is in hierarchical order. When we load the json it has the insertion order because we don't save the hierarchy position in other ways, so with a map you lost that order. Is the scene loading in another order and then some images are in front of others when they shouldn't? Is that what you mean by "it doesn't work right"?

gerard-queralt:
I imagine the issue was that the images were rendered out of order, but since I couldn't see them I didn't want to make assumptions. What I don't understand is, if we rely on the order of the hierarchy, how come an *unordered* map works, but a regular, *ordered* map doesn't?

AuxPort66 (dismissed):

AuxPort66:
The rest of the level seams loaded correctly, but the NPCs in the bar don't have... heads. And the scene is updated, so we have the same scene as in develop.

This is in develop, I open the engine, load the scene from Level 1 and press play
image

And this is in the fix-load branch, the same process
image

It disappears when you hit play. In game it also happens. I clean lib to be sure, and it's the same. Weird.

gerard-queralt:
Should be fixed. Seems the cause was the change I did to the loop that performs the calls to `LinkChild`, I guess the transforms were wrongly computed if linking was done in a different order, but I couldn't really tell you the cause. That also "fixed" the question I had with the UI elements, so I re-reverted the map to be ordered so it's cheaper to iterate.

gerard-queralt:
Got curious about this issue and found this really interesting article that explains what I'm pretty sure is the same problem: https://www.cppstories.com/2017/12/weakptr-memory/
TL;DR: when creating an object with `make_shared`, the memory of the object itself and the control block of the pointer are allocated in the same chunk, and since the control block cannot be de-allocated, neither can the memory reserved for the object. Basically, even though the object was being destroyed and the memory was cleaned, we were not freeing it, so it couldn't be used for other allocations. Which means the deleter is probably not needed to solve the problem, we just need to do two allocations (one with `new` and one for the pointer, like we are doing), but better to also clear the map in my opinion.

AuxPort66 (approved):

diegusmejia117 (approved):

PR #483, opened 2023-06-10
fix: multiple transform changed notifications on load scene

OriolSors:


Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -225,7 +225,7 @@ void GameObject::LinkChild(GameObject* child)
 		ComponentTransform* transform = child->GetComponent<ComponentTransform>();
 		if (transform)
 		{
-			transform->UpdateTransformMatrices();
+			transform->UpdateTransformMatrices(false);
```
gerard-queralt, line 228: Are we sure this won't have unintended side effects? Maybe we can add this boolean as a parameter to the method. It's also possible this is perfectly fine, in which case feel free to ignore me.

OriolSors: I am calling the root GameObject after the whole scene has loaded to call correctly and only once to the OnTransformChange observers.

Dedado3000 (approved):
working as intended

Source/DataModels/GameObject/GameObject.cpp
gerard-queralt: Yes, I don't doubt that the scene loading works. I'm just not sure if the other calls to `LinkChild` will work as they did before. For instance, changing a game object's parent in the hierarchy will no longer notify the update to the components, if I'm not missing anything. Is that the behaviour we want?

AuxPort66: When you change the parent, your global doesn't change, only your local, so you don't need to update the other components or children. So I think that on that side it has no side effects.

AuxPort66 (approved):

PR #487, opened 2023-06-12
Feat: Particle System FileSystem

AuxPort66:
- Module Get/Sets and other member changes necesarry to the correct edition of the ParticleSystem
- Updatable Function to mantain the engine functional meanwhile we edit the ParticleSystem
- Merge Fix solved
- Importing features
- Component serialization
- Add load to the window component so be can use the same ParticleSystem
- Some changes on ResourceParticleSystem to work with our FileSystem


MacXxs:
There's a bug when creating a ParticleSystem and adding an emitter, and then removing it. For example if you have a particle system with two emitters, remove one of them and then save, the removed one appears again.

eddb98d297cccef0ddc41e2a5637590f

AuxPort66:
> There's a bug when creating a ParticleSystem, adding an emitter, and then removing it. For example if you have a particle system with two emitters, remove one of them and then save, the removed one appears again.
> 
> eddb98d297cccef0ddc41e2a5637590f

It seems that the window Delete button is editing the component EmitterInstances instead of the resource, that will be the correct logic of the component and resource using in our engine. I see the same problem in the Add Emitter, but I didn't check the Delete Emitter logic. Solved.

MacXxs:
> > There's a bug when creating a ParticleSystem, adding an emitter, and then removing it. For example if you have a particle system with two emitters, remove one of them and then save, the removed one appears again.
> > eddb98d297cccef0ddc41e2a5637590f
> >       
> >         [
> >           
> >         
> >             eddb98d297cccef0ddc41e2a5637590f
> >           ](https://user-images.githubusercontent.com/18547655/248500710-fefa3e42-d5c3-4c0a-8d50-8b9d7207f020.gif)
> >         
> >         
> >           
> >             
> >               
> >             
> >             
> >               
> >               
> >             
> >           
> >           [
> >             
> >               
> >             

Shouldn't the resource be modified only on save? Like this if you delete by mistake an emitter but don't save, you don't lose your changes, and you simply need to reload the particle system to restore the lost progress.
> >           ](https://user-images.githubusercontent.com/18547655/248500710-fefa3e42-d5c3-4c0a-8d50-8b9d7207f020.gif)
> 
> It looks like the Delete button on the window is editing the EmitterInstances component instead of editing the resource directly, that would be the correct logic for the component and resource logic being used in our engine. I saw the same problem in Add Emision on the original branch, but I didn't check the logic of Delete Emission. Solved.

Shouldn't the resource be modified only on save? Like this if you delete by mistake an emitter but don't save, you don't lose your changes, and you simply need to reload the particle system to restore the lost progress.

MacXxs:
When loading a saved scene with a particle system that uses a texture,  the texture isn't loaded until the particle system is selected

abf0d600b11ff3fe1f27ac0e1358d611

AuxPort66:
> When loading a saved scene with a particle system that uses a texture, the texture isn't loaded until the particle system is selected
> 
> abf0d600b11ff3fe1f27ac0e1358d611

That sounds like a Render problem not a Resource problem, but yeah I can check it

MacXxs (approved):

ericdacal (approved):

PR #490, opened 2023-06-12
style: component windows

MarcGonzalezMoratona:


AuxPort66 (approved):

PaablooCH:
2 things:

- I like the style that we have now, cap letters for Components title and some camel case for the input variable, in my opinion it looks more pretty.
- why are you adding commits from navMesh that are in other pr that points to the physics main branch?

MarcGonzalezMoratona:
> 2 things:
> 
> * I like the style that we have now, cap letters for Components title and some camel case for the input variable, in my opinion it looks more pretty.
> * why are you adding commits from navMesh that are in other pr that points to the physics main branch?

I was trying to unify the styling, so we can make it all uppercase if you prefer.

This commits where added because I branched from the navmesh branch to avoid making the same thing twice

AuxPort66:
> 2 things:
> 
> * I like the style that we have now, cap letters for Components title and some camel case for the input variable, in my opinion it looks more pretty.
> * why are you adding commits from navMesh that are in other pr that points to the physics main branch?

PR #491, opened 2023-06-12
feat[Area Lights]: visual bug fixed && finished the component

PaablooCH:
What is new:
- The visual bug regarding the attenuation is solved.
- If you disable the area component the light stop rendering
- Added area light to debug mode

Other things:
- Added an assert when a shader doesn't compile
- Fix a bug where a spot doesn't stop rendering when you disable the component
- Use class variable instead of getters


AuxPort66 (approved):

markmeliaa (commented):
If I'm not mistaken, everything regarding the changes of active/enable seems fine, but idk anything about area lights so I won't approve this.

MacXxs (approved):
perfect 👍🏼

PR #497, opened 2023-06-14
feat[WIP]: first approach to powerUp

PaablooCH:
What is done:
- When a enemy dies, it drops a powerUp.
- Now we only have 4 types of powerUp: Heal, Defense, Attack, Speed.
- Once you acquired one, you can activate it with "Z" or you can drop it with "X".
-  There is just one powerUp that is not permanent (HEAL), the others have a duration of 10 sec.
- A powerUp will remain 10 sec after it spawns.
- PowerUp are not created nor destroyed, they are moved to a safe location until we need them.
- New Managers scripts. This scripts will contain all the info necessary info for a entity, fe health, attack, speed or the powerUp you have.


Game/Scripts/BixAttackScript.h
```diff
@@ -46,4 +48,10 @@ class BixAttackScript : public Script
 	ComponentTransform* ray3Transform;
 	ComponentTransform* ray4Transform;
 	//--Provisional
-};
\ No newline at end of file
+};
+
+inline void BixAttackScript::IncreaseAttack(float ammountAttack)
+{
+	damageAttack += ammountAttack;
```
markmeliaa, line 55:
```suggestion
	damageAttack += amountAttack;
```

Game/Scripts/BixAttackScript.h
```diff
@@ -16,6 +16,8 @@ class BixAttackScript : public Script
 	BixAttackScript();
 	~BixAttackScript() override = default;
 
+	void IncreaseAttack(float ammountAttack);
```
markmeliaa, line 19:
```suggestion
	void IncreaseAttack(float amountAttack);
```

Game/Scripts/BixAttackScript.h
```diff
@@ -46,4 +48,10 @@ class BixAttackScript : public Script
 	ComponentTransform* ray3Transform;
 	ComponentTransform* ray4Transform;
 	//--Provisional
-};
\ No newline at end of file
+};
+
+inline void BixAttackScript::IncreaseAttack(float ammountAttack)
```
markmeliaa, line 53:
```suggestion
inline void BixAttackScript::IncreaseAttack(float amountAttack)
```

markmeliaa (approved):
A couple refactors will be performed after this is merged so this is not definitive, but thank you for your work, looks and works fine for the time being.

markmeliaa:
@PaablooCH the branch does not compile now

PR #498, opened 2023-06-15
Chore[Level/Art]: Cantina´s advanced enviromental approach 

diegusmejia117:
This PR adds new props and art to the cantina section of the scene.

Added:

Floors, 
Ceiling pipelines, 
subfloor pipelines, 
new dance floor, 
turbines, 
VIP ceilings,
bootles,
More NPCs
and their respective colliders(dance floor colliders have a werid bug)


Source/DataModels/Batch/GeometryBatch.cpp
```diff
@@ -913,7 +913,7 @@ void GeometryBatch::WaitBuffer()
 			GLenum waitReturn = glClientWaitSync(gSync[frame], GL_SYNC_FLUSH_COMMANDS_BIT, 1000);
 			if (waitReturn == GL_ALREADY_SIGNALED || waitReturn == GL_CONDITION_SATISFIED)
 			{
-				return;
+				return;
```
Dedado3000, line 916: please, undo this, was a missclick.

Dedado3000 (changes requested):
Can you please delete unused scenes from level1

Source/DataModels/Batch/GeometryBatch.cpp
diegusmejia117: just checked the code it is written correctly no missclick(maybe bug)?

gerard-queralt: This difference indicates that you added an extra space after the return, which you probably did by mistake. I imagine you didn't notice any difference because it's a space :P

diegusmejia117: Done

Dedado3000 (approved):

ericdacal (approved):

PR #499, opened 2023-06-16
feat: allow in-engine pause

gerard-queralt:
- After starting game in the engine, you can pause it, disabling physics, animations, restoring engine camera...
- Press pause again to resume
- Better UI visual feedback

Not sure about the expected behaviour of animations in particular during pause. Took the easiest approach which is to disable them entirely, but feel free to discuss alternatives.


UriMurcia:
Nice work Gerard! I think it's everything okay, just two things:

- Particles don't stop when game is paused, it's not a big problem but just to mention it.
- And also audio does not stop. This one I don't see it a big problem neither, and I'm not sure if it is possible to pause. I've found online that it can be paused by using "PauseAll" and "ResumeAll", we can try it together if you want.

gerard-queralt:
Thanks Uri! Truth be told I hadn't tried it since the last few merges so I didn't know about the particles, thanks for pointing it out :P
And thank you also for the advice on audio, I had no idea how to pause it. I'd be really thankful if we could take a look together, thanks for the offer!

markmeliaa (commented):
Nice approach Gerard, I just have a couple suggestions/improvements regarding the camera and the animations.

Regarding the camera, resetting to the engine camera when pause is pressed feels wrong IMO, idk if there is any important reason for why did you decide to switch back to engine camera, maybe I'm missing something. What I'd prefer is, either we don't switch to the engine camera when pause is pressed and we keep the view of the main camera, or we switch to the engine camera but we add the possiblity to control said camera so the user can go through the level with WASD (Idk how difficult this could be tho).

Regarding animations, by stopping them entirely do you mean resetting the state machine so when the pause is pressed again the state machine will start again with its `Entry` node or just stopping them and when the pause is pressed again the state machine will continue reproducing the animation of the node it was in before? I feel like the latter will be the expected behaviour for the animations.

gerard-queralt:
> Regarding the camera, resetting to the engine camera when pause is pressed feels wrong IMO, idk if there is any important reason for why did you decide to switch back to engine camera, maybe I'm missing something. What I'd prefer is, either we don't switch to the engine camera when pause is pressed and we keep the view of the main camera, or we switch to the engine camera but we add the possiblity to control said camera so the user can go through the level with WASD (Idk how difficult this could be tho).

As with most things, I copied it from how it's done in Unity :P. I think it's a good thing so the user can navigate freely thought the scene. And actually, I could've sworn you could move with WASD, maybe it broke with some merge. I'll take a look. Let me know if I you see why I think the engine camera is better, and feel free to disagree with me. I'm sure we'll achieve something better through discussion 🙂 

> Regarding animations, by stopping them entirely do you mean resetting the state machine so when the pause is pressed again the state machine will start again with its `Entry` node or just stopping them and when the pause is pressed again the state machine will continue reproducing the animation of the node it was in before? I feel like the latter will be the expected behaviour for the animations.

If I'm not mistaken (and if it didn't break in a merge), it's indeed the latter. We just stop updating and continue from where we left. But talking with Diego he mentioned that he thinks ideally the animations should keep playing, just not update the state. Wanted to gather some feedback regarding that approach, since it'd be harder to implement, so I wanted to make sure it's worth it. If it is, it could be done later down the road, no need to hold this PR hostage any longer :P

markmeliaa:
> As with most things, I copied it from how it's done in Unity :P. I think it's a good thing so the user can navigate freely thought the scene. And actually, I could've sworn you could move with WASD, maybe it broke with some merge. I'll take a look. Let me know if I you see why I think the engine camera is better, and feel free to disagree with me. I'm sure we'll achieve something better through discussion 🙂

I don't see changing to the engine camera wrong when we press pause, but only if we could move said camera with WASD, because if not, the user will gain no advange of pressing pause because they could not see the state of the game when pausing. I remember we used to have that kind of movement, but it seems that now we don't, so it should've been overriden by some merge as you said, so, as long as we bring back that movement, I'm okay with changing the camera to engine camera when we press pause.

> If I'm not mistaken (and if it didn't break in a merge), it's indeed the latter. We just stop updating and continue from where we left. But talking with Diego he mentioned that he thinks ideally the animations should keep playing, just not update the state. Wanted to gather some feedback regarding that approach, since it'd be harder to implement, so I wanted to make sure it's worth it. If it is, it could be done later down the road, no need to hold this PR hostage any longer :P

I tried seeing what happens to the animations but the camera didn't let me see if they were being executed or not in the level 1.1, but if you say that the latter happens, I trust you completely. I don't think the animations should keep playing because that would be kind of missleading and not user friendly (when the user presses pause, the expected behaviour should be that everything stops, and if the animations keep playing (eventhough their state does not change) it would be confusing for the said user IMO). So I think that, as the animations behave now, is how they should behave when pressing pause.

UriMurcia (approved):

Source/DataModels/Camera/CameraEngine.cpp
```diff
@@ -144,7 +144,7 @@ bool CameraEngine::Update()
 		}
 	}
 
-	if ((sceneFocused || rotating) && !App->IsOnPlayMode())
+	if ((sceneFocused || rotating) && App->GetPlayState() != Application::PlayState::RUNNING)
```
MarcGonzalezMoratona, line 147: Why do we use sometimes PlayState != Running and in other places we use PlayState == Stopped?

Before all those code was just !App->IsOnPlayMode().

I understand that now we have a bit more flexibility to make actions in states like running, pause or stop. But I would keep it as simple as possible when we can do it.

With keeping it simple I mean avoiding things like PlayState != Running, and using directly the PlayState == (whatever state we need)

Of course if an action has to be done in more than one state, I approve doing things this way, but as I see different approaches for a code that was not different before, I commented it

Source/DataModels/Batch/GeometryBatch.cpp
```diff
@@ -712,7 +712,7 @@ void GeometryBatch::BindBatch(bool selected)
 #ifdef ENGINE
 			bool draw = false;
 
-			if (selectedGo != nullptr && !App->IsOnPlayMode() && !isRoot)
+			if (selectedGo != nullptr && App->GetPlayState() == Application::PlayState::STOPPED && !isRoot)
```
MarcGonzalezMoratona, line 715: I see that the approach is changing IsOnPlayMode for PlayState == Running and !IsOnPlayMode for PlayState != Running, why is it different here?

Source/DataModels/Windows/EditorWindows/WindowEditorControl.cpp
```diff
@@ -28,15 +29,63 @@ void WindowEditorControl::DrawWindowContents()
...
+
+	if (playState == Application::PlayState::PAUSED)
+	{
+		ImGui::PopStyleColor();
+	}
+
+	ProcessInput();
+}
+
+void WindowEditorControl::ProcessInput()
```
MarcGonzalezMoratona, line 71:
```suggestion
void WindowEditorControl::HandleInput()
```

I think this is a more conventional name

Source/DataModels/Windows/EditorWindows/WindowEditorControl.cpp
```diff
@@ -28,15 +29,63 @@ void WindowEditorControl::DrawWindowContents()
...
 	ImGui::SameLine();
 
+	if (playState != Application::PlayState::STOPPED)
+	{
+		ImGui::PopStyleColor();
+	}
+
+	if (playState == Application::PlayState::PAUSED)
+	{
+		ImVec4 darkGray = ImVec4(.6f, .6f, .6f, 1.f);
```
MarcGonzalezMoratona, line 53:
```suggestion
		ImVec4 darkGrey = ImVec4(.6f, .6f, .6f, 1.f);
```

gray is American English, use the one you prefer

MarcGonzalezMoratona (approved):

MarcGonzalezMoratona:
Hey! nice job!

Friendly reminder: take a look at the conflicts 😉

markmeliaa (approved):

Piterga12 (approved):
Good job!

PR #502, opened 2023-06-17
Graphics/texture compression

ericdacal:


Source/DataModels/Resources/ResourceTexture.h
```diff
@@ -59,8 +60,9 @@ struct ImportOptionsTexture
 {
 	bool flipVertical;
 	bool flipHorizontal;
+	int compression;
```
gerard-queralt, line 63: Could we change this to use an enum class? I feel it be easier to understand what each value represents.

Source/DataModels/Resources/ResourceSkyBox.cpp
```diff
@@ -33,17 +33,78 @@ void ResourceSkyBox::InternalLoad()
 
 		if (textI)
```
gerard-queralt, line 34: Just noticed that the entire contents of this `if` block are copy-pasted from `ResourceTexture`. Should be pretty straig-forward to encapsulte this in a class method of `ResourceTexture`.

Source/DataModels/Resources/ResourceTexture.h
ericdacal: Solve, now we use an Enum with name TextureCompression

Source/DataModels/Resources/ResourceSkyBox.cpp
ericdacal: I create a new function to compute texture compression format that it is in ResourceTexture.

Source/DataModels/Resources/ResourceTexture.h
```diff
@@ -1,13 +1,17 @@
 #pragma once
 
 #include "Resource.h"
+#include <memory>
```
gerard-queralt, line 4: This include might not be needed, since it's in the pre-compiled headers, but I could be wrong. Includes that are already in PCH are only needed if the body (implementation) of the methods that use a given header are in the header file, i.e. inlined methods or method templates.
All this to say, try removing it and see if it compiles :P

Source/DataModels/Windows/EditorWindows/WindowInspector.cpp
```diff
@@ -356,15 +361,23 @@ void WindowInspector::DrawTextureOptions()
...
+		ImGui::Text("Width %d", resourceTexture->GetWidth());
+		ImGui::Text("Height %d", resourceTexture->GetHeight());
 		ImGui::EndTable();
 	}
 	ImGui::Text("");
 	if (ImGui::CollapsingHeader("Import Options", ImGuiTreeNodeFlags_DefaultOpen))
 	{
 		ImGui::Checkbox("Flip Image Vertical", &flipVertical);
 		ImGui::Checkbox("Flip Image Horizontal", &flipHorizontal);
+		const char* compression[] = { "No compress",
```
gerard-queralt, line 373: Since this won't change at runtime, I'd move it to an unnamed namespace in the source file. This way, we won't instantiate this array every frame. You can look at [WindowComponentSlider](https://github.com/Horizons-Games/Axolotl-Engine/blob/develop/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentSlider.cpp#L11) for an example.

Source/DataModels/Windows/EditorWindows/WindowInspector.cpp
```diff
@@ -409,6 +423,46 @@ void WindowInspector::DrawTextureOptions()
 		resourceTexture->Unload();
 		resourceTexture->SetChanged(true);
 		App->GetModule<ModuleResources>()->ReimportResource(resourceTexture->GetUID());
+		/*for (GameObject* g : App->GetModule<ModuleScene>()->GetLoadedScene()->GetSceneGameObjects())
```
gerard-queralt, line 426: Is there a purpose for this commented code? (something that should work but doesn't). If there is, please add a comment indicating it, and maybe even a `#pragma message` so it shows up at compilation time and we don't forget about it (like this one in [ComponentPlayer](https://github.com/Horizons-Games/Axolotl-Engine/blob/develop/Source/DataModels/Components/ComponentPlayer.h#L4)). If it's just old code, I generally prefer to simply remove it.

Source/DataModels/Windows/EditorWindows/WindowResources.cpp
```diff
@@ -46,6 +46,7 @@ void WindowResources::DrawResource(const std::weak_ptr<Resource>& resource,
 		ImGui::TextUnformatted(("UID: " + uidString).c_str());
 		ImGui::TextUnformatted(("Name: " + asShared->GetFileName()).c_str());
 		ImGui::TextUnformatted(("AssetPath: " + asShared->GetAssetsPath()).c_str());
+		if (asShared->GetType() != ResourceType::Texture)
```
gerard-queralt, line 49: Is this a typo? Otherwise, please fix the indentation of the next line and surround it with curly brackets (`{}`).

Source/FileSystem/Importers/TextureImporter.cpp
```diff
@@ -103,14 +103,59 @@ void TextureImporter::Import(const char* filePath, std::shared_ptr<ResourceTextu
...
+		result = DirectX::Compress(imgResult->GetImages(),
+								   imgResult->GetImageCount(),
+								   imgResult->GetMetadata(),
+								   compressFormat,
+								   DirectX::TEX_COMPRESS_DEFAULT,
+								   DirectX::TEX_THRESHOLD_DEFAULT,
+								   compressImg);
+
+		if (FAILED(result)) // Compression fails
+		{
```
gerard-queralt, line 143: Log an error, at least. Don't know if the process should continue after failure, seems like `compressImg` is an output parameter. Make sure there won't be any crashes, but ultimately you know this better than I do so I trust your judgement.

Source/FileSystem/Importers/TextureImporter.cpp
```diff
@@ -103,14 +103,59 @@ void TextureImporter::Import(const char* filePath, std::shared_ptr<ResourceTextu
...
+		}
+	}
+	else
+	{
+		HRESULT result = compressImg.InitializeFromImage(
+			*imgResult->GetImages(), false, DirectX::CP_FLAGS_NONE);
+		if (FAILED(result))
+		{
+			// Handle the error appropriately
```
gerard-queralt, line 152: Same thing here, add a `LOG_ERROR`.

Source/FileSystem/Importers/TextureImporter.cpp
```diff
@@ -144,29 +189,78 @@ void TextureImporter::Import(const char* filePath, std::shared_ptr<ResourceTextu
...
 	resource->SetInternalFormat(internalFormat);
 	resource->SetFormat(format);
 	resource->SetImageType(type);
 
-	resource->SetPixelsSize((unsigned int) imgResult->GetPixelsSize());
+	resource->SetPixelsSize((unsigned int) compressImg.GetPixelsSize());
+
+	std::vector<uint8_t> pixels(compressImg.GetPixels(), compressImg.GetPixels() + compressImg.GetPixelsSize());
+
+	resource->SetPixels(pixels);
```
gerard-queralt, line 237: Could we construct the vector inside the call to `SetPixels`? If it's not used elsewhere, I think we'd avoid a memory allocation. Could be wrong on that :P

Source/FileSystem/Importers/TextureImporter.cpp
```diff
@@ -200,27 +294,135 @@ void TextureImporter::Save(const std::shared_ptr<ResourceTexture>& resource, cha
...
 }
 
-void TextureImporter::Load(const char* fileBuffer, std::shared_ptr<ResourceTexture> resource)
+void TextureImporter::Load(const char* filePath, std::shared_ptr<ResourceTexture> resource)
 {
-	unsigned int header[6];
-	memcpy(header, fileBuffer, sizeof(header));
+	std::string narrowString(filePath);
+	std::wstring wideString = std::wstring(narrowString.begin(), narrowString.end());
+	const wchar_t* path = wideString.c_str();
```
gerard-queralt, line 301: I think we can avoid creating these two intermidiate `std::string` and `std::wstring` instances. Take a look at this SO answer: https://stackoverflow.com/a/5503782.
Downside being, you'll need to manage the memory yourself (so, call `delete` once the `wchar_t*` is no longer needed).

Source/FileSystem/Importers/TextureImporter.cpp
```diff
@@ -200,27 +294,135 @@ void TextureImporter::Save(const std::shared_ptr<ResourceTexture>& resource, cha
...
 
-	fileBuffer += sizeof(unsigned char) * resource->GetPixelsSize();
+		std::vector<uint8_t> vpixels(pixels, pixels + imgResult->GetPixelsSize());
 
-	delete[] pixelsPointer;
+		resource->SetPixels(vpixels);
+	}
+	else //Cubemap
+	{
+		char* fileBuffer = {};
```
gerard-queralt, line 401: Pretty sure you're not deleting this `fileBuffer`.

Source/FileSystem/Importers/TextureImporter.cpp
```diff
@@ -200,27 +294,135 @@ void TextureImporter::Save(const std::shared_ptr<ResourceTexture>& resource, cha
...
+		resource->SetPixelsSize(header[5]);
+
+		fileBuffer += sizeof(header);
+
+		unsigned char* pixelsPointer = new unsigned char[resource->GetPixelsSize()];
+		memcpy(pixelsPointer, fileBuffer, sizeof(unsigned char) * resource->GetPixelsSize());
+		std::vector<unsigned char> pixels(pixelsPointer, pixelsPointer + resource->GetPixelsSize());
+		resource->SetPixels(pixels);
+
+		fileBuffer += sizeof(unsigned char) * resource->GetPixelsSize();
```
gerard-queralt, line 421: Maybe a dumb question: why are we incrementing the fileBuffer here, if it's the end of the context where it's declared? In other words: we increment it, but don't do anything with it. This seems to me that this fileBuffer is intended to be declared outside of this block, especially since it seems to be related to the pipeline failure.

Source/FileSystem/ModuleResources.cpp
```diff
@@ -56,7 +56,7 @@ bool ModuleResources::Init()
 
 #ifdef ENGINE
 	CreateAssetAndLibFolders();
-	monitorThread = std::thread(&ModuleResources::MonitorResources, this);
+	//monitorThread = std::thread(&ModuleResources::MonitorResources, this);
```
gerard-queralt, line 59: Please uncomment this 😬

Source/FileSystem/ModuleResources.cpp
```diff
@@ -346,19 +345,61 @@ std::shared_ptr<Resource> ModuleResources::LoadResourceStored(const char* filePa
 
 void ModuleResources::ImportResourceFromLibrary(std::shared_ptr<Resource>& resource)
 {
-	std::string libPath = resource->GetLibraryPath() + GENERAL_BINARY_EXTENSION;
+	std::string libPath;
+	if (resource->GetType() == ResourceType::Texture){
```
gerard-queralt, line 349: Brackets start at the next line, and `else` block should also be between brackets.

Source/FileSystem/ModuleResources.cpp
```diff
@@ -346,19 +345,61 @@ std::shared_ptr<Resource> ModuleResources::LoadResourceStored(const char* filePa
...
 
 	if (resource->GetType() != ResourceType::Unknown && fileSystem->Exists(libPath.c_str()))
 	{
 		if (resource != nullptr)
 		{
-			char* binaryBuffer = {};
-			fileSystem->Load(libPath.c_str(), binaryBuffer);
 
-			switch (resource->GetType())
+			if (resource->GetType() == ResourceType::Texture)
```
gerard-queralt, line 361: Can't we add this in the `switch` statement?

Source/FileSystem/ModuleResources.cpp
```diff
@@ -346,19 +345,61 @@ std::shared_ptr<Resource> ModuleResources::LoadResourceStored(const char* filePa
...
+					case ResourceType::StateMachine:
+						stateMachineImporter->Load(binaryBuffer,
+												   std::dynamic_pointer_cast<ResourceStateMachine>(resource));
+						break;
+					default:
+						break;
+				}
+				delete binaryBuffer;
+				return;
+			/*case ResourceType::Model:
```
gerard-queralt, line 402: Commenting this out doesn't look correct, especially since stuff like Particle Systems are not added again in the new cases. Please review it.

Source/FileSystem/ModuleResources.h
```diff
@@ -155,21 +155,43 @@ const std::shared_ptr<R> ModuleResources::RequestResource(const std::string path
 		std::string libraryPath = CreateLibraryPath(uid, type);
 
 		long long assetTime = fileSystem->GetModificationDate(assetPath.c_str());
-		long long libTime = fileSystem->GetModificationDate((libraryPath + GENERAL_BINARY_EXTENSION).c_str());
-		if (assetTime <= libTime)
+		if (type == ResourceType::Texture)
```
gerard-queralt, line 158: These `if...else` blocks seem to be almost identical, save the path variables. If I'm not missing anything and that's indeed the only difference, just declare them before and unify the code.

Source/FileSystem/ModuleResources.h
```diff
@@ -194,20 +216,42 @@ const std::shared_ptr<R> ModuleResources::RequestResource(const std::string path
...
```
gerard-queralt, line 219: Same thing here. Which, by the way, all of these "exceptions" seem to be caused by the texture binaries having a different extension. Is that correct, and if it is, is it necessary? We'd avoid this type of checks if we could use the binary extension.

Source/DataModels/Resources/ResourceTexture.h
ericdacal: Is removed

Source/DataModels/Windows/EditorWindows/WindowInspector.cpp
ericdacal: Is remove

Source/DataModels/Windows/EditorWindows/WindowInspector.cpp
ericdacal: Now we use a namespace

Source/DataModels/Windows/EditorWindows/WindowResources.cpp
ericdacal: This was to fix that the dds file has is extension in resource window. Now the if is with {} and correctly indentation.

Source/FileSystem/Importers/TextureImporter.cpp
ericdacal: I put LOG_ERROR in this conditionals

Source/FileSystem/Importers/TextureImporter.cpp
ericdacal: I put LOG_ERROR in this conditionals

Source/FileSystem/Importers/TextureImporter.cpp
ericdacal: I add a delete for this variable

Source/FileSystem/Importers/TextureImporter.cpp
ericdacal: Done

Source/FileSystem/Importers/TextureImporter.cpp
ericdacal: I add delete for the wchar_t* variable, for avoid to create the two intermediate instances use toWCharArray of QT?

Source/FileSystem/ModuleResources.cpp
ericdacal: Done

Source/FileSystem/ModuleResources.h
ericdacal: Yes, to can make texture compression is necessary that the texture be in dds format, that's the reason for this exceptions.

Source/FileSystem/ModuleResources.cpp
ericdacal: Done

Source/FileSystem/ModuleResources.cpp
ericdacal: Done

Source/FileSystem/Importers/TextureImporter.cpp
ericdacal: I remove this line because is not necessary

Source/FileSystem/ModuleResources.cpp
ericdacal: I remove the delete

Source/FileSystem/ModuleResources.h
ericdacal: I unify this code to only change the extension if is an image

Source/DataModels/Windows/EditorWindows/WindowInspector.cpp
gerard-queralt: Thank you! Now we can delete this one inside the `Draw` :P

Source/FileSystem/Importers/TextureImporter.cpp
gerard-queralt: That's not what I meant, I wanted to avoid creating the two strings before it. But don't worry, it's a minor thing.
I think you're not allowed to delete the `path` variable since it's managed by the system, could you make sure the code runs correctly? I recall we had a similar issue deleting `const char*` somewhere, but maybe I'm wrong.

Source/FileSystem/ModuleResources.cpp
gerard-queralt: Didn't you override the importing of ParticleSystems? I don't see it in the switch.

Source/FileSystem/ModuleResources.h
gerard-queralt: Thanks for clarifying! Could you do the same thing you did for the comment above this one?

Source/DataModels/Windows/EditorWindows/WindowInspector.cpp
ericdacal: Done

Source/FileSystem/ModuleResources.cpp
ericdacal: This branch was take of develop before ParticleSystem exists.

Source/FileSystem/ModuleResources.h
ericdacal: Done

Source/FileSystem/ModuleResources.cpp
gerard-queralt: Even so, if you look at the diff of `ModuleResources.cpp`, you can see that with your changes, the branch of the switch that handles Particle Systems is deleted. Probably a result of a wrong merge, but please take a look. Try to load a particle system in your branch and see what happens, hoepfully I'm wrong.

ericdacal: You are right, I add the ParticleSystem Load function in the switch

Source/FileSystem/Importers/TextureImporter.cpp
ericdacal: I test the code in Game and Engine and seems to work fine

ericdacal: When I delete the assets, it crash. I remove the delete in line 257 and 396 and now seems to work fine.

gerard-queralt (approved):

MacXxs:
Okey there seems to be a problem 👀 

I've done this by just importing a cube and selecting the "No texture" button in the component mesh renderer. This generates the red background.
Then by just selecting a texture it loads bugged as in the image.

image

MacXxs:
> Okey there seems to be a problem 👀
> 
> I've done this by just importing a cube and selecting the "No texture" button in the component mesh renderer. This generates the red background. Then by just selecting a texture it loads bugged as in the image.
> 
> image

Then, trying to create a new Scene throws an error

image

PR #503, opened 2023-06-18
feat: double click on script window to open it

gerard-queralt:
IMPORTANT: the current code assumes the path to all scripts is `Game/Scripts/{scriptName}`. Let me know if you want to be able to use sub folders, in which case more checks would be necessary.

### DIFF
--- a/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
+++ b/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
@@ -12,6 +12,11 @@
 #include "Math/float3.h"
 #include "ScriptFactory.h"
 
+namespace
+{
+const float doubleClickTimeFrameInS = .5f;
+}
+
 WindowComponentScript::WindowComponentScript(ComponentScript* component) :
 	ComponentWindow("SCRIPT", component),
 	windowUID(UniqueID::GenerateUID())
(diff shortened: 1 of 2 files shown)


markmeliaa (approved):
Works as expected. Just one thing, right now by double clicking, Visual Studio Code is used instead of regular Visual Studio (at least for me), is there anyway you could ensure that regular Visual Studio is always used to open the scripts?

P.S. For now we won't have any subfolders inside `Game/Scripts`, if we end up using subfolders in the future I'll let you know, but with the filters I think we have enough to organize the scripts.

gerard-queralt:
> right now by double clicking, Visual Studio Code is used instead of regular Visual Studio (at least for me), is there anyway you could ensure that regular Visual Studio is always used to open the scripts?

It should be using your system's default program to open a `.cpp` file. I think I should be able to force it to use visual studio 2019, and probably a checkbox in the configuration to choose between this forceful selection or system default. I'll take a look.

markmeliaa:
> I think I should be able to force it to use visual studio 2019, and probably a checkbox in the configuration to choose between this forceful selection or system default. I'll take a look.

That would be great, but only if you have the time! If something else is more important please, don't mind this request and do as you need to do.

Piterga12 (approved):
Works well, good job!

UriMurcia (commented):
Is there a way to open the script with Visual Studio and not a Visual Studio Code?

gerard-queralt:
> Is there a way to open the script with Visual Studio and not a Visual Studio Code?

Coming soon!

PR #504, opened 2023-06-18
feat: overload array subscript operator in ModuleInput

gerard-queralt:
Small QOL, we are able to replace calls like
```c++
input->GetKey(SDL_SCANCODE_UP)
```
for
```c++
(*input)[SDL_SCANCODE_UP]
```
Important to de-reference the pointer, otherwise we'd be calling the operator overload for pointer types, instead of the one in `ModuleInput`.
`GetKey` remains untouched, so this doesn't require any updates to pre-existing code.

### DIFF
--- a/Source/DataModels/Camera/Camera.cpp
+++ b/Source/DataModels/Camera/Camera.cpp
@@ -209,7 +209,7 @@ void Camera::KeyboardRotate()
 
 	ModuleInput* input = App->GetModule<ModuleInput>();
 
-	if (input->GetKey(SDL_SCANCODE_UP) != KeyState::IDLE)
+	if ((*input)[SDL_SCANCODE_UP] != KeyState::IDLE)
 	{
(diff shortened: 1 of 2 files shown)


markmeliaa (commented):
What improvement does this bring? I found the current way of calling the input keys more intuitive and understandable, does this bring any substantial improvement or upgrade?

gerard-queralt:
> What improvement does this bring? I found the current way of calling the input keys more intuitive and understandable, does this bring any substantial improvement or upgrade?

Not really, it's just to give more options to the user of the module. I think this will be the most useful when moving stuff from inside `ModuleInput` to another class. Instead of changing `keysState[SDL_SCANCODE_LCTRL]` to `input->GetKey(SDL_SCANCODE_LCTRL)`, we could just replace `keysState` for `(*input)`. As I said, this doesn't change the existing methods, so everyone can still access it however they prefer.

Dedado3000 (approved):

PR #505, opened 2023-06-18
refactor/chore: address compilation warnings

gerard-queralt:
Tried to avoid changing types as much as possible and use casts instead, but as far as I know it's better to avoid using `unsigned int`, since its size may vary between architechtures and operative systems, and instead use fixed size types, like `size_t` or `uintX_t`.


Source/DataModels/Resources/EditorResource/EditorResource.h
```diff
@@ -4,6 +4,9 @@
 #include "Resources/EditorResource/EditorResourceInterface.h"
 #include "Resources/Resource.h"
 
+#pragma warning(push)
```
markmeliaa, line 7: What does pragma warning push/pop?

markmeliaa (approved):
Amazing work, does this clean all the warnings that we had in the engine?

gerard-queralt:
> Amazing work, does this clean all the warnings that we had in the engine?

Thank you very much! Not _all_ of them, only the compilation ones, which as you can see in the changes were mostly about unchecked size changes. If we run a code analysis we'll still have some, like usage of C-style casts, for instance.

Source/DataModels/Resources/EditorResource/EditorResource.h
gerard-queralt: "Push" will store the current state of warning configuration, and "pop" will restore it. This way, the warnings disabled in between these two directives will only remain disabled for this file, which in most cases it's what we want in order to decide on a case-by-case basis.
Windows documentation: https://learn.microsoft.com/en-us/cpp/preprocessor/warning?view=msvc-170

Dedado3000 (dismissed):
All chances semestre ok.

Source/DataStructures/Quadtree.cpp
```diff
@@ -135,9 +135,9 @@ bool Quadtree::Remove(const GameObject* gameObject)
 	if (!IsLeaf())
 	{
 		bool childrenRemovedObject = frontRightNode->Remove(gameObject);
-		childrenRemovedObject += frontLeftNode->Remove(gameObject);
-		childrenRemovedObject += backRightNode->Remove(gameObject);
-		childrenRemovedObject += backLeftNode->Remove(gameObject);
+		childrenRemovedObject = frontLeftNode->Remove(gameObject) || childrenRemovedObject;
```
PaablooCH, line 138: You can use |= and remove the last part

PaablooCH: Or even merge all into one sentence

PaablooCH (changes requested):
Hi, all seem good, just 2 things:

Source/DataStructures/Quadtree.cpp
gerard-queralt: What I'm not sure about is if lazy evaluation would prevent the expression from being executed, that's why I wrote it in this "clunky" way. I'll check if there'd be any issues and get back to you.

gerard-queralt: I was right that short-circuit evaluation (forgot C++ doesn't have lazy evaluation) would prevent the methods from being executed if a previous one returned true. However, this won't happen using bitwise OR (`|`), which also means that your first suggestion of using `|=` was correct since it's a short-hand for the same operator. If you don't mind, I'll go with that approach because I feel like using `|` might be confusing and lead to thinking it's a mistake.

PaablooCH (approved):

markmeliaa (approved):

Source/DataModels/Resources/EditorResource/EditorResource.h
markmeliaa: Gotcha, thanks!

PR #506, opened 2023-06-19
feat: if no zip file exists, force user to create it

gerard-queralt:
Basically disables the checkbox if the zip is not found. Necessary because not only you need one to run the game, the startup configuration also expects one, so we run into the risk of getting an error. Also added a tooltip to let the user know why the checkbox is disabled.


MarcGonzalezMoratona (approved):
The user experience for this feature is great. The code looks also fine.

Dedado3000 (approved):

PR #511, opened 2023-06-20
perf: improve compile time

gerard-queralt:
- Split UID into a definition (`using` declaration) and actual implementation, since most of the time it was included the latter wasn't needed
- Removed `Globals.h`, split it into smaller files
- Added pre-compiled headers

Profiling compilation time, we can see it has been improved by more than a minute.
- Before:
image

- After:
image

Profiling has been done using DebugEngine configuration.

NOTE: since user scripts will most likely be moved into a different solution, I've excluded them from using the new pre-compiled headers. I'm not sure if new scripts will require them or not.


gerard-queralt:
Latest changes seem to have also positively improved compilation time:
image

Remaining issues:
- Some inlined methods prevent removal of some includes in header files. Might be worth taking a look at "un-inlining" them.
- `IScript` is the biggest include after the changes (and removal of `Globals.h`), according to the profiling tool. I'll try to tidy it up when the latest scripting changes are merged.

markmeliaa:
Do we have to review all 226 changed files? Or which will be the ones that we should pay more atention to? @gerard-queralt

PaablooCH (dismissed):

gerard-queralt:
> Do we have to review all 226 changed files? Or which will be the ones that we should pay more atention to? @gerard-queralt

Most changes in source files are just adding the pre-compiled headers, so no. If you have time, I'd appreciate if you could review:
- Pre-compiled headers (`StdAfx.h`)
    - This should include headers that won't change and are included a significant amount of times. Basically libraries, and I also added our logging class, since I don't expect it to change much. Thanks to this, we can't get rid of most instances of these includes (except, as said above, when methods are inlined). The main thing I'd like you to consider when reviewing is if there's other stuff that we might want to add.
- `Defines` folder: have I split them too much? Too little?
- Any class (header and source) that you're familiar with, maybe you see that some include is not necessary.

But as always, don't feel forced to do so, every piece of feedback is welcomed, even if you didn't review everything 🙂

markmeliaa (dismissed):

markmeliaa (dismissed):

markmeliaa:
whoops, conflicts again @gerard-queralt

markmeliaa (approved):

PaablooCH (approved):

PR #515, opened 2023-06-21
feat: improve the force powers

DADSILVER:
- Added an example of the attack to Bix.
- Added force attack script.
- Set stun time to enemies (maybe need to be changed to new enemyManager).
-When an enemy is on the range, press Q to push them.


Game/Scripts/PlayerForceAttackScript.cpp
```diff
@@ -0,0 +1,115 @@
...
+	const ModuleInput* input = App->GetModule<ModuleInput>();
+	
+	const ComponentTransform* parentTransform =
+		owner->GetParent()->GetComponent<ComponentTransform>();
+	ComponentRigidBody* rigidBody =
+		owner->GetComponent<ComponentRigidBody>();
+
+	rigidBody->SetPositionTarget(parentTransform->GetGlobalPosition());
+
+	if (input->GetKey(SDL_SCANCODE_Q) != KeyState::IDLE && currentCoolDown <= 0)
```
markmeliaa, line 44: Could you add all the logic of the force attack into a separate function so we have all the code regarding to the attack encapsulated?

Game/Scripts/PlayerForceAttackScript.cpp
```diff
@@ -0,0 +1,115 @@
...
+		for (std::vector<GameObject*>::iterator it = enemiesInTheArea.begin(); it < enemiesInTheArea.end();
+			 it++)
+		{
+			const ComponentTransform* enemyTransform =
+				(*it)->GetComponent<ComponentTransform>();
+			ComponentRigidBody* enemyRigidBody =
+				(*it)->GetComponent<ComponentRigidBody>();
+
+			btRigidBody* enemybtRb = enemyRigidBody->GetRigidBody();
+			enemyRigidBody->DisablePositionController();
```
markmeliaa, line 60: Here you are disabling the position and the rotation controller of the rigidbody, but you are never enabling it again, is that what you intended?

markmeliaa (commented):
I'll try the functionality in today's class, but overall, the code looks good, nice job!

Game/Scripts/EnemyDroneScript.cpp
```diff
@@ -45,6 +46,25 @@ void EnemyDroneScript::Start()
...
+	if (stunned)
+	{
+		if(timeStunned < 0)
+		{
+			timeStunned = 0;
+			stunned = false;
+		}
+		else
+		{
+			if (timeStunnedAux < 0)
```
UriMurcia, line 58: What is timeStunnedAux used for?

Game/Scripts/PlayerForceAttackScript.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+void PlayerForceAttackScript::Start()
+{
+}
+
+void PlayerForceAttackScript::Update(float deltaTime)
+{
+	const ModuleInput* input = App->GetModule<ModuleInput>();
+	
+	const ComponentTransform* parentTransform =
```
UriMurcia, line 37: This should be saved as a member variable and do it once at the start, GetComponent function is costly and should be avoided in the update

Game/Scripts/PlayerForceAttackScript.cpp
```diff
@@ -0,0 +1,120 @@
...
+		if (rigidBody == other)
+		{
+			enemiesInTheArea.erase(it);
+		}
+	}
+}
+
+void PlayerForceAttackScript::PushEnemies()
+{
+	const ComponentTransform* transform =
```
UriMurcia, line 82: same for this, in this case it's not in the update, but it's repeated every time we push enemies, if it can be avoided better :)

Game/Scripts/PlayerForceAttackScript.cpp
```diff
@@ -0,0 +1,120 @@
...
+{
+	if (other->GetOwner()->GetTag() == "Enemy")
+	{
+		enemiesInTheArea.push_back(other->GetOwner());
+	}
+}
+
+void PlayerForceAttackScript::OnCollisionExit(ComponentRigidBody* other)
+{
+	for (std::vector<GameObject*>::iterator it = enemiesInTheArea.begin(); it == enemiesInTheArea.end(); it++)
```
UriMurcia, line 69: Maybe try
```c++
enemiesInTheArea.erase(std::remove_if(
	std::begin(enemiesInTheArea),
	std::end(enemiesInTheArea),
	[other](const GameObject* gameObject)
	{
		return gameObject == other->GetOwner();
	}),
	std::end(enemiesInTheArea));
```

Link: https://en.cppreference.com/w/cpp/algorithm/remove

UriMurcia (commented):
Nice work! Will try to test it when I can, but good code!

markmeliaa (approved):

Game/Scripts/EnemyDroneScript.cpp
```diff
@@ -45,6 +46,20 @@ void EnemyDroneScript::Start()
...
+	{
+		if(timeStunned < 0)
+		{
+			timeStunned = 0;
+			stunned = false;
+		}
+		else
+		{
+			timeStunned -= deltaTime;
+			return;
```
PaablooCH, line 59: why this return? you don't want to execute the rest?

Game/Scripts/EnemyDroneScript.cpp
```diff
@@ -103,4 +118,11 @@ void EnemyDroneScript::Update(float deltaTime)
...
 	}
-}
\ No newline at end of file
+}
+
+void EnemyDroneScript::SetStunnedTime(float newTime)
+{
+	stunned = true;
+	timeStunned = newTime;
+	timeStunnedAux = std::min(0.5f, newTime);
```
PaablooCH, line 127: which is the purpose of this variable? i don't see it anymore

Game/Scripts/PlayerForceAttackScript.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+		EnemyDroneScript* enemyDroneScript =
+			(*it)->GetComponent<EnemyDroneScript>();
+
+		enemyDroneScript->SetStunnedTime(stunTime);
+
+		HealthSystem* enemyHealthScript =
+			(*it)->GetComponent<HealthSystem>();
+
+		enemyHealthScript->TakeDamage(10.0f);
```
PaablooCH, line 118: if this value is gonna repeat it we can put it in a define for example

PaablooCH (changes requested):
some questions

Game/Scripts/PlayerForceAttackScript.cpp
DADSILVER: Done.

Game/Scripts/PlayerForceAttackScript.cpp
DADSILVER: In this case, I think we can keep it. because we can pass a level without using this ability . With this, I want to say that I would love the player to use this ability every time that he has it, but he can avoid to uses.

Game/Scripts/PlayerForceAttackScript.cpp
DADSILVER: Sure, I will put this. But I think it may have the same complexity as the for.

image

Game/Scripts/EnemyDroneScript.cpp
DADSILVER: no

Game/Scripts/EnemyDroneScript.cpp
DADSILVER: deleted :)

Game/Scripts/PlayerForceAttackScript.cpp
DADSILVER: Changed, now is a register field.

PaablooCH (approved):
Pana 👍

PR #520, opened 2023-06-23
feat: limit number of console lines drawn

gerard-queralt:
- The user can select the limit from a given set of values.
- Easily extensible: if we want more values, or different ones, we just need to modify the vector in the unnamed namespace.


MacXxs (dismissed):
Perfect! 👍🏼

markmeliaa (dismissed):

markmeliaa (dismissed):

Source/DataModels/Windows/EditorWindows/WindowConsole.cpp
```diff
@@ -69,10 +90,58 @@ void WindowConsole::DrawOptionsMenu()
...
+void WindowConsole::DrawMaxLengthSelection()
+{
+	int numButtons = static_cast<int>(consoleLineLengths.size());
+	const char* limitText = "Limit console lines:";
+
+	// Align the buttons to the right
+	ImGuiStyle& style = ImGui::GetStyle();
+
+	float buttonWidth = ImGui::CalcTextSize("XXX").x + style.FramePadding.x * 2.0f;
+	float buttonsWidth = (buttonWidth + style.ItemSpacing.x) * numButtons;
```
MarcGonzalezMoratona, line 108: We can rename this variable to make it easier to differenciate from the `buttonWidth` one

```suggestion
	float buttonContainerWidth = (buttonWidth + style.ItemSpacing.x) * numButtons;
```

Source/DataModels/Windows/EditorWindows/WindowConsole.cpp
```diff
@@ -69,10 +90,58 @@ void WindowConsole::DrawOptionsMenu()
...
+
+	float textWidth = ImGui::CalcTextSize(limitText).x + style.FramePadding.x * 2.0f;
+
+	ImGui::SetCursorPosX(ImGui::GetWindowWidth() - buttonsWidth - textWidth);
```
MarcGonzalezMoratona, line 112:
```suggestion
	ImGui::SetCursorPosX(ImGui::GetWindowWidth() - buttonContainerWidth - textWidth);
```

MarcGonzalezMoratona (approved):

MarcGonzalezMoratona:
I see that if we change the number of lines from 25 to 10 and then try to go back to 25, we are not able to recover the erased lines.

Is this the expected behaviour? I add a couple of screenshots for more context:
image
image
image

gerard-queralt:
> Is this the expected behaviour?

I would say so, but now that you mention it I hadn't considered that. Do you know what's the behaviour in other applications?

Ultimately, I don't think it's a big deal, because all output is saved to a log file, but I appreciate your feedback and opinion in order to improve the user experience. If we agree that it's better to restore old contents, it shouldn't be a big deal to implement 🙂

markmeliaa (dismissed):

markmeliaa (approved):

MarcGonzalezMoratona (approved):

PR #525, opened 2023-06-24
fix: debugdraw spots and points

OriolSors:
The condition was wrong


gerard-queralt (commented):
Whoops, that's my bad, copy-pasted and forgot to invert the condition. I also did it wrong in Area lights, could you update it there as well? I can do it myself if you're okay with it.

OriolSors:
> Whoops, that's my bad, copy-pasted and forgot to invert the condition. I also did it wrong in Area lights, could you update it there as well? I can do it myself if you're okay with it.

I have to go out now, so you can update of course!

gerard-queralt (approved):

PaablooCH (approved):

PR #526, opened 2023-06-24
feat[Graphics]: Lights updated when parent moved & light optimizations

MacXxs:
* Now light inside gameObjects are Updated when their parents are moved
* Optimized light updates so now only the light being modified is updated instead of updating all the lights of the same type
* The same from the previous point but applied to the rendering and the updating of the ssbo of the light type
* Lights being cached so the access to them is faster


Source/DataModels/Scene/Scene.h
```diff
@@ -60,10 +60,22 @@ class Scene
 	void RenderPointLights() const;
 	void RenderSpotLights() const;
 	void RenderAreaLights() const;
+	void RenderAreaSpheres() const;
```
gerard-queralt, line 63: I feel like all of these rendering methods should go in ModuleRender rather than directly on Scene.

Source/DataModels/Scene/Scene.h
```diff
@@ -134,6 +143,11 @@ class Scene
 	std::vector<AreaLightSphere> sphereLights;
 	std::vector<AreaLightTube> tubeLights;
 
+	std::vector<std::pair<const ComponentPointLight*, unsigned int>> cachedPoints;
```
gerard-queralt, line 146: Might be a dumb question, but could we add this information to the respective structs? If so, we would avoid having twice the amount of vectors.

Source/DataModels/Scene/Scene.cpp
```diff
@@ -721,13 +732,128 @@ void Scene::RenderAreaLights() const
...
+	{
+		glBufferSubData(GL_SHADER_STORAGE_BUFFER, 16, sizeof(AreaLightTube) * numTube, &tubeLights[0]);
+	}
+
+	glBindBuffer(GL_SHADER_STORAGE_BUFFER, 0);
+}
+
+void Scene::RenderPointLight(const ComponentPointLight* compPoint) const
+{
+	bool found = false;
```
gerard-queralt, line 771: Instead of having this boolean, could we not just return from the method once we found and process the light? Same for all of the other ones.

Source/DataModels/Scene/Scene.cpp
```diff
@@ -742,6 +868,9 @@ void Scene::UpdateScenePointLights()
 				pl.color = float4(pointLightComp->GetColor(), pointLightComp->GetIntensity());
 
 				pointLights.push_back(pl);
+				cachedPoints.push_back(std::make_pair(pointLightComp, pos));
```
gerard-queralt, line 871: Is this `pos` variable necessary? Couldn't we not do
```suggestion
				cachedPoints.push_back(std::make_pair(pointLightComp, cachedPoints.size()));
```
Same for the other loops.

gerard-queralt (commented):
Left some questions, if you don't mind my meddling.

Source/DataModels/Scene/Scene.h
MacXxs: No it can't be done, that's why it's in a separate vector. If you test it, you will see that it will break the lights because the information in those structs needs to be the same as the structs in the shaders, they contain just the necessarary information for the shaders to render the lights.

Source/DataModels/Scene/Scene.cpp
MacXxs: Yeah we could, but it's a short function, ones it founds the light to process the next iteration will end the function so there's not a big improvement in performance, and for readability it's fine as it is.

Source/DataModels/Scene/Scene.cpp
MacXxs: Yeah we could to do it. Actually, it would be cachedPoints.size() - 1, but i prefer having it in a separate variable. I find it more clear for it's purpose and readability, and the complexity is the same as doing a size().

AlexisDieu:
Changing from spot to point light or other way make the engine crash : 
image
Else the rest seems working correctly !

Source/DataModels/Scene/Scene.h
MacXxs: They are fine in the Scene class with all the information with the lights to render, actually if i recall correctly Carlos told us that ModuleRender shoudln't have all the functions realated to rendering, just do the calls to them, so actually the ModuleRender should just call the rendering functions for the objects to render. And in the case of the lights, for optimization purposes, its rendering (modification of info in the ssbo of the lights) is only done when a value has been modified and updated, so the the call for the rendering isn't done in ModuleRender every frame, but only when a modification has been detected in its transform (so the ComponentLight directly updates the light and the render) or during a modification on the values of the light (so the window calls for the update of the light and it's render).

AlexisDieu (approved):
The bug is in develop so nevermind need another PR for this.

MacXxs:
> Changing from spot to point light or other way make the engine crash : image Else the rest seems working correctly !

This is happening also in develop, it seems that again for the fourth or fifth time someone broke changing from one light to another....

MacXxs:
> > Changing from spot to point light or other way make the engine crash : image Else the rest seems working correctly !
> 
> This is happening also in develop, it seems that again for the fourth or fifth time someone broke changing from one light to another.... We'll have to fix it again in a separate PR

Don't mind me, already fixed it, gonna upload the change in this PR

AlexisDieu (approved):

Source/DataModels/Scene/Scene.h
gerard-queralt: Even if it's not in Module Render, I feel like Scene is getting too bloated. Ideally, it should just represent a Scene in the engine, and have much less logic, at least in my opinion. Maybe we should consider a new module for all of this logic, even if it's in a different MR?

Source/DataModels/GameObject/GameObject.inl
```diff
@@ -24,12 +27,18 @@ std::vector<C*> GameObject::GetComponents() const
...
+									  if (comp)
+									  {
+										  return comp->GetType() == ComponentToEnum<C>::value;
+									  }
 								  }) |
 							  std::views::transform(
 								  [](const std::unique_ptr<Component>& comp)
 								  {
-									  return static_cast<C*>(comp.get());
+									  if (comp)
```
gerard-queralt, line 38: Is this change part of the fix? I could've sworn the components vector should never have null values, and if that's the case, we can save ourselves a check.

gerard-queralt: Also, I don't think this is right, because now the lambda might not return any value. If we need to check for nullity, we could either add a return after the if block or add a filter view before the transform one.

gerard-queralt: > Is this change part of the fix?

Nevermind, I've just seen the other changes in the commit are basically aesthetic. Still, I don't know how there can be null pointers in that vector :/

gerard-queralt: Just repo'd locally. I imagine it's because this call is performed during the destruction of the light object, so the vector hasn't been cleared yet. Would you mind if I updated the code to have the lambda return something? It'd be just integrating the condition into the return statement.

Source/DataModels/Scene/Scene.cpp
```diff
@@ -803,6 +939,9 @@ void Scene::UpdateSceneAreaLights()
 					sl.attRadius = areaLightComp->GetAttRadius();
 
 					sphereLights.push_back(sl);
+					cachedSpheres.push_back(std::make_pair(areaLightComp, posSpheres));
```
PaablooCH, line 942: why you keep this function if you created the other two that do the same?

Source/DataModels/Windows/EditorWindows/WindowScene.cpp
```diff
@@ -11,6 +11,8 @@
 #include "Modules/ModuleUI.h"
 
 #include "Components/ComponentTransform.h"
+#include "Components/ComponentAreaLight.h"
```
PaablooCH, line 14: unused

PaablooCH (changes requested):
Hi, here are somethings that i see, now i'm going to test it

EDIT: seems work fine

Source/DataModels/GameObject/GameObject.inl
PaablooCH: > I imagine it's because this call is performed during the destruction of the light object, so the vector hasn't been cleared yet

yep, is for this reason that it crash

Source/DataModels/Scene/Scene.cpp
MacXxs: Because during initalization its called, and it saves us to go through the vector of objects in scene two times

Source/DataModels/GameObject/GameObject.inl
MacXxs: Yes this is because of the bug that crashes the engine when changing the light type, for example from point to spot

MacXxs: > Just repo'd locally. I imagine it's because this call is performed during the destruction of the light object, so the vector hasn't been cleared yet. Would you mind if I updated the code to have the lambda return something? It'd be just integrating the condition into the return statement.

Yeah, as long as the lights keep working and you can switch between them its fine by me

MacXxs: Did a commit putting the null check inside the return of the lambdas, if you wanted something else go ahead

AlexisDieu (approved):

PaablooCH (approved):

PR #527, opened 2023-06-24
feat: powerUp follows the player

PaablooCH:
Things added:
- If you don't have a powerUp stored, once you approach near one powerUp, it will follow you.
- You can deactivate the debugdraw of the powerUp.


Game/Scripts/PowerUpLogicScript.cpp
```diff
@@ -32,6 +41,32 @@ void PowerUpLogicScript::Update(float deltaTime)
...
+			ownerRb->SetRotationTarget(errorRotation.Normalized());
+			if (powerUpManagerScript->GetSavedPowerUpType() != PowerUpType::NONE)
+			{
+				btRigidBody* btRb = ownerRb->GetRigidBody();
+				btRb->setLinearVelocity(btVector3(0.0f, 0.0f, 0.0f));
+				isSeeking = false;
+			}
+		}
+#ifdef DEBUG
+		if (powerUpManagerScript->GetDebugDraw())
```
markmeliaa, line 65: Due to de `#ifdef`, this check (and thus the function `GetDebugDraw()`) is not needed.

Game/Scripts/PowerUpsManagerScript.h
```diff
@@ -22,6 +23,9 @@ class PowerUpsManagerScript : public Script
 	const PowerUpType& GetSavedPowerUpType() const;
 	const PowerUpType& GetActivePowerUpType() const;
 
+	const float GetRadiusSeeking();
+	const bool GetDebugDraw();
```
markmeliaa, line 27: There's no need to add `const` both in this funcion and the function above, they are returning primitives.

markmeliaa (approved):
Everything works correctly, just a couple comments

DADSILVER (approved):

Game/Scripts/PowerUpLogicScript.cpp
PaablooCH: GetDebugDraw is done to give the ability to the user to disable the debugdraw through the scripts, is just to don't have too much information in the scene

PR #529, opened 2023-06-26
feat: add search bar to WindowHierarchy

gerard-queralt:
- Filter game objects in the hierarchy, only draw those which's names contain the string written by the user.
- Filter is only set up when the search bar text changes, so we don't iterate the whole hierarchy each frame. When filter is set up, hierarchy is only traversed once (cost should be N).
- Use `unordered_set`, which according to the internet, is the most efficient container to use for searches.

### DIFF
--- a/Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
+++ b/Source/DataModels/Windows/EditorWindows/WindowHierarchy.cpp
@@ -52,13 +52,13 @@ void WindowHierarchy::DrawWindowContents()
 	lastSelectedGameObject = App->GetModule<ModuleScene>()->GetSelectedGameObject();
 }
 
-bool WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject)
+WindowHierarchy::DrawHierarchyResultCode WindowHierarchy::DrawRecursiveHierarchy(GameObject* gameObject)
 {
 	assert(gameObject);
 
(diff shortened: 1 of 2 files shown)


markmeliaa (approved):
Everything works correctly, but in my opinion, an improvement could be made. Right now, when you search for the name of a GameObject, only the parent of that GameObject shows up, and ideally the GameObject itself should be showing up with its parent open.

E.G.: when I try to search for `RootNode` in the Gameplay scene, the hierarchy shows `Bix` and `Drones` alone, without them being opened by the `RootNode`. Ideally, both `Bix` and `Drones` will still show up, but they will be opened by the `RootNode` (idk if that could be done or if its too much work so don't mind me if you do not have the time to perform this improvement).

gerard-queralt:
> ideally the GameObject itself should be showing up with its parent open.

Done, was pretty easy. What I don't like now is that it leaves a lot of nodes open, and for the life of me I haven't been able to make them all collapse :/

markmeliaa:
> Done, was pretty easy. What I don't like now is that it leaves a lot of nodes open, and for the life of me I haven't been able to make them all collapse :/

Don't worry, I still prefer this way so the node you searched is showed directly in the hierarchy eventhough it leaves a lot of nodes open, thanks for the update!

markmeliaa (approved):

DADSILVER (approved):

PR #539, opened 2023-06-27
fix: Force is unattached from camera

OriolSors:
- The Force mechanic now doesn't need the Camera
- The Player cannot move while using the Force
- The Player rotation updates the hitted game object
- Added vertical offset when the force starts


Game/Scripts/PlayerRotationScript.cpp
```diff
@@ -5,36 +5,50 @@
...
 void PlayerRotationScript::Start()
 {
-	rotationSensitivity /= 50.0f;
+	
 }
 
 void PlayerRotationScript::PreUpdate(float deltaTime)
 {
-	Rotation(deltaTime);
+	if (owner->GetComponent<PlayerForceUseScript>()->IsForceActive())
```
PaablooCH, line 28: Avoid getting components in pre/update/post, store them in class variable

PaablooCH (changes requested):

Game/Scripts/PlayerForceUseScript.cpp
```diff
@@ -165,15 +146,35 @@ void PlayerForceUseScript::Update(float deltaTime)
 		}
 
 		// Get next rotation of game object
-		ComponentTransform* parentTransform = owner->GetParent()->GetComponent<ComponentTransform>();
+		ComponentTransform* transform = owner->GetComponent<ComponentTransform>();
```
markmeliaa, line 149: `GetComponent` calls are very costly, and we are avoiding using them in the `Update`s of the scripts. You could create a class variable in this script, call it `transform` and call the `GetComponent` just in the `Start`.

Game/Scripts/PlayerForceUseScript.cpp
```diff
@@ -74,23 +70,14 @@ void PlayerForceUseScript::Update(float deltaTime)
...
-
-			if (rotationVerticalScript)
+			if (rotationScript)
 			{
-				lastVerticalSensitivity = rotationVerticalScript->GetField<float>("RotationSensitivity")->getter();
-				rotationVerticalScript->GetField<float>("RotationSensitivity")->setter(lastVerticalSensitivity / 2.0f);
+				lastHorizontalSensitivity = rotationScript->GetField<float>("RotationSensitivityHorizontal")->getter();
+				rotationScript->GetField<float>("RotationSensitivityHorizontal")->setter(lastHorizontalSensitivity / 2.0f);
+				lastVerticalSensitivity = rotationScript->GetField<float>("RotationSensitivityVertical")->getter();
+				rotationScript->GetField<float>("RotationSensitivityVertical")->setter(lastVerticalSensitivity / 2.0f);
```
DADSILVER, line 78: Until I know, We have to create the Getter and Setter instead of using GetField(). Mark told me this today.

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -35,7 +36,11 @@ void PlayerMoveScript::Start()
 
 void PlayerMoveScript::PreUpdate(float deltaTime)
 {
-	Move(deltaTime);
+	if (!owner->GetComponent<PlayerForceUseScript>()->IsForceActive())
```
markmeliaa, line 39: Same here, if you could get the script in the `Start` and here just call `forceScript->IsForceActive()`, thanks.

Game/Scripts/PlayerRotationScript.cpp
markmeliaa, line 28: Same here.

Game/Scripts/PlayerForceUseScript.cpp
```diff
@@ -109,19 +96,10 @@ void PlayerForceUseScript::Update(float deltaTime)
...
+		if (rotationScript)
 		{
-			rotationVerticalScript->GetField<float>("RotationSensitivity")->setter(lastVerticalSensitivity);
-		}
-
-		if (moveScript)
-		{
-			moveScript->GetField<float>("Speed")->setter(lastMoveSpeed);
+			rotationScript->GetField<float>("RotationSensitivityHorizontal")->setter(lastHorizontalSensitivity);
+			rotationScript->GetField<float>("RotationSensitivityVertical")->setter(lastVerticalSensitivity);
```
DADSILVER, line 102: Same here.

markmeliaa (approved):
Great work overall, the door of the level 1.1 is not working as expected but we need this merged asap, thanks Oriol!

DADSILVER (approved):
Good Job, Only one thing. When I use the force the rotation is so fast.

PaablooCH (approved):

PR #541, opened 2023-06-27
feat: updated Bix model with new animations

markmeliaa:
Not much to say here, a new Bix was modelled by the art department, so it had to be updated into the game.
Animations were also updated so they feel more smooth and realistic (the dash is not working right now, it's discarded for this vertical slice).

You can try all this in the Gameplay scene.

P.S. old Bix animations, meshes and materials were deleted, that's why you see so many files changed, ignore those.


Game/Scripts/BixAttackScript.cpp
```diff
@@ -163,7 +151,59 @@ void BixAttackScript::CheckCollision()
 	//--Provisional
 }
 
-bool BixAttackScript::isAttackAvailable()
+bool BixAttackScript::IsAttackAvailable()
```
DADSILVER, line 154: Const here

DADSILVER (approved):
Good Job, only 1 thing

PaablooCH (approved):

PR #547, opened 2023-06-27
refactor: open scripts with a button

gerard-queralt:


markmeliaa (commented):
Instead of opening the scripts with a specific keyboard/mouse input (LALT + Click seems super wacky), why don't we add an "Open Script" button under the "Remove Script" one so it is a lot more intuitive?

gerard-queralt:
Alas, there goes my dream of avoiding this button :_)

markmeliaa:
> Alas, there goes my dream of avoiding this button :_)

Why would you like to avoid this button?

gerard-queralt:
To mimic Unity, and I also think it's better as a concept to be able to open scripts with a double click. It still could be done if we told ImGui to ignore clicks on widgets (buttons, fields...), but it's not a big deal in the grand scheme of things.

markmeliaa (approved):
Well, I thought the button would look better, but it looks awful in the ComponentScript UI. Anyway, that would be enough for now, if we find a better solution another day we can always change it.

PaablooCH (approved):

PR #557, opened 2023-06-28
feat: Camera Sample focus offset and fixed offset

OriolSors:
New features "à la carte":

- New Focus Offset: custom default in Camera config and custom for each camera sample. This focus can be modified horizontally and vertically.
- New Fixed Offset: now you can stablish a sample point to be fixed in terms of position offset, setting a certain 3D point as target. The orientation continues focusing the player + focus offset (let me know if this fits you)

You can enable or disable these options on each sample point.


PaablooCH (approved):

Dedado3000 (approved):
All seems be working, thanks for updating the scene

PaablooCH (approved):

PR #558, opened 2023-06-28
feat: add Vector fields to scripts

Piterga12:
Implemented the vectors and it's serialization in scripting:
-Vector Float
-Vector String
-Vector Bool
-Vector GO
-Vector Float3
(the vector can have as many components as the user defines on the script)
(there is an example in DefaultScript)

### DIFF
--- a/Source/DataModels/Components/ComponentScript.cpp
+++ b/Source/DataModels/Components/ComponentScript.cpp
@@ -109,6 +109,37 @@ void ComponentScript::SaveOptions(Json& meta)
 				break;
 			}
 
+			case FieldType::VECTOR:
+			{
+				VectorField vectorField = std::get<VectorField>(enumAndValue.second);
+				std::function<std::any(const std::any&, const std::string&)> elementDrawer =
+					[this, &vectorField](const std::any& value, const std::string& name) -> std::any
+				{
+					switch (vectorField.innerType)
+					{
+					case FieldType::FLOAT:
+						return float();
+					case FieldType::STRING:
+						return std::string();
+					case FieldType::BOOLEAN:
+						return bool();
+					case FieldType::GAMEOBJECT:
+						return std::any();
+					case FieldType::VECTOR3:
+						return float3();
+					break;
+					}
+					return std::any();  // Default return 
+				};
+
+				std::vector<std::any> vectorValue = vectorField.getter();
+
+				for (int i = 0; i < vectorValue.size(); ++i)
+				{
+					vectorValue[i] = elementDrawer(vectorValue[i], vectorField.name + std::to_string(i));
+				}
+			}
+
 			case FieldType::STRING:
 			{
 				field["name"] = std::get<Field<std::string>>(enumAndValue.second).name.c_str();
--- a/Source/DataModels/Components/ComponentScript.cpp
+++ b/Source/DataModels/Components/ComponentScript.cpp
@@ -319,7 +319,7 @@ void ComponentScript::LoadOptions(Json& meta)
 			{
 				continue;
 			}
-			std::vector<std::any> vectorCase = vectorField.value().getter();
+			std::vector<std::any> vectorCase;
 
 			for (unsigned int j = 0; j < vectorElements.Size(); ++j)
 			{
@@ -330,19 +330,19 @@ void ComponentScript::LoadOptions(Json& meta)
 				{
 				case FieldType::FLOAT:
 
-					vectorCase[j] = (float)vectorElements[j]["value"];
+					vectorCase.push_back((float)vectorElements[j]["value"]);
 
 					break;
 
 				case FieldType::STRING:
 
-					vectorCase[j] = (std::string)vectorElements[j]["value"];
+					vectorCase.push_back((std::string)vectorElements[j]["value"]);
 
 					break;
 
 				case FieldType::BOOLEAN:
 
-					vectorCase[j] = (bool)vectorElements[j]["value"];
+					vectorCase.push_back((bool)vectorElements[j]["value"]);
 
 					break;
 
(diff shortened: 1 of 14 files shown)


Source/DataModels/Components/ComponentScript.cpp
```diff
@@ -90,60 +90,123 @@ void ComponentScript::SaveOptions(Json& meta)
...
+			for (int i = 0; i < vectorValue.size(); ++i) {
+
+				switch (vectorField.innerType)
 				{
-					field["value"] = 0;
-				}
+				case FieldType::FLOAT:
+					vectorElementsWithName[i]["name"] = std::string(vectorField.name + std::to_string(i)).c_str();
+					vectorElementsWithName[i]["value"] = std::any_cast<float>(vectorValue[i]);
+					vectorElementsWithName[i]["innerType"] = 0;
```
gerard-queralt, line 164: I don't think this `innerType` needs to be saved for each of the elements of the vector, we can probably just do:
```c++
vectorElements["innerType"] = 0;
```
But most importantly, I don't like that we're using integers directly here, since we already have an enum. I'd write it:
```c++
vectorElements["innerType"] = static_cast<int>(vectorField.innerType);
```
To be perfectly clear, what you did doesn't have any issues in terms of performance or robustness (any change in the order of the enum will still break it), but I think my suggestion makes it more readable, and you avoid having to check the value in the switch like you're doing.

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
```diff
@@ -22,6 +24,80 @@ WindowComponentScript::~WindowComponentScript()
 {
 }
 
+std::string WindowComponentScript::DrawStringField(std::string value, const std::string name)
+{
+	if (ImGui::InputText(name.c_str(), value.data(), 24))
```
gerard-queralt, line 29: In all of these `DrawXXXField` methods (except GameObject), you are returning the value regardless, so we can probably just do
```c++
ImGui::InputText(name.c_str(), value.data(), 24);
return value;
```
Same for the others.

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.h
```diff
@@ -19,6 +24,12 @@ class WindowComponentScript : public ComponentWindow
 private:
 	void ChangeScript(ComponentScript* newScript, const char* selectedScript);
 
+	std::string DrawStringField(std::string value, const std::string name);
```
gerard-queralt, line 27: Remember to pass strings as const reference to avoid creating copies 😉 
```suggestion
	std::string DrawStringField(const std::string& value, const std::string& name);
```
Same for the others.

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.h
```diff
@@ -19,6 +24,12 @@ class WindowComponentScript : public ComponentWindow
...
+	bool DrawBoolField(bool value, const std::string name);
+	float DrawFloatField(float value, const std::string name);
+	math::float3 DrawFloat3Field(math::float3 value, const std::string name);
+	GameObject* DrawGOField(GameObject* value, const std::string name);
```
gerard-queralt, line 31:
```suggestion
	GameObject* DrawGameObjectField(GameObject* value, const std::string name);
```
Don't be lazy :P

markmeliaa (changes requested):
Can we have, in the ScriptComponent UI, a better display of vectors? Something like:
`random boolean from the script`
`random float from the script`
`vector name`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; `vector value 1`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; `vector value 2`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; `vector value 3`
`anoter random bool`
This way we could visually diferentitate between when a value is on a vector and when a value is on its own.

(Haven't checked the code, this is just an idea that came after seeing the image in yesterday's class presentation)

Source/DataModels/Components/ComponentScript.cpp
Piterga12: Done!

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.h
```diff
@@ -19,6 +24,12 @@ class WindowComponentScript : public ComponentWindow
 private:
 	void ChangeScript(ComponentScript* newScript, const char* selectedScript);
 
+	std::string DrawStringField(char* value, const std::string& name);
+	bool DrawBoolField(bool value, const std::string& name);
+	float DrawFloatField(float value, const std::string& name);
+	math::float3 DrawFloat3Field(math::float3 value, const std::string& name);
```
gerard-queralt, line 30: Whoops, just realized that this float3 is passed by copy.
```suggestion
	math::float3 DrawFloat3Field(const math::float3& value, const std::string& name);
```
Sorry for not noticing before.

Piterga12:
> Can we have, in the ScriptComponent UI, a better display of vectors? Something like: `some random boolean from the script` `some random float from the script` `vector name`        `vector value 1`        `vector value 2`        `vector value 3` `another random boolean from the script` This way we could visually diferentitate between when a value is on a vector and when a value is on its own.
> 
> (Haven't checked the code, this is just an idea that came after seeing the image in yesterday's class presentation)



> Can we have, in the ScriptComponent UI, a better display of vectors? Something like: `some random boolean from the script` `some random float from the script` `vector name`        `vector value 1`        `vector value 2`        `vector value 3` `another random boolean from the script` This way we could visually diferentitate between when a value is on a vector and when a value is on its own.
> 
> (Haven't checked the code, this is just an idea that came after seeing the image in yesterday's class presentation)

It should look like that now

markmeliaa:
> It should look like that now

Thanks! Could you add some kind if indentation (a TAB would be more that enough IMO) to every member of a vector, so each item of the vector is clearly differentiated from those elements that are not in a vector?

Piterga12:
> > It should look like that now
> 
> Thanks! Could you add some kind if indentation (a `TAB` would be more that enough IMO) to every member of a vector, so each item of the vector is clearly differentiated from those elements that are not in a vector?

Done, now the value appears indented

markmeliaa (commented):
Nice, thanks Peter! I'll try it tomorrow and if everything works you'll have my approve!

markmeliaa (approved):
Looks amazing now, thanks Peter!

gerard-queralt (approved):

PR #566, opened 2023-06-29
fix: don't init/start scripts while scene is loading

gerard-queralt:


markmeliaa (approved):
Works nice, ty for the fix.

AuxPort66 (approved):

PR #570, opened 2023-07-02
feat: detect controller input

MarcGonzalezMoratona:
Proof of concept to detect controller input via SDL.

By default it maps the buttons to an XBOX controller.

Right now we are detecting A,B,X,Y, dpad, left joystick, options, back and shoulder buttons.
Right joystick is missing. ZR, ZL triggers are missing too.

We are making a debug log on every button pressed. Be patient when testing joystick movement, it makes a lot of iterations in a loop for debug logs

Feel free to comment or collaborate with code to this PR


Source/Modules/ModuleInput.cpp
```diff
@@ -162,6 +169,137 @@ UpdateStatus ModuleInput::Update()
...
+
+			case SDL_JOYAXISMOTION: 
+				if (controller)
+				{
+					int axis = sdlEvent.jaxis.axis;
+					Sint16 axisValue = sdlEvent.jaxis.value;
+					switch (axis)
+					{
+						case 0:
+							if (axisValue < -3200)
```
MarcGonzalezMoratona, line 270: this is a safety 10% recommended by SDL docs for the joystick movement

PR #572, opened 2023-07-03
fix: stop Bix movement and allow for the death screen to load

markmeliaa:
Now all the scripts that Bix has get disabled upon death except the one in charge of loading the death screen.

### DIFF
--- a/Game/Scripts/PlayerDeathScript.cpp
+++ b/Game/Scripts/PlayerDeathScript.cpp
@@ -38,7 +38,8 @@ void PlayerDeathScript::ManagePlayerDeath() const
 	{
 		LOG_VERBOSE("Player is dead");
 	}
-	//DisablePlayerActions();
+
+	DisablePlayerActions();
 }
 
 void PlayerDeathScript::DisablePlayerActions() const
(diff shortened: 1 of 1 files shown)


Game/Scripts/PlayerDeathScript.cpp
```diff
@@ -48,7 +49,10 @@ void PlayerDeathScript::DisablePlayerActions() const
 
 	for (ComponentScript* script : gameObjectScripts)
 	{
-		script->Disable();
+		if (script->GetConstructName() != "PlayerDeathScript")
```
UriMurcia, line 52: This made me think, do we have a state machine for the player? Should we start doing one before it's too late? Or one enum state like in the drones. Maybe we already have it, but I can't find it

markmeliaa: We were asked to move the enum states of the drones into actual state machines (which will make the enum states disappear because we'll have those state machines to manage the behaviours), so we could leverage that task and create a state machine for the player too, well thought!

P.S. Did this had anything to do with the line of code that you highlighted? Or was it just a random thought?

UriMurcia: Okay perfect! No, just that I think that in the future it can be a problem if we deactivate all the scripts when we die, that's why I thought that controlling it with a state machine would be much better. But for the moment it's the best idea I can think of, so great job!

UriMurcia (approved):

DADSILVER (approved):

PR #578, opened 2023-07-08
feat: add Components and Scripts as valid fields in Scripts

gerard-queralt:
Simply drag a GameObject into the UI element, and the corresponding Component/Script will be "extracted" from it.

Also removed `REGISTER_VECTOR_FIELD` macro and allowed vectors to be used with the generic `REGISTER_FIELD` one, with the "drawback" of having to specify the full type (`std::vector<X>`) instead of just the element type (`X`). Not really a drawback IMO, since it makes it clearer, but worth pointing out.


gerard-queralt:
Also, vectors of either Components or Scripts are **not allowed** (won't compile). The code is there, commented out, but uncommenting it causes conflicts with the other template overloads. I'll keep working on it, and appreciate any brave souls willing to lend a hand 😅

markmeliaa (commented):
I've been looking at the code and everything seems good, just a stupid question... what do you do to extract the script when a GameObject is dragged into a field in the user interface? I'm failing to see that in the code, I'm a bit slow today, sorry.

gerard-queralt:
Please, don't worry! I'm more than pleased to answer your questions 🙂 

The trick is in the "getters" and "setters" of the Field. When you call `REGISTER_FIELD`, it now redirects to a method that has overloads for those types that require special treatment, such as Vectors or Components. These overloads basically wrap the typical getter and setter (either directly accessing the member or the function if `ACCESSORS` are specified). You can see this overload in [line 90 of `RegisterFieldMacros.h`](https://github.com/Horizons-Games/Axolotl-Engine/blob/scripting/component-field/Source/Scripting/RegisterFieldMacros.h#L90). But the method is small enough to paste it here :P
```c++
Field<GameObject*> field(
	name,
	[getter]()
	{
		Comp member = getter();
		return member ? member->GetOwner() : nullptr;
	},
	[setter](GameObject* value)
	{
		setter(value ? value->GetComponent<std::remove_pointer_t<Comp>>() : nullptr);
	});
return std::make_pair(FieldType::GAMEOBJECT, field);
```
What we do in the code is that we convert the field to one of type `GameObject`. Remember that these `Field`s are only required for UI and serialization (i.e. internal for the engine) and, in actuallity, for both of those purposes what we want is the GameObject.

What the getter does is try to get the actual component, and if found, return its owner. Again, this is what we want for both the UI and serialization. The setter, when invoked, calls `GetComponent` on the `GameObject` received as parameter. Note that, since you will use `REGISTER_FIELD` with pointer types (like `ComponentTransform*`), we need to remove that pointer, because that's what `GetComponent` expects; hence the usage of `std::remove_pointer_t`.

The advantadge of this approach is that, since `GetComponent` works for both Components and Scripts, it works for both without requiring extra work!

Hopefully I clarified the process a little. Feel free to ask further questions if you have them, and suggest improvements to boost the clarity of the code. The fact that you had to ask suggests I can do a better job documenting how this works 😄

markmeliaa:
Thanks for the clarification, I totally get it now 😄 

Regarding the vector of components, I can't think of any situation right now in which we'll need a GameObject store a vector of Components, so I would say is not needed as for now. 
Aside from that, as you said, a vector of Components for me is a vector in which you could add Components from different GameObjects, that would be the ideal approach. The latter would feel a bit missleading regarding user experience IMO.

markmeliaa (approved):

Piterga12 (approved):
Very good job! This looks much better than what we had, thanks Gerard!

gerard-queralt:
Also, spoiler alert 👀 :
image

PR #579, opened 2023-07-09
feat[WIP]: bloom & tone mapping

PaablooCH:
First approach to bloom, once this pr will be merged, i will do anotaré pr with kawase dual filtering that is more efficient than gaussian, with F1 you can activate/deactivate the bloom.

Tone mapping: with F2 you can switch between 3 different tone mappings, by default is selected the ACESFilm, the next one will be Reinhard (the one we already have) and the last one will be the uncharted 2.


Source/Modules/ModuleRender.cpp
```diff
@@ -378,7 +383,48 @@ UpdateStatus ModuleRender::Update()
 
 	glDisable(GL_BLEND);
 
-	// -- DRAW ALL COMPONENTS IN THE FRUSTRUM --
+	// -------- POST EFFECTS ---------------------
+
+	// Blur bloom
```
MacXxs, line 388: Try putting all of this code for bloom and color correction in a separate void function inside the moduleRender so we don't end up having an inmense Update function. (i'm doing the same for shadows so everything is more readable)

PaablooCH: Ok i will put it in the next pr so i will not have problems with kawase

MacXxs (approved):

PR #580, opened 2023-07-11
feat: StateMachine General

AuxPort66:
GamePlay asked me to make StateMachine a general feature that they can use in a script and not just in ComponentAnimation.

You can also now see in orange the last transition and the actual state of the StateMachine you select with the "Edit StateMachine" button. And the actual parameter state.

(The script owns the StateMachine, so remember to initialize the pointer and delete the pointer)


Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
```diff
@@ -339,6 +378,22 @@ void WindowComponentScript::ReplaceSubstringsInString(std::string& stringToRepla
 	}
 }
 
+void WindowComponentScript::CheckStateMachinesInput(const IScript* scriptObject)
+{
+	if (inputStates.size() != stateMachineCount)
```
gerard-queralt, line 383: Just to double-check, could there be issues with this "plain" check? I recall that I did something similar for the inspector window and it caused issues when changing light types, because even though the size was the same, the contents were not. Doesn't look like this can happen here but I'd rather ask just in case.

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
```diff
@@ -220,6 +227,38 @@ void WindowComponentScript::DrawWindowContents()
 				break;
 			}
 
+			case FieldType::STATEMACHINE:
```
gerard-queralt, line 230: Could you wrap this in a method, similar to what Peter did in #558? Just so there are less conflicts.

Source/Modules/ModuleEditor.h
```diff
@@ -28,7 +28,7 @@ class ModuleEditor : public Module
 
 	void Resized();
 
-	void SetStateMachineWindowEditor(const std::weak_ptr<ResourceStateMachine>& resource);
+	void SetStateMachineWindowEditor(StateMachine*& resourceInstance, std::string& instanceName);
```
gerard-queralt, line 31: Maybe a dumb question, but doesn't the signature make more sense like this:
```suggestion
	void SetStateMachineWindowEditor(StateMachine* resourceInstance, const std::string& instanceName);
```

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
AuxPort66: No, because I want to add an extra step in the input window on the switch. That's something I said in the meeting and it's a draft because the code is not finished, I made this draft because I want gameplay check the functionality and let me know if they need anything else.

Source/DataModels/Components/ComponentAnimation.h
```diff
@@ -43,25 +44,20 @@ class ComponentAnimation : public Component, public Drawable, public Updatable
...
 	void InternalLoad(const Json& meta) override;
 
-	bool CheckTransitions(const State* state, Transition& transition);
 	void SaveModelTransform(GameObject* gameObject);
 	void LoadModelTransform(GameObject* gameObject);
 
 	AnimationController* controller;
-	std::shared_ptr<ResourceStateMachine> stateMachine;
-	std::unordered_map<std::string, TypeFieldPairParameter> parameters;
+	StateMachine* stateMachineInstance;
```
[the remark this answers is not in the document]
AuxPort66: What I want is to put it a forward declaration. That's why is a pointer.

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentScript.cpp
AuxPort66: Ok

markmeliaa (commented):
Hey, just checked the functionality until now and I have a couple things.

First, I've seen how now, the state of the selected statemachine is highlighted in orange, but its next transition is maked as well as the state. Do you think that you could highlight the transition arrow only when that transition is happening?

Second, how do we do to create a new statemachine but we don't want to link it to an animator? Is a new component statemachine yet to be created?

AuxPort66:
> Hey, just checked the functionality until now and I have a couple doubts.
> 
> First, I've seen how now, the state of the selected statemachine is highlighted in orange, but its next transition is maked as well as the state. Do you think that you could highlight the transition arrow only when that transition is happening?

In terms of transitions, when StateMachine detect a valid transition the editor colored the transition on orange, so is colored since the first frame of the transition not before. Also, transitions are something that last two updates. So I decided to keep in orange the last transition you do so we can be able to see it. I don't understand the petition. Contact me in discord to see the problem it will be easier to see.

> Second, how do we do to create a new statemachine but we don't want to link it to an animator? Is a new component statemachine yet to be created?

Like I said in the PR you need to create the stateMachine in the script I add it to registerfield so you can upload the resource on the ImGUI, like you do in the componentAnimation.

markmeliaa:
> Like I said in the PR you need to create the stateMachine in the script I add it to registerfield so you can upload the resource on the ImGUI, like you do in the componentAnimation.

Whoops, my bad. I thought we had to create a component and then assign that component to the script, your approach seems better tho, thanks for pointing it out.

Source/DataModels/Components/ComponentAnimation.h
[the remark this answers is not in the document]
gerard-queralt: Could I ask if there's a particular reason for that? If you're worried about compilation time, you could try profiling with the include and the forward declaration, to see if it's a noticeable dent in performance.

Vlukkee (dismissed):
Very good! Very bueno!

markmeliaa:
Conflicts 😬

Vlukkee (approved):
Approved by the latinos force

markmeliaa (approved):

PR #586, opened 2023-07-14
perf: cleaned up IScript interface

gerard-queralt:
IScript got progressively more bloated, which shouldn't be the case for interfaces. This has a negative impact in compilation time (according to my profiling, `IScript.h` is the heaviest include in the project by a considerable margin). With this refactor I tried to hide as much as possible in the implementation (`Script`), so classes that only need to get or set values in a script object don't need to pull as many headers (like `RegisterFieldMacros`, which should obviously only be included by those who actually need it).

This had the consequence of having to remove `GetField` from the class, since method templates cannot be virtual, so the implementation would have needed to be in `IScript`; alternatively we could've implemented it in `Script` and do a cast when it was needed, but that's terrible design (breaks all abstraction). Besides, it was only used in `WindowComponentScript`, so I just moved it there as a helper function in an unnamed namespace.


Piterga12 (approved):

Game/Scripts/PlayerRotationScript.h
```diff
@@ -37,7 +36,6 @@ class PlayerRotationScript : public Script
 	PlayerForceUseScript* forceScript;
 	ComponentRigidBody* rigidBody;
 	btRigidBody* btRb;
-
 };
 
 inline float PlayerRotationScript::GetHorizontalSensitivity() const
```
markmeliaa, line 41: Please, if you could change this inline to the .cpp that would be great.

Game/Scripts/RangedFastAttackBehaviourScript.cpp
```diff
@@ -85,7 +85,6 @@ void RangedFastAttackBehaviourScript::PerformAttack()
 	script->SetScript(App->GetScriptFactory()->ConstructScript("RangedFastAttackBullet"));
 	script->SetConstuctor("RangedFastAttackBullet");
 	script->GetScript()->SetOwner(bullet);
-	script->GetScript()->SetApplication(App);
```
markmeliaa, line 88: Why was this line deleted? Does the script and the bullet created still work regardless of this line?

Game/Scripts/SeekBehaviourScript.h
```diff
@@ -28,4 +28,9 @@ class SeekBehaviourScript : public Script
 	ComponentTransform* targetTransform;
 	ComponentRigidBody* ownerRigidBody;
 	ComponentTransform* ownerTransform;
-};
\ No newline at end of file
+};
+
+inline GameObject* SeekBehaviourScript::GetTarget() const
```
markmeliaa, line 33: To the .cpp please

Game/Scripts/RangedFastAttackBehaviourScript.cpp
markmeliaa: I tried it myself and they still work indeed, so now I'm curious, how was this line needed before but not now?

gerard-queralt: Because, before, the Application had to be set manually. That kind of makes sense for the owner since it may change (gameplay programmers don't need to change it manually anyway, it's done by the engine). However, we only have one application, and that's never going to change. So what I did was get rid of these methods and set it [in the constructor of Script](https://github.com/Horizons-Games/Axolotl-Engine/pull/586/files#diff-33670c9b222d0578f89534bbaf314f956384bb6bd208533a24b36fccf91ac019). The member `App` shadows the global pointer `App` since they have the same name, but by adding `::` before we tell the compiler we're referring to the one in the global namespace.
As a bit more context, this isn't actually needed right now since you can still access the engine's global namespace. However, if we do manage to get hot reloading working and move all scripts to a different project, you won't be able to, and thus you'll need this member.

markmeliaa: Oh, I see it now, thanks for the explanation!

markmeliaa (approved):

PR #596, opened 2023-07-20
feat : Scaling and Translation on BoundingBox (AABB)

AlexisDieu:
- Add a Bounding Box for Component Particle and Component Line to make them take in count frustum to draw or not
- Put the value "Draw Bouding Box" in ComponentTransform to true allow you to see it and scale or translate it as you wish
- If you wand to load an old scene and apply the serialization modification, take a look at "InternalSave()" and "InternalLoad()" of ComponentTransform.cpp


Source/DataModels/Components/ComponentTransform.h
```diff
@@ -257,4 +292,25 @@ inline void ComponentTransform::Encapsule(const vec* vertices, unsigned numVerti
 inline void ComponentTransform::SetGlobalTransform(const float4x4& transform)
 {
 	globalMatrix = transform;
-}
\ No newline at end of file
+}
+
+inline void ComponentTransform::ScaleLocalAABB(float3& scaling)
```
gerard-queralt, line 297: I don't think neither this method nor the following one should be inclined, they look a tad too big.

gerard-queralt (commented):
As an outsider to this task, would you mind explaining why the bounding box in the transform may need to me moved or scaled? Conceptually, at least to me, that bounding box is kind of an internal thing of the engine, which represents the actual bounds of whatever the object has (be it a mesh or a UI element). As such, it should be immutable. To me, what you're doing here looks deserving of a different, new component (I'd recommend taking a look a Unity's [Bounds](https://docs.unity3d.com/ScriptReference/Bounds.html)). Let me know what you think ;)

AlexisDieu:
> As an outsider to this task, would you mind explaining why the bounding box in the transform may need to me moved or scaled? Conceptually, at least to me, that bounding box is kind of an internal thing of the engine, which represents the actual bounds of whatever the object has (be it a mesh or a UI element). As such, it should be immutable. To me, what you're doing here looks deserving of a different, new component (I'd recommend taking a look a Unity's [Bounds](https://docs.unity3d.com/ScriptReference/Bounds.html)). Let me know what you think ;)

Majoritarily for ComponentLine and ComponentLine but I add it to all of the Transform because
Firstly, I didn't need to repeat it to particle and line (and hypothetically for a new features). Particle and Line need it to don't be draw while not in the frustum because actually both of them, not having a mesh, didn't create a AABB depending of their vertices or more. 
Secondly, sometime you will need to modify the bounding box for particle or componentLine to make it modifiable depending of how far the particle is going etc.

In our case with our implementation I don't think boundingBox should be immutable but I mostly do it like that because I think it's the faster an easiest way.

Hope it answer to everything !

Source/DataModels/Components/ComponentTransform.h
AlexisDieu: inlined you mean ? yeah maybe I can move it

gerard-queralt:
> Hope it answer to everything !

You did, thank you very much for taking the time to do so 🙂

I still don't quite understand why the bounding box needs to be modified for the particle. If it's tied to it, and the transform is moving, shouldn't the bounding box be moved implicitly as well? Or is that not the case currently? And even so, I'm not sure it should be modifiable by the user of the engine throught the UI. Would you mind giving me a use case for this?

But aside from that, I see why you did it this way, and agree with you that it's the best approach given our time constraints.

AlexisDieu:
> > Hope it answer to everything !
> 
> You did, thank you very much for taking the time to do so 🙂
> 
> I still don't quite understand why the bounding box needs to be modified for the particle. If it's tied to it, and the transform is moving, shouldn't the bounding box be moved implicitly as well? Or is that not the case currently? And even so, I'm not sure it should be modifiable by the user of the engine throught the UI. Would you mind giving me a use case for this?
> 
> But aside from that, I see why you did it this way, and agree with you that it's the best approach given our time constraints.

Actually by default particle didn't have bounding Box because the bounding box of all models are create depending of their vertices but particles didn't have so the bounding box was 0 and then they was always drawn. For an exemple, if you make some particle falling from the ceiling or like lava particle that pop from the ground, if the bounding box is to small and you didn't have it in your camera frustum, you will not draw the component and then don't see the particle so make able to modify the it will draw it also if it's not in the camera and that can not be done before except with doing a big bounding box to all particle.

Also maybe it was confuse but by default the box is linked to the transform so moving it will move the box by the same case so don't worry this is not changing, it's just adding a possibility to move it independently !

OriolSors (approved):
Regarding your needs and as we already tested it, I approve it

gerard-queralt:
> it's just adding a possibility to move it independently !

I'm still unsure if that's something we want, but it's a very minor thing regardless. See if you can take a look at my other comments and I'll approve the PR. Thank you very much for taking the time to answer my questions 😄

MacXxs:
Hi Alexis! Some things i've found. When entering play mode, the colliders for bix seem to be displaced, don't know if this is also happening to the other objects.

image

Also, the bbox origin por particles isn't centered, the pivot point should be the center of the bbox, not the lower left corner of the bbox. 

image
image

MacXxs (changes requested):
Requesting changes for what I mentioned before

Source/DataModels/Components/ComponentParticleSystem.cpp
```diff
@@ -133,14 +134,21 @@ void ComponentParticleSystem::Draw() const
 		{
 			instance->DrawDD();
 			//instance->SimulateParticles();
+			//Draw the BoundingBox of ComponentParticle
+			ComponentTransform* transform = GetOwner()->GetComponent<ComponentTransform>();
+			if (transform->IsDrawBoundingBoxes())
```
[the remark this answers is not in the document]
AlexisDieu: I look more into this one, I am not gonna change it because it's use a few times yes but the condition is not always the same

AlexisDieu:
> Requesting changes for what I mentioned before

both fixed normally

gerard-queralt (approved):
Thanks for your work!

PaablooCH (changes requested):
looks good, but i have a question apart of the code, in this video we can see that the upper right particles stop rendering when we don't see the bounding box, but we should keep seeing them, is that the behaviour we want? or instead of having only into account the component we should have also in mind the particles itself? correct me if i'm wrong

https://github.com/Horizons-Games/Axolotl-Engine/assets/62140600/e46d9552-46d7-49e9-aab0-e39d07719d34

AlexisDieu:
> looks good, but i have a question apart of the code, in this video we can see that the upper right particles stop rendering when we don't see the bounding box, but we should keep seeing them, is that the behaviour we want? or instead of having only into account the component we should have also in mind the particles itself? correct me if i'm wrong
> 
>  particulas.mp4

It's the behaviour we want, It's just I didn't modify all the particle of the level manually right now, I add a bounding box that is a cube 1,1,1 so if we want to see it when the camera is down, we need to scale the bounding box in the bottom and then scale the bouding box to the particles itself like you said but it's only the component that will manage that size depending of where and how far the particles are intended to go !

PaablooCH (approved):

PR #598, opened 2023-07-21
feat: Add implementation of NavMesh 

totisama:
To bake the navMesh, there is a new window "Navigation", it has a list of parameters to change how to navMesh will be baked and there is button to bake it. **Only the GO's with the tag "NAVIGABLE" will be taken into account.**

There are two new Components, the AgentComponent and the ObstacleComponent. To see how they work, there is a test scene named "navMeshTest".


OriolSors (approved):
Excellent job @totisama!! As you already showed me how everything works, I approve this PR to our secondary branch

AuxPort66 (approved):

PR #600, opened 2023-07-22
feat: Combat camera

OriolSors:
New combat sample points option:

- A Combat Point will be prioritized over general samples. The order now is the following:

1. Closest point of Combat Points
2. Combat Point with influence 
3. Closest point of General Sample Points
4. General Sample Point with influence

- Different color for distinguish both types
- For enabling Combat Point only tick the checkbox
- For QoL, if you select the parent samples object, all the children will be drawn. If not, only the ones selected will be drawn.

### DIFF
--- a/Game/Scripts/CameraControllerScript.cpp
+++ b/Game/Scripts/CameraControllerScript.cpp
@@ -134,16 +134,33 @@ void CameraControllerScript::CalculateFocusOffsetVector(float2 offset)
 ComponentCameraSample* CameraControllerScript::FindClosestSample(float3 position)
 {
 	ComponentCameraSample* closestSample = nullptr;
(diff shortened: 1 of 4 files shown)


Game/Scripts/CameraControllerScript.cpp
```diff
@@ -134,16 +134,33 @@ void CameraControllerScript::CalculateFocusOffsetVector(float2 offset)
...
+				minDistance = distance;
+			}
 		}
 	}
-	return closestSample;
+
+	if (closestCombatSample)
+		return closestCombatSample;
+	else
+		return closestSample;
```
PaablooCH, line 164: add  braces

PaablooCH (approved):
looks good

Dedado3000 (approved):

PR #601, opened 2023-07-24
fix[gameplay]: enemies range attacks

DADSILVER:
this pr fixes the problems with enemy range attacks. you can test in the scene of Gameplayfulltrial.

- When an object is duplicated in run time. His updatable components would be added to the updateable list.
- Now ranges attacks work with bullets.
- The bullet's origin is changed, Now the bullet is shot from the bullet object position.
- The explosion area of the drone is half.
- Prefab of venomites and drone are updated.


Game/Scripts/RangedFastAttackBehaviourScript.cpp
```diff
@@ -24,12 +24,11 @@ REGISTERCLASS(RangedFastAttackBehaviourScript);
 
 RangedFastAttackBehaviourScript::RangedFastAttackBehaviourScript() : Script(), attackCooldown(5.f), 
 	lastAttackTime(0.f), laserParticleSystem(nullptr),audioSource(nullptr),
-	animation(nullptr), transform(nullptr), bulletOriginGO(nullptr), bulletOrigin(nullptr), loadedScene(nullptr), 
+	animation(nullptr), transform(nullptr), loadedScene(nullptr), 
 	bulletVelocity(0.2f), bulletPrefab(nullptr), needReposition(false), newReposition(0,0,0)
 {
 	REGISTER_FIELD(attackCooldown, float);
 
-	REGISTER_FIELD(bulletOriginGO, GameObject*);
```
markmeliaa, line 32: If the bullet origin is needed no more, could you delete it from the drones and the venomites (both in gameplay and in the prefabs)?

Game/Scripts/RangedFastAttackBullet.cpp
```diff
@@ -13,67 +13,65 @@
...
-	transform->SetLocalPosition(transform->GetGlobalPosition() + transform->GetGlobalForward() * velocity * deltaTime * 1000);
-	transform->UpdateTransformMatrices();
-}
-
-void RangedFastAttackBullet::CheckCollision()
-{
-	Ray ray(transform->GetGlobalPosition(), transform->GetGlobalForward());
-	LineSegment line(ray, rayAttackSize);
-	RaycastHit hit;
+	if (other->IsTrigger())
```
markmeliaa, line 62: Why did you decide to perform this check?

Game/Scripts/RangedFastAttackBullet.h
```diff
@@ -26,6 +28,13 @@ class RangedFastAttackBullet : public Script
...
-	ComponentTransform* transform;
+	ComponentRigidBody* rigidBody;
+	ComponentTransform* parentTransform;
 	ComponentAudioSource* audioSource;
-};
\ No newline at end of file
+};
+
+
+inline void RangedFastAttackBullet::SetBulletVelocity(float nVelocity)
```
markmeliaa, line 37: Afaik, we don't do inlines in gameplay scripts, change this function to the cpp please.

markmeliaa (approved):
Now the bullets feel great, thanks for your work, I know it's been rough, but it's finally done! 🎉

Game/Scripts/RangedFastAttackBullet.h
gerard-queralt: Actually, we just didn't inline at the beginning because it didn't work with hot reloading, but now it does (even though hot reloading doesn't ^^'). So you don't need to take that into consideration.

DADSILVER: I will remove to be like the others scrips :)

Game/Scripts/RangedFastAttackBehaviourScript.cpp
DADSILVER: Yeah, thnx

Game/Scripts/RangedFastAttackBullet.cpp
DADSILVER: If a don´t check IsTrigger() the bullet will collide with areas like force or explosion.

Game/Scripts/RangedFastAttackBullet.h
markmeliaa: > Actually, we just didn't inline at the beginning because it didn't work with hot reloading, but now it does (even though hot reloading doesn't ^^'). So you don't need to take that into consideration.

Ikik, but as we started not using inline, we will keep not using inline if it's not an issue, just so we don't have to change back all the existent functions.

Game/Scripts/RangedFastAttackBullet.cpp
markmeliaa: But, if it works correctly, `OnCollisionEnter` should not be called between a trigger and a non trigger, does the code execute that line of code?

DADSILVER: Is triggered let other rigid pass over this but enter in on collision.

image

PaablooCH (approved):

Game/Scripts/RangedFastAttackBullet.cpp
markmeliaa: Whelp, then its a valid check, gj.

PR #606, opened 2023-07-26
fix: Deserialize works for Component&Script fields, deep-copy of Script instance when copying ComponentScript

gerard-queralt:


markmeliaa (approved):
Ty for this <3

DADSILVER (approved):
Great job :)

PR #608, opened 2023-07-27
Feat: New Force V1.0

Dedado3000:
(Note: This force isn't finished, it has some bugs with the offset of free movement objects, in order to deliver 1st level this milestone this isn't needed and will resume the work in 2nd alpha)

Force Make Some changes:
- You can block Rotation/Position of the axis of an object.
- Now you pickup the object by the position you raycast it.
- Forceable Doors are implemented
- Some changes to give player better sensations about force
- Now you can get up an object using up/down mouse


AuxPort66:
This Change will be pushed with the rest of Level Rework

PR #609, opened 2023-07-27
feat: asynchronous scene load

gerard-queralt:
- Moved all code related to Scene loading to `SceneLoader` file
- Load can be done either synchronous or asynchronous
- User can set a callback which will be invoked when loading process finishes


Source/Engine.vcxproj.filters
```diff
@@ -1253,6 +1253,10 @@
     <ClInclude Include="Auxiliar\ComponentNotFoundException.h">
       <Filter>Auxiliar</Filter>
     </ClInclude>
+
```
MarcGonzalezMoratona, line 1256:
```suggestion
```

Source/Modules/ModuleScene.cpp
```diff
@@ -183,14 +194,18 @@ bool ModuleScene::CleanUp()
...
 }
 
 void ModuleScene::SetSelectedGameObject(GameObject* gameObject)
 {
-	AddGameObjectAndChildren(selectedGameObject);
-	selectedGameObject->SetStateOfSelection(StateOfSelection::NO_SELECTED);
+	if (selectedGameObject)
+	{
+		AddGameObjectAndChildren(selectedGameObject);
+		selectedGameObject->SetStateOfSelection(StateOfSelection::NO_SELECTED);
```
MarcGonzalezMoratona, line 207: just a tiny suggestion, we could rename this state to NOT_SELECTED

Source/Modules/ModuleScene.cpp
```diff
@@ -628,12 +397,16 @@ void ModuleScene::RemoveGameObjectAndChildren(const GameObject* object)
...
 void ModuleScene::ParticlesSystemUpdate(bool forceRecalculate)
 {
-	for(ComponentParticleSystem* particleComponent : loadedScene->GetSceneParticleSystems())
+	for (ComponentParticleSystem* particleComponent : loadedScene->GetSceneParticleSystems())
 	{
 		particleComponent->CheckEmitterInstances(forceRecalculate);
 	}
 }
 
+bool ModuleScene::IsLoading() const
```
MarcGonzalezMoratona, line 406: could this be done inline in the header file?

Source/Modules/ModuleScene.h
```diff
@@ -67,11 +69,12 @@ class ModuleScene : public Module
 	rapidjson::Document tmpDoc;
 	std::map<UID, UID> uidMap;
 
-	bool loading;
+	std::mutex setSceneMutex;
```
MarcGonzalezMoratona, line 72: why is it called setSceneMutex? I would use `sceneMutex` becuase this is not a setter function. But probably there's a different reason for using this name. Can you explain it?

Source/Modules/ModuleScene.h
```diff
@@ -48,15 +52,13 @@ class ModuleScene : public Module
 
 	bool IsLoading() const;
 
+	void AddGameObject(GameObject* object);
```
MarcGonzalezMoratona, line 55: can this GameObject be const?

I don't see the difference for using a const reference in RemoveGameObject, but not using it in AddGameObject. Are we making changes to the object inside AddGameObject method?

Source/Modules/ModuleScene.cpp
MarcGonzalezMoratona: I see that before it was inlined, why this changed?

Source/Auxiliar/SceneLoader.cpp
```diff
@@ -0,0 +1,481 @@
...
+				startLoadThread.detach();
+			});
+	}
+	else
+	{
+		StartLoadScene();
+	}
+}
+
+bool IsLoading()
```
MarcGonzalezMoratona, line 476: why isn't this inlined in the header file?

MarcGonzalezMoratona: and why isn't it const?

Source/Modules/ModuleScene.cpp
gerard-queralt: To avoid including `SceneLoader.h` in the header file. It's most likely not a big deal in terms of performance, since this function is not called a whole lot, and is an improvement in compile time.

Source/Modules/ModuleScene.h
gerard-queralt: Because, originally, I only locked this mutex on the setter function `SetLoadedScene`. That was a mistake, since in general mutexes must be locked on both read and write. Forgot to change the name afterwards, thanks for pointing it out 🙂

Source/Auxiliar/SceneLoader.cpp
```diff
@@ -0,0 +1,481 @@
...
+	GameObject* gameObject;
+	UID parentUID;
+	bool enabled;
+};
+
+std::vector<GameObject*> gameObjects{};
+std::map<UID, GameObjectDeserializationInfo> gameObjectMap{};
+std::map<UID, UID> uidMap;
+
+//////////////////////////////////////////////////////////////////
```
MarcGonzalezMoratona, line 61: If you think using this separators is useful, we should propose it as a standard for our programming rules. Maybe changing this everywhere is a bit difficult, so I would not change the styling if it will not be the same in the whole project. But this is just my opinion, so feel free to disagree

Source/Auxiliar/SceneLoader.cpp
gerard-queralt: Because `SceneLoader` isn't a class, it's a file that contains helper methods to load a Scene. As such, it can't be `const`, since that only applies to class methods. It also can't be inlined since the variable `currentLoadingConfig` that we access inside this method only exists in the source file, because it's an implementation detail that we hide from the caller of the class (which also improves compile time).
Hopefully I explained myself, but please let me know if I didn't.

MarcGonzalezMoratona: I think it was really clear, but I don't see why this is better than setting `currentLoadingConfig` as a private member to hide it from the caller of the class.

MarcGonzalezMoratona: it feels a bit confusing to me because I see two nested namespaces, one named and one unnamed, and probably changes a bit the way we work with this class

Source/Modules/ModuleScene.h
gerard-queralt: Checked, and the issue is that we then call `Add(Non)StaticObject`, which ends up pushing the game object into the corresponding collection. However, your comment led me to find out `GameObject::IsRenderable` can be made const, so there's that 😅

Source/Auxiliar/SceneLoader.cpp
gerard-queralt: I added them mainly to differentiate what the different structs and variables are used for, and ended up adding them in the methods for consistency and because, since they are not declared in the header file, we don't have an easy way to see which methods are in the file. I personally don't think it's necessary to enforce this, I just use them when I feel it helps the readability of the file.

Source/Auxiliar/SceneLoader.cpp
gerard-queralt: The thing is that `SceneLoader` is not a class, it just holds global functions scoped inside a namespace. We could convert it into a class that only has static members (both functions and variables), but personally I prefer this type of pattern of global functions in a namespace. Then, in the source file, we have all the implementation details, like how we actually load a scene and all the data structures we need to do so. It's not only hidden from a member visibility perspective, but also from the IDE and from someone reading the API, with the added benefit of hiding includes.
There's a bit of discussion regarding which approach is better. For starters, here are the top 3 links that came up in my search, but in the end it comes down to coding style and personal preference:
- https://stackoverflow.com/questions/1434937/namespace-functions-versus-static-methods-on-a-class
- https://stackoverflow.com/questions/14361408/using-a-namespace-in-place-of-a-static-class-in-c
- https://stackoverflow.com/questions/63950115/what-to-prefer-namespace-or-static-functions-in-c

Piterga12 (dismissed):
Tested it and works fine! Great job here

MarcGonzalezMoratona (dismissed):

MarcGonzalezMoratona:
There are merge conflicts 👀

Piterga12 (approved):

MarcGonzalezMoratona (approved):

ericdacal (approved):

PR #612, opened 2023-07-28
feat: Allura jumping finisher

markmeliaa:
(This was harder to do than it seemed, 😅)

Allura jumping finisher is completed, you can try it in the Gameplay scene by pressing the T Key (this will later be connected to the combo system, but for now the T Key will serve us so we can try the attack).
As you can see, some scripts were renamed to better match their current functionality.

~This depends on #606 being merged first, which helped in the development of this pr.~

Aside from Allura's finisher, this also fixed a minor bug that was left in #607 when despawning enemies and another one in the `PlayerDeathScript` which checked for a child camera (which changed when the camera system changed, so it was outdated).


Game/Scripts/JumpFinisherAttackBullet.cpp
```diff
@@ -0,0 +1,86 @@
...
+	btRb->setLinearVelocity(
+		btVector3(
+			forward.x,
+			bulletHeightForce,
+			forward.z) * bulletVelocity);
+
+	originTime = SDL_GetTicks() / 1000.0f;
+
+	// Set the forceArea as the new child duplicated and set it to trigger
+	forceArea = owner->GetChildren().front()->GetComponent<JumpFinisherArea>();
```
gerard-queralt, line 73: I know we arrived at this line together, but thinking about it a bit, I don't really like the fact that it overrides whatever the user set through the UI, as it makes the field basically useless. We could get rid of it entirely, then, but I like it because it makes it so we don't rely on conventions (anything can be the force area, not just the child).
I was thinking we could change the current hierarchy (which is `BixPrefab` <- `Bullet` <- `ForceArea`) to something like `BixPrefab` <- `FinisherObjects` <- `Bullet` \ `ForceArea`. That is, have both objects be siblings instead of parent and child. With this, copying the `Bullet` won't copy the `ForceArea`. Then, once you have duplicated the `Bullet`, you can copy the `ForceArea` object and set it afterwards to the field of the script (meaning you'd need to add a setter).
This way we respect what the user set in the fields, and we also don't create two copies of the area. Hopefully that made sense!

gerard-queralt (approved):

DADSILVER (approved):
Look pretty good.

PR #615, opened 2023-07-30
feat: improve dash

MarcGonzalezMoratona:
- Dash can now be done without moving.
- The player dashes in the direction where it looks.
- Dash has been changed to V key.
- Jump script logic is now inside movement script.
- Keyboard input debug options added to the engine.
- Dash during a jump is now allowed and controlled.
- The dash is faster, more powerful and shorter.
- The player does not rotate during a dash.
- New player actions added (jump, double jump, dash, attack)
- Typo fixing
- Setters and getters for some script fields created


Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -154,139 +203,126 @@ void PlayerMoveScript::Move(float deltaTime)
...
-
 	// Dash
-	if (input->GetKey(SDL_SCANCODE_C) == KeyState::DOWN && canDash)
+	if (input->GetKey(SDL_SCANCODE_V) == KeyState::DOWN && canDash)
 	{
-		/*
 		if (!isDashing)
 		{
 			componentAnimation->SetParameter("IsDashing", true);
+			componentAnimation->SetParameter("IsRunning", false);
```
MarcGonzalezMoratona, line 211: I would like to change the name of this parameter to IsWalking, because running is a different action. We should change Bix's state machine accordingly, so I didn't want to change that without asking. Please comment what you think about this. 

@Piterga12 
@AuxPort66 
@markmeliaa 

Everyone is invited to comment about this

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -154,139 +203,126 @@ void PlayerMoveScript::Move(float deltaTime)
...
+			//{
+			//	movement /= 2;
+			//}
 
-		if (nextDash == 0)
-		{
-			canDash = false;
-			nextDash = 3000 + static_cast<float>(SDL_GetTicks());
+			Quat rotation = componentTransform->GetGlobalRotation();
+			float3 dashDirection(0.f, 0.f, 1.f);
```
MarcGonzalezMoratona, line 222: @Piterga12 do you think that it is okay to initialize this to 0,0,1?

I think it is a default value to go front, isn't it? Probably we will never need it, but if we end up with this default value, isn't it better to have a 0,0,0 (meaning no dash) rather than a front dash even when the player is not looking front?

Source/DataModels/Windows/SubWindows/WindowInput.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "ModuleInput.h"
+
+WindowInput::WindowInput() : SubWindow("Input")
+{
+}
+
+WindowInput::~WindowInput()
+{
+}
+void WindowInput::DrawWindowContents()
```
MarcGonzalezMoratona, line 14:
```suggestion

void WindowInput::DrawWindowContents()
```

Game/Scripts/PlayerMoveScript.cpp
diegusmejia117: Hi Marc, i would advise against this as right now we indeed only have the running state(walking was taken down on a game meeting, as it was deemed to be running).

Piterga12: The the commented code in line 171 for run, should be erased?

Game/Scripts/PlayerMoveScript.cpp
Piterga12: I did initialise it like that because otherwise it wouldn't work in some weird cases, but maybe it's not necessary anymore to initialise it like that

Game/Scripts/PlayerMoveScript.h
```diff
@@ -44,15 +48,22 @@ class PlayerMoveScript :
     void Move(float deltaTime);
 	void MoveRotate(const float3& targetDirection, float deltaTime);
 
-	bool GetIsParalized() const;
-	void SetIsParalized(bool isParalized);
+	bool GetIsParalyzed() const;
```
gerard-queralt, line 51:
```suggestion
	bool IsParalyzed() const;
```

Game/Scripts/PlayerJumpScript.h
gerard-queralt, line 1: May I ask why this was moved into `PlayerManagerScript`? As a rule of thumb, the more scripts the better. And if there was a reason for it, why leave this script basically empty?

Game/Scripts/PlayerMoveScript.h
```diff
@@ -44,15 +48,22 @@ class PlayerMoveScript :
...
+	void SetIsParalyzed(bool isParalyzed);
+
+	PlayerActions GetPlayerState() const;
+	void SetPlayerState(PlayerActions playerState);
+
+	bool GetCanJump() const;
```
gerard-queralt, line 57:
```suggestion
	bool CanJump() const;
```

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentTransform2D.cpp
```diff
@@ -1,6 +1,6 @@
 #include "StdAfx.h"
 
-#include "ComponentWindows/WindowComponentTransform2D.h"
+#include "WindowComponentTransform2D.h"
```
gerard-queralt, line 3: I personally prefer full paths, since it's more strict, but that's just me. Feel free to ignore me ;)

Source/DataModels/Windows/SubWindows/WindowInput.cpp
gerard-queralt, line 1: The user can still click on the buttons, even though they do nothing. Could you check if there's an ImGui flag to prevent this?

Source/DataModels/Windows/SubWindows/WindowInput.cpp
```diff
@@ -0,0 +1,40 @@
...
+}
+
+void WindowInput::DrawWindowContents()
+{
+	ModuleInput* input = App->GetModule<ModuleInput>();
+
+	const int numRows = 3;
+	const int buttonsPerRow = 4;
+	const int numButtons = numRows * buttonsPerRow;
+	const char* buttonLabels[numButtons] = { "Q", "W", "E", "R", "A", "S", "D", "F", "Z", "X", "C", "V" };
```
gerard-queralt, line 22: This is all constant, but this arrays are allocated and freed each frame. It'd be better if you could define them in an unnamed namespace in the source file. You can look at [WindowComponentSlider](https://github.com/Horizons-Games/Axolotl-Engine/blob/develop/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentSlider.cpp#L11) for an example.

Game/Scripts/PlayerMoveScript.cpp
MarcGonzalezMoratona: perfect, we can leave it as it is right now, then. Thank you Diego!

Source/DataModels/Windows/SubWindows/WindowInput.cpp
MarcGonzalezMoratona: nice catch, thank you Gerard!

Source/DataModels/Windows/SubWindows/WindowInput.cpp
MarcGonzalezMoratona: I'll check it. I used ImGui::Selectable because buttons force you to click before the action, and selectable make what we need. I'll take a deeper look to the Selectable component

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentTransform2D.cpp
MarcGonzalezMoratona: This was changed because the component was in the wrong folder. I could use full paths, but I think there are too much places to make this change right now

Game/Scripts/PlayerJumpScript.h
MarcGonzalezMoratona: The initial idea was to remove the PlayerJumpScript because we need some information about the jump inside the movement script and I didn't find a better way to get it. I thought that we could delete it because both of them (movement and jump) are basic player actions, and also related to each other.

The file was not deleted because of a merge conflict that messed it up.

Anyway, I'll try to find a better way to comunicate info between scripts without deleting the PlayerJumpScript.

Game/Scripts/PlayerMoveScript.h
MarcGonzalezMoratona: I'll apply this change, it was a method name inherited from previous PRs

Game/Scripts/PlayerMoveScript.h
MarcGonzalezMoratona: Same situation that in the one before

Source/DataModels/Windows/SubWindows/WindowInput.cpp
MarcGonzalezMoratona: image

MarcGonzalezMoratona: After testing a bit, disabled flag does what we need

Source/DataModels/Windows/SubWindows/WindowInput.cpp
MarcGonzalezMoratona: @gerard-queralt should I put only the labels and keys arrays or the numeric constants too?

Game/Scripts/PlayerMoveScript.cpp
MarcGonzalezMoratona: You are right. The static dash does not work without this values.

Game/Scripts/PlayerJumpScript.h
gerard-queralt: My main concern is that we might be able to reuse the jump logic for another game object, like the boss (with some adjustments, of course). If that's not something we want there's not really an issue. An option would be to add the jump script as a field of the move one, but again, might not be necessary. I trust your judgement ;)

Source/DataModels/Windows/SubWindows/WindowInput.cpp
gerard-queralt: Yes, better to have them all in one place. Easier to modify them this way too.

Game/Scripts/PlayerMoveScript.cpp
gerard-queralt: If that's the case, maybe we should initialize it with the forward of the player. Unless I'm misunderstanding your conversation, this seems to work because the player starts facing the Z axis, but that might not be the case in future levels.

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -126,7 +129,7 @@ void PlayerMoveScript::Move(float deltaTime)
 		}
 	}
 	else {
-		if (GetPlayerState() != PlayerActions::WALKING && !isDashing && jumpScript->IsGrounded())
+		if (GetPlayerState() != PlayerActions::WALKING && !isDashing && jumpScript->IsGrounded() && !bixAttackScript->IsAttacking())
```
gerard-queralt, line 132: This is just a coding style thing so feel free to disagree with me, but when conditions get this long I prefer declaring a variable and naming it with what the condition actually is. As someone who has never touched this script, I understand that we are checking if the player is not walking, is not dashing, is grounded and is not attacking. But I have no idea what this all actually means. I need to check the code inside the `if`, and it looks like it means the player is running

Game/Scripts/BixAttackScript.h
```diff
@@ -76,3 +79,28 @@ class BixAttackScript : public Script
...
+
+inline void BixAttackScript::SetIsDeathTouched(bool isDeathTouched)
+{
+	this->isDeathTouched = isDeathTouched;
+}
+
+inline bool BixAttackScript::IsAttackAvailable() const
+{
+	return !isAttacking;
+}
```
markmeliaa, line 106: Good call making this functions inline, but we don't do inlines in gameplay scripts. I know inlines are more optimal, but we started not doing inlines due to having no hot reloading in the header files, so we decided to leave all the small functions in the source files (even after having hot reloading).

So, if you could add them back to the source file please.

Game/Scripts/ComboManager.h
```diff
@@ -48,4 +48,9 @@ class ComboManager : public Script
 
 	float comboTime;
 	float actualComboTimer;
-};
\ No newline at end of file
+};
+
+inline bool ComboManager::IsSpecialActivated() const
```
markmeliaa, line 53: Same here with this inline.

Game/Scripts/DebugGame.cpp
```diff
@@ -195,13 +195,13 @@ void DebugGame::FillHealth() const
...
 {
-	if (!playerHealthSystem->GetIsImmortal()) 
+	if (!playerHealthSystem->IsImmortal()) 
 	{
 		playerHealthSystem->SetIsImmortal(true);
 		LOG_VERBOSE("Immortal ON");
 	}
 
-	else if (playerHealthSystem->GetIsImmortal())
+	else if (playerHealthSystem->IsImmortal())
```
markmeliaa, line 204: Not your fault, but this could be just a simple `else statement`.

Game/Scripts/DebugGame.cpp
```diff
@@ -210,13 +210,13 @@ void DebugGame::BeImmortal() const
...
 {
-	if (!playerAttackScript->GetIsDeathTouched())
+	if (!playerAttackScript->IsDeathTouched())
 	{
 		playerAttackScript->SetIsDeathTouched(true);
 		LOG_VERBOSE("Death Touch ON");
 	}
 
-	else if (playerAttackScript->GetIsDeathTouched())
+	else if (playerAttackScript->IsDeathTouched())
```
markmeliaa, line 219: And same here.

Game/Scripts/HealthSystem.h
```diff
@@ -37,3 +37,33 @@ class HealthSystem : public Script
...
+
+inline void HealthSystem::SetIsImmortal(bool isImmortal)
+{
+	this->isImmortal = isImmortal;
+}
+
+inline float HealthSystem::GetCurrentHealth() const
+{
+	return currentHealth;
+}
```
markmeliaa, line 69: Again, swap these to the source file pls.

Game/Scripts/PlayerJumpScript.h
```diff
@@ -10,37 +10,70 @@ class ComponentAudioSource;
...
+inline void PlayerJumpScript::SetCanJump(bool canJump)
+{
+	this->canJump = canJump;
+}
+
+
+inline PlayerMoveScript* PlayerJumpScript::GetMovementScript() const
+{
+	return moveScript;
+}
```
markmeliaa, line 79: To the source file.

Game/Scripts/PlayerMoveScript.cpp
markmeliaa: Yea, if I'm catching this correctly too, the forward of the player would be a more suitable value than just `1.f` in the z.

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -154,139 +144,126 @@ void PlayerMoveScript::Move(float deltaTime)
...
+			dashImpulse.Normalized();
+
+			// Cast impulse and direction from float3 to btVector3
+			btVector3 btDashDirection(dashDirection.x, dashDirection.y, dashDirection.z);
+			btVector3 btDashImpulse(dashImpulse.x, dashImpulse.y, dashImpulse.z);
+
+			btRigidbody->setLinearVelocity(btDashDirection);
+			btRigidbody->applyCentralImpulse(btDashImpulse);
+
+			isDashing = true;
```
markmeliaa, line 179: Could you encapsulate all this logic of the dash into its own function? It would improve the readability.

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -154,139 +144,126 @@ void PlayerMoveScript::Move(float deltaTime)
...
 	}
 }
 
 void PlayerMoveScript::MoveRotate(const float3& targetDirection, float deltaTime)
 {
-	btTransform worldTransform = btRb->getWorldTransform();
-	Quat rot = Quat::LookAt(componentTransform->GetGlobalForward().Normalized(), targetDirection, float3::unitY, float3::unitY);
-	Quat rotation = componentTransform->GetGlobalRotation();
-	Quat targetRotation = rot * componentTransform->GetGlobalRotation();
+	if (!isDashing)
```
markmeliaa, line 216: Instead of indexing all the code inside this `if statement` you could actually just write
```c++
if (isDashing)
{
   return;
}

//The rest of the code
```
So this way we would have less levels of indentation and the code will look cleaner.

Game/Scripts/PlayerMoveScript.h
```diff
@@ -63,12 +70,40 @@ class PlayerMoveScript :
...
+
+inline void PlayerMoveScript::SetPlayerState(PlayerActions playerState)
+{
+	this->playerState = playerState;
+}
+
+inline PlayerJumpScript* PlayerMoveScript::GetJumpScript() const
+{
+	return jumpScript;
+}
```
markmeliaa, line 109: To the source file.

markmeliaa (changes requested):
Thank you both for your work, I just have one concern aside from some code suggestions.

Right now, if I'm not seeing it wrong, when the character dashes, it automatically gets teleported "x" distance infront of them, which makes me worry about how would we add the animation here. Ideally, instead of a teleporting we need to see the player during the whole transition of the dash, so the animation can be seen correctly. If you have any doubt about that, do not hesitate in contacting me.

markmeliaa:
> We wanted to make the dash a bit more spectacular by increasing the dash force. The player is moving fast, but by reducing this value we can remove this sense of "teleport". We can play with the value to reduce the force and align correctly with the animation once it's ready. I think we should talk about this together to find the best visual feedback for the player

Sure, we could talk about this in tomorrow's game meeting, at 18:30!

markmeliaa:
I've been trying values and seems that 2000 for the dash force is the best value we can give to it right now, so if you change it in the scene and fix the conflicts, we can merge this pr.

markmeliaa (approved):

Game/Scripts/PlayerJumpScript.cpp
```diff
@@ -32,23 +34,19 @@ coyoteTime(0.4f), groundedCount(0), grounded(false)
...
-}
-
 void PlayerJumpScript::Start()
 {
 	rigidbody = owner->GetComponent<ComponentRigidBody>();
 	//Avoid Y rotation on player this should go on movement not here but
 	rigidbody->GetRigidBody()->setAngularFactor(btVector3(1.0f, 0.0f, 1.0f));
 	componentAnimation = owner->GetComponent<ComponentAnimation>();
 	componentAudio = owner->GetComponent<ComponentAudioSource>();
+	moveScript = owner->GetComponent<PlayerMoveScript>();
```
gerard-queralt, line 44: I could be missing something, but are we using this variable anywhere? Because if not, we're potentially creating an unnecessary dependency. With the latest changes in scripting, a script will stop running if one of the components it asks for is not found. That will obviously not happen here, but I think it's a good thing to keep in mind.

diegusmejia117 (approved):

PR #616, opened 2023-07-30
feat: HasComponent method

gerard-queralt:
should avoid many try...catch blocks


markmeliaa (approved):
So, this would avoid doing as much `try...catch` blocks, wouldn't it?

gerard-queralt:
> So, this would avoid doing as much `try...catch` blocks, wouldn't it?

Ideally. From what I've seen, most of the `try...catch` blocks that are currently needed just check for a particular component. As you and David have pointer out, that can be avoided by checking the tag of the game object in the majority of the cases. But there are others, like [this snippet in `ActivationLogic::Start`](

DADSILVER (approved):

PR #617, opened 2023-07-31
feat: allow vectors of Components and Scripts as fields of Scripts

gerard-queralt:
- Use C++ 20 [`concept`s](https://en.cppreference.com/w/cpp/language/constraints) to define what's a component or a script (`IsComponentOrScript`). A `concept` is a way to define a constraint for a template, which is then evaluated at compile time. I've just recently learnt about them so I don't fully understand them yet, so if you want a proper explanation I advise you to take a look at the documentation linked before.
- Use only two overloads, one for vectors and one for the rest, instead of the previous four. To do this, we use [`if constexpr`](https://en.cppreference.com/w/cpp/language/if) (section "Constexpr if" in the link). These type of `if` are evaluated at compile time, and only the block that matches the condition is compiled (from the same link: `If the value is true, then statement-false is discarded (if present), otherwise, statement-true is discarded`). In other words, this has no impact in runtime performance.

New vectors can be tested with `DefaultScript`.

### DIFF
--- a/Game/Scripts/DefaultScript.cpp
+++ b/Game/Scripts/DefaultScript.cpp
@@ -1,5 +1,7 @@
 #include "DefaultScript.h"
 #include "DataModels/Components/ComponentTransform.h"
+#include "DataModels/Components/ComponentAnimation.h"
+#include "DataModels/Components/ComponentMeshCollider.h"
 #include "DataModels/Components/ComponentScript.h"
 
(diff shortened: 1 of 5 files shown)


markmeliaa (approved):

Piterga12 (approved):
Evrything looks great! Good job!

PR #619, opened 2023-08-01
feat: vectors of Components and Scripts as Script Fields + performance improvements

gerard-queralt:
Merges:
- #586 
- #617 

Also added a few improvements and fixes, mainly to memory performance and to the mask used to ignore double clicks in `WindowComponentScript`. You can check the individual commits:
- 92e31df1b5e0ea4b27c39f691cfdfe7fe155f866
- a4d22432d902d7bdc0fb1825e293eeeac4227f35
- f3a1876505ab2d6f5d6ce786113f103924c01967

Or [the unified diff of the three](https://github.com/Horizons-Games/Axolotl-Engine/pull/619/files/269c38686bd95bbefb62b6333f8cfa9379fe682f..f3a1876505ab2d6f5d6ce786113f103924c01967).


Source/Auxiliar/Reflection/VectorField.h
Piterga12, line 1: I understand that now it we get vector and any from FieldType.h, but why is it better to not use const for the members of VectorField?

Piterga12 (approved):

Source/Auxiliar/Reflection/VectorField.h
gerard-queralt: It's not that we're using a non-const reference; we're passing the parameters as rvalues using a double ampersand ([link to the documentation](https://en.cppreference.com/w/cpp/language/reference)). If you want a proper explanation, I encourage you to check the above documentation and other online resources, and also the class we had related to rvalues (and other things). You can find it in the campus, in `C++ for videogames` -> `Advanced C++`.
As a TL;DR, by using rvalues we can optimize variable forwarding amongst our methods. Not only that, we avoid using traditional copy constructors, and instead we use the more efficient `std::move`. Overall, it's more efficient both in memory usage and runtime performance. However, it's not always possible to use rvalues, since it implies passing ownership of the memory. As such, we can only use them when the method will take ownership of that memory, which is the case for the constructors of both `Field` and `VectorField`, as well as existing code in the engine like `Script::AddMember` or upcoming one like most of what I wrote in #609.
Again, this is just my broad understanding of the matter, which is incomplete and might even be wrong, so I recommend you take a look at the resources above 😉

markmeliaa (approved):

PR #624, opened 2023-08-03
feat: bloom modificable for each material

PaablooCH:
As mentioned during wednesday's meeting i added a drag to manipulate the bloom intensity of each material, you can find it below the emissive texture. I have limited the value  between 0.05 and 1. For test it you can use the level 1. Remember that with F1 you can disable the bloom.

### DIFF
--- a/Source/DataModels/Batch/GeometryBatch.cpp
+++ b/Source/DataModels/Batch/GeometryBatch.cpp
@@ -180,7 +180,8 @@ void GeometryBatch::FillMaterial()
 				resourceMaterial->HasEmissive(),
 				resourceMaterial->GetSmoothness(),
 				resourceMaterial->GetMetalness(),
-				resourceMaterial->GetNormalStrength()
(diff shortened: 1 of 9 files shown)


AlexisDieu (approved):
It's working, the only things I want to mention is that it seems the value is only take in account between 0 and 1 , over 1 there is no modificaiton. I don't know if this is the purpose (because you put it to 5 I tried with more but as I said over 1 nothing change)

PaablooCH:
> It's working, the only things I want to mention is that it seems the value is only take in account between 0 and 1 , over 1 there is no modificaiton. I don't know if this is the purpose (because you put it to 5 I tried with more but as I said over 1 nothing change)

i take a look

PaablooCH:
fixed @AlexisDieu

AlexisDieu (approved):

MacXxs (approved):

PR #626, opened 2023-08-03
Feat: Heavy Finisher Attack

UriMurcia:
This pull request contains the heavy finisher attack, in this case for Bix. To test this attack you need to make a combo of three attacks with the last attack as a heavy attack (right button mouse).
We have two types of heavy finisher attack, one where the attack goes to the target selected by EntityDetection, and the other one in case there is no target in the EntityDetection; in this case the attack will go forward to a specified distance and in case of finding an enemy will hit it.


markmeliaa (approved):
GJ, everything works correctly!

Game/Scripts/HeavyFinisherAttack.cpp
```diff
@@ -0,0 +1,315 @@
...
+	physics->GetCollisions(rigidBody, collisions, "Enemy"); //Get enemies inside the collider
+
+	float3 currentPos = transform->GetGlobalPosition();
+
+	for (ComponentRigidBody* enemy : collisions)
+	{
+		ComponentTransform* enemyTransform = enemy->GetOwner()->GetComponent<ComponentTransform>();
+
+		if (!enemy->GetOwner()->GetComponent<HealthSystem>()->EntityIsAlive() //If enemy is dead
+			|| (std::find(enemiesAlreadyHit.begin(), enemiesAlreadyHit.end(), enemyTransform)
```
gerard-queralt, line 236: This works fine, but since you don't actually want to find that value, you just want to know if there's one, I'd suggest you use `std::any_of`. This higher order function returns true if at least one of the elements in the range matches your condition.
```suggestion
			|| (std::any_of(enemiesAlreadyHit.begin(), enemiesAlreadyHit.end(), [&enemyTransform](const ComponentTransform* other){ return other == enemyTransform; }))
```
(I'm on my phone, excuse the formatting :P)

Game/Scripts/HeavyFinisherAttack.cpp
```diff
@@ -0,0 +1,315 @@
...
+		target = attackOwner;
+		isReturningToOwner = true;
+		if (!returnToPlayer)
+		{
+			ResetValues();
+		}
+	}
+}
+
+void HeavyFinisherAttack::ResetValues()
```
gerard-queralt, line 266: I'd suggest overloading `Script::CleanUp` and calling this function from there as well, for extra protection.

Game/Scripts/HeavyFinisherAttack.cpp
```diff
@@ -0,0 +1,315 @@
...
+}
+
+void HeavyFinisherAttack::OnCollisionEnter(ComponentRigidBody* other)
+{
+	if (attackHasTarget)
+	{
+		return;
+	}
+
+	if (other->GetOwner()->GetTag() == "Enemy" && other->GetOwner()->IsEnabled())
```
gerard-queralt, line 294: Based on recent pull requests, looks like the preferred way to make this check is with `CompareTag`. Not sure if there are other instances where you could change it.
```suggestion
	if (other->GetOwner()->CompareTag("Enemy") && other->GetOwner()->IsEnabled())
```

Game/Scripts/HeavyFinisherAttack.h
```diff
@@ -0,0 +1,66 @@
...
+	Scene* loadedScene;
+
+	bool returnToPlayer;
+	bool rotateWhileAttacking;
+	float rotationVelocity;
+	float damage;
+	float speed;
+	float hitDistance;
+	float maxEnemyHits; //Should be int but can't REGISTER_FIELD ints
+	float countEnemyHits;
```
gerard-queralt, line 64: Is this variable necessary? Could you not just get the size of the `enemiesAlreadyHit` collection?

Game/Scripts/HeavyFinisherAttack.h
```diff
@@ -0,0 +1,66 @@
...
+	GameObject* mesh;
+	ComponentRigidBody* rigidBody;
+	ComponentParticleSystem* vfx;
+
+	ComponentTransform* target;
+	ComponentTransform* attackOwner;
+	bool isActivated;
+	bool isReturningToOwner;
+	bool attackHasTarget;
+	std::vector<ComponentTransform*> enemiesInTheArea;
```
gerard-queralt, line 50: Couldn't this vector and the one below be converted to sets? It feels more idiomatic to me.

gerard-queralt (commented):
As usual, haven't tested, but code looks great. Left a few suggestions 😉

Game/Scripts/HeavyFinisherAttack.h
UriMurcia: true xd

Game/Scripts/HeavyFinisherAttack.cpp
UriMurcia: But CleanUp isn't called when the script is deleted?

Game/Scripts/HeavyFinisherAttack.h
UriMurcia: Why we would use the key value for?

diegusmejia117 (approved):
Just testedm works correctly, good job uri!

Game/Scripts/HeavyFinisherAttack.cpp
gerard-queralt: Not quite, it's when the script stops being updated. If I'm not wrong, these variables will be kept alive in memory unless this is calles, and if the script stops before it can call this function for whatever reason, they will be hoarding resources until, when we run the game again, they are overridden. It's definetly not a big deal, but would be nice to avoid.

Game/Scripts/HeavyFinisherAttack.h
gerard-queralt: You might be thinking of a map, which is a collection of key and value pairs. Without getting into implementation details (which, for the most part, I don't even know), a set is like a vector except it doesn't accept duplicates. I don't know if it will be more efficient (I imagine so, element look-up should be easier to do, but I don't actually know for sure), but it makes more sense as a concept, in my opinion, since we only want an enemy to be hit once.

PR #628, opened 2023-08-03
feat: Controller input mapping

MarcGonzalezMoratona:


Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -328,6 +329,12 @@ void GameObject::CopyComponent(Component* component)
 			break;
 		}
 
+		//case ComponentType::PLAYERINPUT:
```
MarcGonzalezMoratona, line 332: @gerard-queralt I can't make this work because it is an abstract class. Do you know how to solve it?

Source/Modules/ModuleInput.h
```diff
@@ -27,8 +27,19 @@ class ModuleInput : public Module
 	UpdateStatus Update() override;
 	bool CleanUp() override;
 
+	int GetAxis() const;
```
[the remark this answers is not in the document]
gerard-queralt: Also, what does this method do exactly, get the current axis of the joystick? If so, maybe we could use a more descriptive name? And also, is it an option for the user to use both sticks? What does this method return in such cases? Same goes for the other getter.

Source/Modules/ModuleInput.h
```diff
@@ -180,3 +223,13 @@ inline KeyState ModuleInput::operator[](SDL_Scancode index)
...
+
+inline int ModuleInput::GetAxis() const
+{
+	return axis;
+}
+
+inline Sint16 ModuleInput::GetAxisValue() const
+{
+	return axisValue;
+}
```
gerard-queralt, line 235: Picky, but add a newline at the end of the file, please 😜

Source/Modules/ModuleInput.h
```diff
@@ -27,8 +27,19 @@ class ModuleInput : public Module
 	UpdateStatus Update() override;
 	bool CleanUp() override;
 
+	int GetAxis() const;
+	Sint16 GetAxisValue() const;
```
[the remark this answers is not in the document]
MarcGonzalezMoratona: It's a signed short int, used for range of values -32768 to 32767

Source/Modules/ModuleInput.h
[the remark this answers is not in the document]
MarcGonzalezMoratona: In this case we need a static cast from Uint8 to SDL_GameControllerAxis
image

Before, I thought that this was only used for the left joystick, but now I see in the docs that it can also be used for the right joystick and the triggers. 

https://wiki.libsdl.org/SDL2/SDL_GameControllerAxis

I renamed the methods accordingly and made changes to use the enum. In this first iteration we don't expect to use triggers or right joystick, but I don't know how we will handle a movement in two joysticks at the same time.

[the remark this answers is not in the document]
gerard-queralt: According to the documentation regarding [SDL_JoyAxisEvent](https://wiki.libsdl.org/SDL2/SDL_JoyAxisEvent), it occurs when a user moves a stick. We'd have to test it, but to me this reads like, if I move the left joystick to the right, it fires a single event, and if I hold it there no more events are fired. If that's indeed the case, we might need a map the same way we do for the buttons. We can try to avoid it for this first iteration, see if it works.

Source/DataModels/Components/ComponentPlayerInput.h
```diff
@@ -0,0 +1,19 @@
...
+class ComponentPlayerInput : public Component, public Updatable
+{
+public:
+	ComponentPlayerInput(bool enabled, GameObject* owner);
+	~ComponentPlayerInput() override;
+
+	void Update() override;
+
+	void InternalSave(Json& meta) override;
+	void InternalLoad(const Json& meta) override;
```
gerard-queralt, line 15:
```suggestion
private:
	void InternalSave(Json& meta) override;
	void InternalLoad(const Json& meta) override;
```

Source/Modules/ModuleInput.cpp
```diff
@@ -17,6 +17,11 @@
 	#include "optick.h"
 #endif // DEBUG
 
+namespace
+{
+	const float COOLDOWN_SECONDS = 0.25f;
```
gerard-queralt, line 22:
```suggestion
const float COOLDOWN_SECONDS = 0.25f;
```
`Ctrl+K`, `Ctrl+D` should automatically fix this.

Source/Modules/ModuleInput.cpp
```diff
@@ -70,7 +75,16 @@ UpdateStatus ModuleInput::Update()
 	{
 		if (keysState[i] == KeyState::DOWN)
 		{
-			keysState[i] = KeyState::REPEAT;
+			keysCooldown[i] -= App->GetDeltaTime();
+			if (keysCooldown[i] <= 0.0f)
```
gerard-queralt, line 79: If I'm reading this correctly, this prevents that, if clicked in quick succession, one of the inputs might get lost (key remains in `REPEAT`). Is that it? I assume you encountered some issues with the gamepad?

Source/Modules/ModuleInput.h
```diff
@@ -54,6 +65,11 @@ class ModuleInput : public Module
 private:
 	KeyState keysState[SDL_NUM_SCANCODES] = { KeyState::IDLE };
 	KeyState mouseButtonState[NUM_MOUSEBUTTONS] = { KeyState::IDLE };
+	KeyState gamepadState[SDL_CONTROLLER_BUTTON_MAX] = { KeyState::IDLE };
+
+	float keysCooldown[SDL_NUM_SCANCODES] = { 0.25f };
+	float mouseButtonCooldown[NUM_MOUSEBUTTONS] = { 0.25f };
+	float gamepadButtonCooldown[SDL_CONTROLLER_BUTTON_MAX] = { 0.25f };
```
gerard-queralt, line 72: This is a bad use of initialization in header files. The issue with this is that, if we ever want to change this value, which we might, we will need to recompile this header, which will then force a recompile of every file that includes it. This is not an issue with things like the `gamepadState` array above, since it's almost certainly never going to change it's default value, but for this one it might. Not only that, but we should use the variable defined in the unnamed namespace in the source file just so it's consistent. You can probably initialize it in the initializer list.

Source/DataModels/Components/ComponentPlayerInput.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "Application.h"
+#include "Modules/ModuleInput.h"
+
+namespace
+{
+const std::map<SDL_GameControllerButton, SDL_Scancode> defaultGamepadMapping = {
+	{ SDL_CONTROLLER_BUTTON_A, SDL_SCANCODE_SPACE },
+	// O has to be used everywhere as left mouse button
+	// and P as right mouse button
+	// because this map cannot hold mouse buttons
```
gerard-queralt, line 14: My bad, this is a pretty big limitation. We might be able to just run with it for this iteration, but we should address this in the following one. Maybe we need two maps? Another option is to use an `std::variant`, but I'm not sure if it'll work since both of these are unescoped enums.

Source/Modules/ModuleInput.cpp
```diff
@@ -17,6 +17,11 @@
 	#include "optick.h"
 #endif // DEBUG
 
+namespace
+{
+	const float COOLDOWN_SECONDS = 0.25f;
+} // namespace
+
 ModuleInput::ModuleInput() : mouseWheel(float2::zero), mouseMotion(float2::zero), mousePosX(0), mousePosY(0)
```
gerard-queralt, line 25: Regarding the initializer listst comment, you might be able to do this
```suggestion
ModuleInput::ModuleInput() : mouseWheel(float2::zero), mouseMotion(float2::zero), mousePosX(0), mousePosY(0), keysCooldown{ COOLDOWN_SECONDS }
```
Same for the rest.

Source/Modules/ModuleInput.cpp
MarcGonzalezMoratona: Yes, I found a problem when pressing a button on the gamepad controller. The state goes directly from IDLE to REPEAT

Source/DataModels/Components/ComponentPlayerInput.cpp
MarcGonzalezMoratona: I tried to do it using a std::variant but there is no enum for the mouse buttons, so I didn't know how to type it

Source/Modules/ModuleInput.h
MarcGonzalezMoratona: nice catch, thank you!

Source/Modules/ModuleInput.cpp
MarcGonzalezMoratona: done!

Source/Modules/ModuleInput.h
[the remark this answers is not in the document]
MarcGonzalezMoratona: Check this, because this one is more specific for controllers, the old one is for joysticks
https://wiki.libsdl.org/SDL2/SDL_ControllerAxisEvent

Source/Modules/ModuleInput.cpp
gerard-queralt: I'll try and find a reason for this, unless I'm reading the code wrong it shouldn't happen. Not saying your solution is not needed, but I'd like to understand better the issue.

Source/DataModels/Components/ComponentPlayerInput.cpp
gerard-queralt: Looks like it's a plain `Uint8`, according to the documentation. We could try that. If you decide to go this route, you can use [holds_alternative](https://en.cppreference.com/w/cpp/utility/variant/holds_alternative) to discriminate each case.

Source/Modules/ModuleInput.cpp
MarcGonzalezMoratona: It's not after multiple inputs. The first time you press a button on the controller it goes directly to REPEAT state

Source/Modules/ModuleInput.h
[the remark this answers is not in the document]
MarcGonzalezMoratona: I finally found a way to get input from both axis at the same time. I made a Direction enum class that holds the movement direction in a pair, one value for x and another for y axis

Source/Modules/ModuleInput.h
```diff
@@ -27,8 +36,22 @@ class ModuleInput : public Module
...
+
+	// This setter methods will override user input
+	// Use them with care
+	void SetKey(SDL_Scancode scanCode, KeyState newState);
+	void SetMouseButton(Uint8 mouseButtonCode, KeyState newState);
+
+	SDL_GameController* FindController();
+	SDL_JoystickID GetControllerInstanceID(SDL_GameController* controller) const;
+
+	std::pair<Direction, Direction> GetDirection() const;
```
gerard-queralt, line 54: If you don't mind, I think your solution can be improved. I'll list the specific things I don't like, and as always feel free to discuss and disagree.
- Both `Direction` and `GetDirection` feel too generic to me. I'd somehow specify we're referring to joystic direction.
- The `enum class` (rather, its usage) feels off to me. On paper the values of the enum make sense, but on practice for the first element of the pair we only expect `RIGHT` or `LEFT`, and `FORWARD` or `BACK` for the second one. In other words, we have basically invalid values for both elements, and the user has no way of knowing that.
- In a similar note as the last point, the use of a `pair` here requires the user to know which value represents which direction. I think we could let them know just by the signature.

I think we should, at least, create two different enums, one for horizontal and one for vertical. Personally, I'd also replace the `pair` with a custom struct, something like:
```c++
struct JoystickMovement
{
    JoystickHorizontalDirection horizontalMovement;
    JoystickVerticalDirection verticalMovement;
};
```
But obviously that's just personal preference and doesn't impact the solution in any way, so feel free to choose the one you like the most 😉

MarcGonzalezMoratona: I was thinking that maybe Direction and GetDirection could be used in more contexts than just the joystick input. Probably this could be considered for keyboard input.

I will create two enum classes, one for the X axis movement and another for the Z axis. Finding a great name for this enums is also something to consider. Maybe we can call them HorizontalDirection and VerticalDirection (although vertical could be more related to the Y axis than the Z one).

Using a pair was my first approach to make it work, but I agree that a custom struct is more specific and usable for the developer experience

Source/Modules/ModuleInput.cpp
gerard-queralt: Ok, summary of my findings:
- The buttons work correctly, as expected
- Order of update (should be independent, but anyways): 1) `Input` 2) `ComboManager` 3) `PlayerInput`

So, what ends up happening is the following:
1. `ModuleInput` sets the gamepad button to `down`
2. `ComboManager` reads the state of the mouse buttons, which are both `idle`
3. `ComponentPlayerInput` sets the state of the mouse button to `down`
4. `ModuleInput` sees the state is `down`, and automatically sets it to `repeat`. Relevant code:
```c++
if (mouseButtonState[i] == KeyState::DOWN)
{
	mouseButtonState[i] = KeyState::REPEAT;
}
```
5. `ComboManager` reads the state of the mouse buttons, one being `idle` and the other being `repeat`
6. `ComponentPlayerInput` updates.........

Not sure if thre's an alternative to your solution... If we made the mapping a global thing instead of a component, we would force to always support a controller, which I don't particularly like but it's not the end of the world; however, my biggest concern is that we don't have a UI to modify global elements of the scene (like the Skybox, for instance), so we'd need to develop one.

Bottom line is: let's roll with your solution to get the mapping going, and let's try to keep thinking of alternatives. To me, it feels like we're modifying the `ModuleInput` because of a limitation of the `ComponentPlayerInput`, and that irks me. But it might be that this Component has exposed a limitation of the Module. I tend to put too much trust in existing code 😅 

Would love to hear your thoughts (or anyones) on this conclusion.

Source/Modules/ModuleInput.h
gerard-queralt: > I was thinking that maybe Direction and GetDirection could be used in more contexts than just the joystick input. Probably this could be considered for keyboard input. I will change the name, but leave the door open to reuse it for other purposes

That would be ideal, but I don't think we can do it with our current input system. We'd need to let the user define which button implies which direction, and at that point it would probably be better to revamp our input system API entirely and use messages instead, a la Unity. But I agree that it's worth keeping in mind, maybe we can do something more simple.

gerard-queralt:
Overall the controls feel great to me. You've already mentioned to me that the jump feels acts weird at times, like double jumping immediately. Another thing I noticed is that, by spamming the jump, I can get Bix stuck in the air. Not sure of an exact repro steps, or even if it's an existing bug with the jumping script (I assume it's not but you never know). Here's a video of the issue:

https://github.com/Horizons-Games/Axolotl-Engine/assets/79362519/5a392887-1f8e-48d7-aabb-ab472bb3170e

Another thing I noticed is that the `ComponentPlayerInput` breaks the keyboard input, since it's setting the state of the keys to the state of the ones in the gamepad, meaning it'll always set them to idle. Updating the code in the component to not set the state to idle (which is done automatically by the module) fixes it:
```c++
	for (const auto& [gamepadButton, keyboardButton] : defaultGamepadMapping)
	{
		KeyState gamepadButtonState = input->GetGamepadButton(gamepadButton);
		if (gamepadButtonState == KeyState::IDLE)
		{
			continue;
		}
		if (std::holds_alternative<SDL_Scancode>(keyboardButton))
		{
			input->SetKey(std::get<SDL_Scancode>(keyboardButton), gamepadButtonState);
		}
		else if (std::holds_alternative<Uint8>(keyboardButton))
		{
			Uint8 mouseButton = std::get<Uint8>(keyboardButton);
			input->SetMouseButton(mouseButton, gamepadButtonState);
		}
	}
```

MarcGonzalezMoratona:
> Overall the controls feel great to me. You've already mentioned to me that the jump acts weird at times, like double jumping immediately. Another thing I noticed is that, by spamming the jump, I can get Bix stuck in the air. Not sure of an exact repro steps, or even if it's an existing bug with the jumping script (I assume it's not but you never know). Here's a video of the issue:
> 
>  40c1877c78af506dcc98f2c379d35d50.mp4 
> Another thing I noticed is that the `ComponentPlayerInput` breaks the keyboard input, since it's setting the state of the keys to the state of the ones in the gamepad, meaning it'll always set them to idle. Updating the code in the component to not set the state to idle (which is done automatically by the module) fixes it:
> 
> ```c++
> 	for (const auto& [gamepadButton, keyboardButton] : defaultGamepadMapping)
> 	{
> 		KeyState gamepadButtonState = input->GetGamepadButton(gamepadButton);
> 		if (gamepadButtonState == KeyState::IDLE)
> 		{
> 			continue;
> 		}
> 		if (std::holds_alternative<SDL_Scancode>(keyboardButton))
> 		{
> 			input->SetKey(std::get<SDL_Scancode>(keyboardButton), gamepadButtonState);
> 		}
> 		else if (std::holds_alternative<Uint8>(keyboardButton))
> 		{
> 			Uint8 mouseButton = std::get<Uint8>(keyboardButton);
> 			input->SetMouseButton(mouseButton, gamepadButtonState);
> 		}
> 	}
> ```

I also noticed the the issue with the jump, but I think this kind of problems are being solved with the state machine. I hope this will be solved after merging some other stuff.

Source/Modules/ModuleInput.cpp
MarcGonzalezMoratona: Update: I had to remove this solution because it broke some jump states. Right now we don't have a way to attack just once per input in the button using the controller. I'll work on this and update you soon

MarcGonzalezMoratona: Looks like the problem is partially solved after merging changes from the game-programming branch

AuxPort66:
Could we map the rest of the buttons to already have it done?

Is missing:
- The tab to activate the combo.
- The Z for power ups
- The esc for the pause.

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -79,43 +78,41 @@ void PlayerMoveScript::Move(float deltaTime)
...
-	// Left
-	if (input->GetKey(SDL_SCANCODE_A) != KeyState::IDLE)
-	{
-		totalDirection += -cameraFrustum.WorldRight().Normalized();
-		currentMovements |= MovementFlag::A_DOWN;
-	}
-
 	// Back
-	if (input->GetKey(SDL_SCANCODE_S) != KeyState::IDLE)
+	if (input->GetKey(SDL_SCANCODE_S) != KeyState::IDLE || input->GetDirection().verticalMovement == JoystickVerticalDirection::BACK)
```
markmeliaa, line 89: Line length over 120 characters.

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -79,43 +78,41 @@ void PlayerMoveScript::Move(float deltaTime)
...
 	{
 		totalDirection += -cameraFrustum.Front().Normalized();
 		currentMovements |= MovementFlag::S_DOWN;
 	}
 
 	// Right
-	if (input->GetKey(SDL_SCANCODE_D) != KeyState::IDLE)
+	if (input->GetKey(SDL_SCANCODE_D) != KeyState::IDLE || input->GetDirection().horizontalMovement == JoystickHorizontalDirection::RIGHT)
```
markmeliaa, line 96: Line length over 120 characters.

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -79,43 +78,41 @@ void PlayerMoveScript::Move(float deltaTime)
...
 	{
 		totalDirection += cameraFrustum.WorldRight().Normalized();
 		currentMovements |= MovementFlag::D_DOWN;
 	}
 
+	// Left
+	if (input->GetKey(SDL_SCANCODE_A) != KeyState::IDLE || input->GetDirection().horizontalMovement == JoystickHorizontalDirection::LEFT)
```
markmeliaa, line 103: Line length over 120 characters.

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -79,43 +78,41 @@ void PlayerMoveScript::Move(float deltaTime)
...
-	if (input->GetKey(SDL_SCANCODE_LSHIFT) != KeyState::IDLE)
-	{
-		newSpeed *= 2;
-		shiftPressed = true;
-	}
-	*/
-
 	// Forward
-	if (input->GetKey(SDL_SCANCODE_W) != KeyState::IDLE)
+	if (input->GetKey(SDL_SCANCODE_W) != KeyState::IDLE || input->GetDirection().verticalMovement == JoystickVerticalDirection::FORWARD)
```
markmeliaa, line 82: Line length over 120 characters.

markmeliaa (commented):
I forgot my controller at home so I can't really try the functionality of this pr, but the code looks good, just commented a couple things.
The ImGui::Separator of the `ComponentPlayerInputController` is not being drown correctly in the UI, could you take a look at it please?
Also, this pr should be directed to `develop` it's not really related just with gameplay, but with the game in general, and that way UI can have it too.

MarcGonzalezMoratona:
> Could we map the rest of the buttons to already have it done?
> 
> Is missing:
> 
> * The tab to activate the combo.
> * The Z for power ups
> * The esc for the pause.

I see in this branch pause is done with SDL_SCANCODE_M, so I change it to ESCAPE. I will map the rest of buttons to the controller. Thank you for the feedback!

MarcGonzalezMoratona:
> I forgot my controller at home so I can't really try the functionality of this pr, but the code looks good, just commented a couple things. The ImGui::Separator of the `ComponentPlayerInputController` is not being drown correctly in the UI, could you take a look at it please? Also, this pr should be directed to `develop` it's not really related just with gameplay, but with the game in general, and that way UI can have it too.

Hi Mark! I solved everything you asked in the comments. Please, take a look when you are free. Thank you for the feedback!

Vlukkee:
I was testing the new inputs.
ESC works well, although you have to tweak the entry to the menu since it activates the mouse, but it is a UI issue.
Z - It is not working for me even though I have the power up
TAB - The same does not activate the combo system

markmeliaa (dismissed):
Code looks good and I trust in the reviewers that everything works correct.

AuxPort66 (approved):

markmeliaa (approved):

PR #630, opened 2023-08-04
Fix: Enemy Detection

AuxPort66:
- Fixed issue where enemy detection was working erratically. The dot product check that verified that the enemy was in front of any offset has been deleted. Now it is assumed that if the offset is very large, any enemy within that area can be attacked.

- Added a function to prioritize targets with the "PriorityTarget" tag.

- Added theshold in the angle comparison to decide if one enemy is in front of another. With this we have better detection of which enemies are in the middle.

- Added distance threshold for attacks at different ranges (it is drawn as a dark red circle on the ground), keep in mind that the comparison to know if it's in the distance is with the center of the enemy, so you have to be a little generous with the distance. If there is no distance threshold (0), the maximum range of the detection dome is used.

- BixAttack is in charge of calling the entity update in order to always have the selected enemy updated before attacking, it also informs entity detection of possible attacks the user can use in order to change the detection range when it's need it.

- Added new target particle that activates and deactivates depending on the enemy selected.

Bug:
- The target particle has little lifespan to correctly follow the movement of the enemy, but that makes it flicker. I have already asked graphics team for a solution.


diegusmejia117 (approved):

Game/Scripts/EntityDetection.cpp
```diff
@@ -72,13 +76,26 @@ void EntityDetection::DrawDetectionLines()
...
+		if (distanceFilter != 0)
+		{
+			float3 enemyPosition = enemy->GetGlobalPosition();
+			enemyPosition.y = originPosition.y;
+			insideDistanceFilter = originPosition.Distance(enemyPosition) <= distanceFilter;
+		}
+
+		bool equalPriorityLevel = !actualIsSpecialTarget || enemy->GetOwner()->GetTag() == "PriorityTarget";
+
+		if (!enemy->GetOwner()->GetComponent<HealthSystem>()->EntityIsAlive() || !equalPriorityLevel || !insideDistanceFilter)
```
markmeliaa, line 98: Brackets for this if

Game/Scripts/EntityDetection.cpp
```diff
@@ -125,11 +144,22 @@ void EntityDetection::SelectEnemy()
 		dd::arrow(originPosition, enemySelected->GetGlobalPosition(), dd::colors::Red, 0.1f);
 	}
 #endif // ENGINE
+
+	if (lastenemySelected != enemySelected) 
+	{
+		if (lastenemySelected != nullptr) VisualParticle(false, lastenemySelected->GetOwner());
```
markmeliaa, line 150: Brackets here

Game/Scripts/EntityDetection.cpp
```diff
@@ -125,11 +144,22 @@ void EntityDetection::SelectEnemy()
...
+		if (enemySelected != nullptr) VisualParticle(true, enemySelected->GetOwner());
```
markmeliaa, line 151: And here too

markmeliaa (approved):

PR #632, opened 2023-08-04
feat: merge together all the combat

markmeliaa:
In this pr, all the behaviour for the combat system that was developed during these past weeks is combined and fully working, you can try it in the `Gameplay_FullTrial_VS3` scene (it will be renamed after this release, I promise).

How does the combat work? 
- Accumulate combo by using the light attack (`LEFT MOUSE BUTTON`), the heavy attack (`RIGHT MOUSE BUTTON`) and the jump attack (`LEFT MOUSE BUTTON` or `RIGHT MOUSE BUTTON` after jumping in the air). You can see them chaining below the combo bar (the jump attack does not appear in the combo bar yet).
- Once you have the combo bar fully charged (green), you can press `TAB` to activate the special movements (the bar will turn yellow).
- When the special movements are active, any combo you perform will end with a light, heavy or jumping finisher, depending on which the last input of the combo was (e.g. last button was `LEFT MOUSE BUTTON`? you will perform a LIGHT FINISHER).
(For now, the normal jump attack and the finisher jump attack look the same, but the finisher has more damage and more pushing force).

Aside from that:
- Animations were added where they were missing (dash, basic attacks, finishers...).
- Particles were added to the finishers so they look better.
- Other minor tasks.


Game/Scripts/BixAttackScript.cpp
```diff
@@ -243,6 +212,72 @@ void BixAttackScript::HeavyFinisher()
...
+			}
+			break;	
+
+		case AttackType::HEAVYNORMAL:
+			if (!animation->isPlaying())
+			{
+				animation->SetParameter("IsAttacking", false);
+				isAttacking = false;
+			}
+			break;
```
gerard-queralt, line 246: Since these two cases are the same, we could merge them.
```suggestion
		case AttackType::LIGHTNORMAL:
		case AttackType::HEAVYNORMAL:
			if (!animation->isPlaying())
			{
				animation->SetParameter("IsAttacking", false);
				isAttacking = false;
			}
			break;	
```

Game/Scripts/BixAttackScript.cpp
```diff
@@ -243,6 +212,72 @@ void BixAttackScript::HeavyFinisher()
...
+			if (!heavyFinisherAttack->IsAttacking()) // Heavy Finisher attack has finished
+			{
+				animation->SetParameter("HeavyFinisherInit", false);
+				animation->SetParameter("HeavyFinisherExit", true);
+				bixLightSaber->Enable();
+				isAttacking = false;
+			}
+			break;
+
+		case AttackType::JUMPFINISHER:
```
gerard-queralt, line 274: We could also merge this one with `AttackType::JUMPNORMAL`

Game/Scripts/ComboManager.cpp
```diff
@@ -12,14 +12,14 @@ REGISTERCLASS(ComboManager);
...
+	comboCount(0.0f), 
+	maxComboCount(3.0f),
 	uiComboManager(nullptr),
 	comboManagerUIReference(nullptr),
-	comboTime(10),
-	actualComboTimer(0)
+	comboTime(10.0f),
+	actualComboTimer(0.0f)
 {
 	REGISTER_FIELD(comboManagerUIReference, GameObject*);
```
gerard-queralt, line 24: We could change the type of this field.
```suggestion
	REGISTER_FIELD(uiComboManager, UIComboManager*);
```

Game/Scripts/ComboManager.cpp
```diff
@@ -29,20 +29,16 @@ ComboManager::ComboManager() : Script(),
...
 	}
 }
 
-void ComboManager::Update(float deltaTime)
-{
-	
-}
-
-int ComboManager::GetcomboCount() const
+float ComboManager::GetcomboCount() const
```
gerard-queralt, line 41:
```suggestion
float ComboManager::GetComboCount() const
```

Game/Scripts/ComboManager.h
```diff
@@ -24,30 +25,29 @@ class ComboManager : public Script
...
-
 	void ClearCombo(bool finisher);
 
 	GameObject* comboManagerUIReference;
 	UIComboManager* uiComboManager;
 
 	ModuleInput* input;
 	bool specialActivated;
-	int specialCount;
+	float specialCount;
```
gerard-queralt, line 47: Out of curiosity, why was this changed? "Count", to me, suggests it should be an unsigned integer, if anything.

Game/Scripts/LightAttackBullet.cpp
```diff
@@ -61,13 +64,18 @@ void LightAttackBullet::SetEnemy(GameObject* nEnemy)
 
 void LightAttackBullet::OnCollisionEnter(ComponentRigidBody* other)
 {
-	if (other->GetOwner() == enemy)
+	if (other->GetOwner()->CompareTag("Enemy"))
```
gerard-queralt, line 67: This is checking something different than it was before. I don't know the context of this script so it's very likely it's the same thing, but this to me reads like: "If we collide with any enemy, doesn't matter which, we damage the one we have assigned". Again, it's probably correct in the context of the script, but pointing it out just in case.

Game/Scripts/BixAttackScript.cpp
markmeliaa: They are now, but they won't be.

Game/Scripts/BixAttackScript.cpp
markmeliaa: They would likely have different animation/sound effects to be stopped, so for now they are good separated.

Game/Scripts/ComboManager.h
markmeliaa: Changed every `int` in this script to `float` because I thought they were used as `REGISTER_FIELDS`, turns out there were some that could remain as `int`s. Ty for pointing it out.

Game/Scripts/LightAttackBullet.cpp
markmeliaa: This attack has an objective (the variable `enemy`), but this was changed so in case any other enemy is hit, regardless if it is the selected enemy or another one, the attack will damage the enemy that was hit first.

Game/Scripts/BixAttackScript.cpp
markmeliaa: They aren't anymore.

Game/Scripts/DebugGame.cpp
```diff
@@ -209,7 +209,7 @@ void DebugGame::BeImmortal() const
 
 void DebugGame::DeathTouch() const
 {
-	if (!playerAttackScript->IsDeathTouched())
+	if (!playerAttackScript->GetIsDeathTouched())
```
gerard-queralt, line 212: I like this better the way it was, it's generally the preferred way of writing getters for booleans.

Game/Scripts/ComboManager.cpp
```diff
@@ -122,14 +131,21 @@ AttackType ComboManager::CheckAttackInput(bool jumping)
 
 void ComboManager::SuccessfulAttack(float specialCount, AttackType type)
 {
-	if(specialCount < 0 || !specialActivated)
+	if (specialCount < 0 || !specialActivated)
 	{
-		this->specialCount = std::max(0.0f, std::min(this->specialCount + specialCount, maxSpecialCount));
+		this->specialCount = static_cast<int>(
+			std::max(0.0f, std::min(static_cast<float>(this->specialCount) + specialCount, maxSpecialCount)));
```
gerard-queralt, line 137: For this, you could use [`std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp). It basically does the same, but is less verbose. We unfortunately still need a cast, though.
```suggestion
		this->specialCount = std::clamp(this->specialCount + specialCount, 0, static_cast<int>(maxSpecialCount));
```

Game/Scripts/ComboManager.cpp
```diff
@@ -122,14 +131,21 @@ AttackType ComboManager::CheckAttackInput(bool jumping)
 
 void ComboManager::SuccessfulAttack(float specialCount, AttackType type)
```
gerard-queralt, line 132: If the member `specialCount` is an int, why is this parameter a float? Seems like wherever it's calles it actually receives an int, too.

Game/Scripts/JumpFinisherArea.cpp
```diff
@@ -27,11 +31,35 @@ void JumpFinisherArea::Start()
...
+	if (particleSystemCurrentTimer <= 0.0f)
+	{
+		particleSystemCurrentTimer = particleSystemTimer;
+		triggerParticleSystemTimer = false;
+		particleSystem->Stop();
+
+		if (throwableForceArea)
+		{
+			// If the force area is from a bullet, destroy the area after playing the particle effects
+			App->GetModule<ModuleScene>()->GetLoadedScene()->RemoveParticleSystem(particleSystem);
```
gerard-queralt, line 59: Is this not done automatically? 😕

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -157,21 +157,12 @@ void PlayerMoveScript::Move(float deltaTime)
...
-			//if (shiftPressed)
-			//{
-			//	movement /= 2;
-			//}
-			Dash();
 		}
 
-		canDash = false;
-		nextDash = 3000 + static_cast<float>(SDL_GetTicks());
+		nextDash = 1000.0f + static_cast<float>(SDL_GetTicks());
```
gerard-queralt, line 162: Does this work? As per [the documentation](https://wiki.libsdl.org/SDL2/SDL_GetTicks):
> Get the number of milliseconds since SDL library initialization.

Which is basically the number of milliseconds since the engine started running. I'm almost positive you should use the `deltaTime` for this:
```suggestion
		nextDash = 1000.0f + deltaTime;
```

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -190,6 +181,20 @@ void PlayerMoveScript::Move(float deltaTime)
...
+	// Turn off dash animation correctly
+	if (componentAnimation->GetActualStateName() == "BixDashingInit" ||
+		componentAnimation->GetActualStateName() == "BixDashingKeep" ||
+		componentAnimation->GetActualStateName() == "BixDashingEnd")
+	{
+		componentAnimation->SetParameter("IsDashing", false);
+	}
+
 	// Cooldown Dash
 	if (!canDash && nextDash < SDL_GetTicks())
```
gerard-queralt, line 199: Same as the previous comment.

gerard-queralt (commented):
Not done being annoying yet 😛

Game/Scripts/ComboManager.cpp
```diff
@@ -13,36 +13,27 @@ ComboManager::ComboManager() : Script(),
 	input(nullptr), 
 	specialActivated(false), 
 	specialCount(0),
-	maxSpecialCount(100),
+	maxSpecialCount(100.0f),
```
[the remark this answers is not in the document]
markmeliaa: `maxComboCount` needs to be a `float` because it is set as a `REGISTER_FIELD`, but I reverted `maxSpecialCount` to an `int`

Game/Scripts/ComboManager.cpp
```diff
@@ -68,12 +64,18 @@ void ComboManager::CheckSpecial(float deltaTime)
 			ClearCombo(false);
 			actualComboTimer = comboTime;
 		}
+
 		else if (specialCount > 0 && specialCount < maxSpecialCount)
 		{
-			specialCount = std::max(0.0f, specialCount - (5 * deltaTime));
-			if (uiComboManager) uiComboManager->SetComboBarValue(specialCount);
+			specialCount = static_cast<int>(std::max(0.0f, static_cast<float>(specialCount) - (5.0f * deltaTime)));
```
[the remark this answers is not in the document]
markmeliaa: `std::max` seems to not work with an `int` and a `float`, so I'll just cast deltaTime into an `int`.

image

Game/Scripts/ComboManager.cpp
markmeliaa: We need another additional cast because `specialCount` (the parameter of the funcion) is a float, so we need to cast that back to an int to match the three other ints (`std::clamp` does not work with `float`, `int`, `int`, it raises an error such as with std::max).

Game/Scripts/PlayerMoveScript.cpp
markmeliaa: I've been trying around because simply changing the `SDL_GetTicks()` to `deltaTime` does not work. What is being done with `SDL_GetTicks()` is, as you said, get the miliseconds since SDL library initialization, so in here, `nextDash` will always be maked 1000 miliseconds more than the current SDL miliseconds, and when SDL miliseconds reaches that mark, the line that you commented in the next comment gets triggered. I'm gonna change it so it works with `deltaTime` instead of `SDL_GetTicks()` because it seems more intuitive and we are already using `deltaTime` in other scripts rather than `SDL_GetTicks()`.

Game/Scripts/JumpFinisherArea.cpp
markmeliaa: Sadly, nop. It took me a long time to realize that I had to add that line because, if not added, a particle system is never deleted from the vector of particle systems of the scene (even after destroying the GameObject who owned the particle system), so I had to add this line before deleting a gameobject to prevent the engine from crashing.

Game/Scripts/DebugGame.cpp
markmeliaa: Mb, I think I messed up in a merge and this got changed back to how it was. I restored it now.

Game/Scripts/ComboManager.cpp
markmeliaa: There were two calls which passed a `float`, but I just casted them because it's true that the other calls are all passing `int`s.

Game/Scripts/JumpFinisherArea.cpp
gerard-queralt: Noted, will try to make it automatically, should be easy enough. Sorry for the inconvenience!

Game/Scripts/ComboManager.cpp
```diff
@@ -13,36 +13,27 @@ ComboManager::ComboManager() : Script(),
...
-		uiComboManager = comboManagerUIReference->GetComponent<UIComboManager>();
-		maxSpecialCount = uiComboManager->GetMaxComboBarValue();
-		uiComboManager->SetComboBarValue(specialCount);
-	}
-}
 
-void ComboManager::Update(float deltaTime)
-{
-	
+	maxSpecialCount = static_cast<float>(uiComboManager->GetMaxComboBarValue());
```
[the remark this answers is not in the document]
markmeliaa: Oh, I deleted the check here by mistake, mb. I will just bring back the check because the whole functionality of the script was not performed by me so, if there are checks it must be for something.

Game/Scripts/JumpFinisherArea.cpp
markmeliaa: No problem, for now it only happens here and with the Light Finisher Bullet, because they are objects that, on contact, they show some particles and then get destroyed, so I'd say that this is the first time a particle system is being deleted.

Game/Scripts/ComboManager.cpp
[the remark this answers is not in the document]
AuxPort66: MaxSpecialCount is a float, slider values are floats.
SpecialCount too. It's not the same as maxComboCount

Game/Scripts/ComboManager.h
AuxPort66: > Out of curiosity, why was this changed? "Count", to me, suggests it should be an unsigned integer, if anything.

Is connected to the actualValue of a slider, slider use float values.

Game/Scripts/UIComboManager.cpp
```diff
@@ -63,7 +63,7 @@ void UIComboManager::SetActivateSpecial(bool activate)
 
 void UIComboManager::SetComboBarValue(float value)
 {
-	comboBar->ModifyCurrentValue(value);
+	comboBar->ModifyCurrentValue(static_cast<float>(value));
```
AuxPort66, line 66: Cast forgotten in the last refactor

markmeliaa: Ty for pointing it out!

gerard-queralt (approved):
All the code looks great. Thank you very much to you and the rest of the gameplay team for the awesome work!

AuxPort66 (approved):

Game/Scripts/JumpFinisherArea.cpp
MacXxs: This is already fixed in the Graphics/Particles-optimization branch. When the developments in Graphics-main are merged to develop it will be solved.

Game/Scripts/EntityDetection.cpp
```diff
@@ -41,10 +41,13 @@ void EntityDetection::Start()
 	rigidBody->SetKpForce(50);
 }
 
-void EntityDetection::UpdateEnemyDetection(float distanceFilter)
+void EntityDetection::Update(float deltaTime)
```
UriMurcia, line 44: is this update separated from the other useful? Can't see for what we are using it

Game/Scripts/HealthSystem.cpp
```diff
@@ -78,7 +83,7 @@ void HealthSystem::TakeDamage(float damage)
 {
 	if (!isImmortal) 
 	{
-		if (owner->CompareTag("Player"))
+		if (owner->CompareTag("Player") && !attackScript->IsPerfomingJumpAttack())
```
UriMurcia, line 86: Is having a reference to the attack script in the health system necessary? For what is this line?

UriMurcia (commented):
Hey great job Marc! Just two things to make sure

Game/Scripts/JumpFinisherArea.cpp
markmeliaa: Nice!

Game/Scripts/EntityDetection.cpp
markmeliaa: The light finisher uses its own enemy detection because it needs more range, so I had to take that line and move it to the Update. I already talked with Sara about this and I will see with David Castro what can we do with the light finisher detection in the future.

Game/Scripts/HealthSystem.cpp
markmeliaa: I was asked that, when performing the jump finisher, Bix should take no damage, so I had to add this in the health system (although I know it looks wrong to have it here because the `HealthSystem` should only have healt system related stuff, this was the only solution that I could think of).

PR #634, opened 2023-08-07
fix[WIP]: diagonal dash

MarcGonzalezMoratona:
Diagonal dash has now a similar force than other directions.

We are still facing the issue that the player does not animate when walking after a dash, but it's a previous bug, so don't consider this problem for approving the current PR

### DIFF
--- a/Game/Scripts/PlayerMoveScript.cpp
+++ b/Game/Scripts/PlayerMoveScript.cpp
@@ -276,10 +276,12 @@ void PlayerMoveScript::Dash()
 		dashImpulse.x = -dashForce;
 	}
 
-	if (dashDirection.z > 0.5f) {
+	if (dashDirection.z > 0.5f)
+	{
 		dashImpulse.z = dashForce;
 	}
-	else if (dashDirection.z < -0.5f) {
+	else if (dashDirection.z < -0.5f)
+	{
 		dashImpulse.z = -dashForce;
 	}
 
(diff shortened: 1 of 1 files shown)


MarcGonzalezMoratona:
Looks good to me. Good job @Piterga12. I can't approve it because I created the PR, so you can approve it in my name.

Piterga12 (approved):

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -260,12 +260,30 @@ void PlayerMoveScript::Dash()
...
+	}
 
-	float3 dashImpulse = dashDirection.Normalized() * dashForce;
-	dashImpulse.Normalized();
+	if (dashDirection.z > 0.5f) {
+		dashImpulse.z = dashForce;
+	}
+	else if (dashDirection.z < -0.5f) {
+		dashImpulse.z = -dashForce;
+	}
```
gerard-queralt, line 284:
```suggestion
	if (dashDirection.z > 0.5f)
	{
		dashImpulse.z = dashForce;
	}
	else if (dashDirection.z < -0.5f)
	{
		dashImpulse.z = -dashForce;
	}
```
😇

markmeliaa:
> We are still facing the issue that the player does not animate when walking after a dash, but it's a previous bug, so don't consider this problem for approving the current PR

This is already solved by #632, so nothing to worry about.

markmeliaa (approved):
Looks amazing guys, good job!

PR #635, opened 2023-08-07
fix[WindowComponentMeshRenderer]: remove call to BeginChild

gerard-queralt:
As per `ImGui::BeginChild` documentation (skipping to the relevant part):

> For each independent axis of 'size': ==0.0f: use remaining host window size / >0.0f: fixed size / <0.0f: use remaining window size minus abs(size) / Each axis can use a different mode, e.g. ImVec2(0,400).

Looks like this was the reason the window took the remaining space of the parent window.

### DIFF
--- a/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentMeshRenderer.cpp
+++ b/Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentMeshRenderer.cpp
@@ -66,15 +66,13 @@ void WindowComponentMeshRenderer::SetMaterial(const std::shared_ptr<ResourceMate
 
 void WindowComponentMeshRenderer::DrawWindowContents()
 {
-	if (changed)
-	{
-		ImGui::PushStyleColor(ImGuiCol_ChildBg, IM_COL32(75, 25, 25, 255));
-	}
-	else
-	{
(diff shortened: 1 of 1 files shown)


markmeliaa (dismissed):
LGTM!

AuxPort66 (changes requested):
If you remove the beginChild the material doesn't put to red when you change things so Level and art don't have any visual input to know they didn't save the edited material. Don't erase that functionality. And ask me if you didn't know why something is there pls.

gerard-queralt:
> If you remove the beginChild the material doesn't put to red when you change things so Level and art don't have any visual input to know they didn't save the edited material. Don't erase that functionality.

Could you be more specific on what's the functionality so I know how to test it? I was hoping `PushID` would achieve the same thing, but I guess you've tested it and it doesn't.

> And ask me if you didn't know why something is there pls.

Already did, pinged you in the engine channel.

gerard-queralt:
Nevermind, got it. But the need to use `BeginChild` only comes from using the flag `ImGuiCol_ChildBg` to set the background.

gerard-queralt:
@AuxPort66 should be fixed, check it when you can

AuxPort66:
> Already did, pinged you in the engine channel.
Before the PR, about the issue.

I check it. Everything seems correct. Nice.

AuxPort66 (approved):

markmeliaa (approved):

PR #637, opened 2023-08-08
Feat: Attack Movement Improvement

UriMurcia:
Hi! I made this changes because I felt that the attack was a little bit off. I made it without consulting so please, any change or suggestion is welcomed.

This PR makes the character look at the enemy targeted while attacking. Also, slows down the velocity movement while attacking, that's to gain more control of the character and because it felt weird moving the character while the attack animation was still.

### DIFF
--- a/Game/Scripts/BixAttackScript.cpp
+++ b/Game/Scripts/BixAttackScript.cpp
@@ -39,8 +39,7 @@ BixAttackScript::BixAttackScript() : Script(),
 	isAttacking(false), attackCooldown(0.6f), attackCooldownCounter(0.f), audioSource(nullptr),
 	animation(nullptr), transform(nullptr),
 	playerManager(nullptr), attackComboPhase(AttackCombo::IDLE), enemyDetection(nullptr), jumpFinisherScript(nullptr),
-	lightFinisherScript(nullptr), normalAttackDistance(0), heavyFinisherAttack(nullptr), bixLightSaber(nullptr),
-	isJumpAttacking(false)
+	lightFinisherScript(nullptr), normalAttackDistance(0), heavyFinisherAttack(nullptr), bixLightSaber(nullptr)
 {
 	//REGISTER_FIELD(comboInitTimer, float);
 	REGISTER_FIELD(comboCountHeavy, float);
@@ -157,7 +156,7 @@ void BixAttackScript::LightNormalAttack()
 	if(enemyAttacked != nullptr)
 	{
 		LOG_VERBOSE("Enemy hit with light attack");
-		comboSystem->SuccessfulAttack(comboCountSoft, AttackType::LIGHTNORMAL);
+		comboSystem->SuccessfulAttack(comboCountSoft, currentAttack);
 		DamageEnemy(enemyAttacked, attackSoft);
 	}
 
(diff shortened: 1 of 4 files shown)


Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -140,12 +147,27 @@ void PlayerMoveScript::Move(float deltaTime)
 			SetPlayerState(PlayerActions::WALKING);
 		}
 
+		//Low velocity while attacking
+		AttackType currentAttack = bixAttackScript->GetCurrentAttackType();
+		switch (currentAttack)
+		{
+		case AttackType::LIGHTNORMAL:
+				newSpeed = newSpeed / 2.0f;
```
gerard-queralt, line 155: Instead of adjusting the `newSpeed` here, maybe we could return just the factor. My reasoning being, if the computation somewhy needs to change, we wouldn't need to change each case block.

Game/Scripts/PlayerMoveScript.cpp
```diff
@@ -205,15 +227,40 @@ void PlayerMoveScript::Move(float deltaTime)
...
+		{
+		case AttackType::LIGHTNORMAL:
+			desiredRotation = vecForward + vecTowardsEnemy;
+			break;
+		case AttackType::HEAVYNORMAL:
+			desiredRotation = vecForward + vecTowardsEnemy;
+			break;
+		case AttackType::LIGHTFINISHER:
+			desiredRotation = vecForward + vecTowardsEnemy;
+			break;
```
gerard-queralt, line 255: If these three are indeed intended to be the same, we could combine the cases.
```suggestion
		case AttackType::LIGHTNORMAL:
		case AttackType::HEAVYNORMAL:
		case AttackType::LIGHTFINISHER:
			desiredRotation = vecForward + vecTowardsEnemy;
			break;
```
Also, I don't know the `AttackType` enum off the top of my head, but do we need a `default` case here?

Game/Scripts/BixAttackScript.cpp
```diff
@@ -340,3 +335,13 @@ void BixAttackScript::SetIsDeathTouched(bool isDeathTouched)
 {
 	this->isDeathTouched = isDeathTouched;
 }
+
+AttackType BixAttackScript::GetCurrentAttackType() const
+{
+	return currentAttack;
+}
+
+GameObject* BixAttackScript::GetEnemyDetection() const
```
markmeliaa, line 344: Instead of `GetEnemyDetection` I'd name this as `GetEnemyDetected` or something like this, because rather than getting the `enemyDetection` object itself you are getting the selected enemy of that `enemyDetection`.

markmeliaa (approved):
Looks good Uri, gj!

As `game-programming` was early merged into `develop` you could direct this pr to `develop` directly so we would have this improvement in this release.

Game/Scripts/PlayerMoveScript.cpp
UriMurcia: Thanks Gerard! I changed bot things :)

Game/Scripts/BixAttackScript.cpp
UriMurcia: True! Better to understand, thanks Marc! :)

markmeliaa (approved):

PR #647, opened 2023-08-16
feat[animatedTextures]: Added percentage selection in texture to tiling

MacXxs:
Now the user can select the % of the texture that wants to use.

Examples: 
|Image|Tiling Config|
|:----:|:----:|
|image|image|
|image|image|


Source/DataModels/Components/ComponentDirLight.cpp
```diff
@@ -23,20 +23,24 @@
 
 ComponentDirLight::ComponentDirLight() : ComponentLight(LightType::DIRECTIONAL, false)
 {
+	shadowBias = float2(0.006, 0.0009);
```
PaablooCH, line 26: indicate that they are float to avoid warnings

Source/DataModels/Resources/ResourceMaterial.cpp
```diff
@@ -59,6 +60,8 @@ void ResourceMaterial::SaveLoadOptions(Json& meta)
 	meta["tilingy"] = static_cast<float>(loadOptions.tiling.y);
 	meta["offsetx"] = static_cast<float>(loadOptions.offset.x);
 	meta["offsety"] = static_cast<float>(loadOptions.offset.y);
+	meta["percentagex"] = static_cast<float>(loadOptions.percentage.y);
```
PaablooCH, line 63:
```suggestion
	meta["percentagex"] = static_cast<float>(loadOptions.percentage.x);
```

PaablooCH (changes requested):
2 things and i will aprove it

PaablooCH:
specular frag doesn't count with tiling, have you solve it?

PaablooCH (approved):

PR #651, opened 2023-08-17
fix: from kp to update rigidbodies

markmeliaa:
--Affected objects--
From Bix:
- `Bix Force Area`
- `Allura Force Area`
- `EnemyDetection`

From the drones:
- `ExplosionArea`

These objects' rigidbodies were set to kinematic and they changed from `rigidBody->SetPositionTarget()` to `rigidBody->UpdateRigidbody()` to update their movement at the same time the parent (Bix) moves. It was stated in today's meeting that this is the preferred way to move objects along with their parents.

Also, the scenes that had those objects (the newest Bix prefab or the newest drone prefab) were updated to match this change.

### DIFF
--- a/Game/Assets/Scenes/BixPrefab.axolotl
+++ b/Game/Assets/Scenes/BixPrefab.axolotl
@@ -146,7 +146,13 @@
                             "height": 0.699999988079071,
                             "rbPos_X": 0.0,
                             "rbPos_Y": 0.6499999761581421,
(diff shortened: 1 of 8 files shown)


DADSILVER:
I'm sorry to say that, but drone explosion area have the same problem. Could you change too?

DADSILVER (approved):

AuxPort66 (approved):

PR #653, opened 2023-08-19
refactor[GameObject]: use concepts instead of SFINAE for component methods

gerard-queralt:
This also allows `HasComponent` to work with scripts.

Info regarding SFINAE (Substitution Failure Is Not An Error): https://en.cppreference.com/w/cpp/language/sfinae


markmeliaa (approved):
> This also allows HasComponent to work with scripts.

Fixed it correctly, it works now, ty!

PaablooCH (approved):

PR #654, opened 2023-08-19
feat: Global Format method, Script::Assert

gerard-queralt:
Couple things done:

- All formatting methods that were in `AxoLog` were moved to their own `Formatter` file. I've wanted to do this for a while so they can be used everywhere we want, and I thought this PR was a good opportunity to do so.
- Added a `Script::Assert` method. Gameplay programmer can pass a condition that will be checked that will be checked. If the condition evaluates to false, we'll log an error and stop the execution of the script.

I've extracted the formatting methods now because, on my first approach to the Assert method, I did the following:
```c++
template<typename... Args>
Assert(bool&& condition, std::string&& format, Args&&... args)
{
	if (!condition)
	{
		LOG_ERROR(format, args);
		throw ScriptAssertFailedException("Assert failed during execution of script " +
										  std::string(GetConstructor()->GetName()));
	}
}
```
However, you cannot pass an argument pack to a macro, so the line `LOG_ERROR(format, args)`. In order to allow gameplay programmers to write more descriptive errors and do so with ease, I opted for this refactor.


markmeliaa (approved):
LGTM!

markmeliaa (dismissed):

Piterga12 (approved):
Looks good!

PR #655, opened 2023-08-20
feat[Filesystem]: Video Component for UI Canvas

ericdacal:
Video Component for UI that can be include in Canvas gameobject. Component includes flip and loop.


Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+					avFrame->data[1] += avFrame->linesize[1] * (videoCodecCtx->height / 2 - 1);
+					avFrame->linesize[1] *= -1;
+					avFrame->data[2] += avFrame->linesize[2] * (videoCodecCtx->height / 2 - 1);
+					avFrame->linesize[2] *= -1;
+				}
+
+				break;
+			}
+
+			response = avcodec_send_packet(videoCodecCtx, avPacket);
```
gerard-queralt, line 258: What's the difference between `response` and `error`? They seem to be used very similarly.

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+			response = avcodec_receive_frame(videoCodecCtx, avFrame);
+			if (response == AVERROR(EAGAIN) || response == AVERROR_EOF)
+			{
+				av_packet_unref(avPacket);
+				continue;
+			}
+			if (response < 0)
+			{
+				LOG_ERROR("Failed to decode frame");
+				// LOG_ERROR("Failed to decode frame: %s.", libav_err2str(response));
```
gerard-queralt, line 275: If you want to, you can still add this information in the log.
```suggestion
				LOG_ERROR("Failed to decode frame: {}", libav_err2str(response));
```

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+			avFrame->data[0] += avFrame->linesize[0] * (videoCodecCtx->height - 1);
+			avFrame->linesize[0] *= -1;
+			avFrame->data[1] += avFrame->linesize[1] * (videoCodecCtx->height / 2 - 1);
+			avFrame->linesize[1] *= -1;
+			avFrame->data[2] += avFrame->linesize[2] * (videoCodecCtx->height / 2 - 1);
+			avFrame->linesize[2] *= -1;
+		}
+
+
+		uint8_t* dest[4] = { frameData, nullptr, nullptr, nullptr };
```
gerard-queralt, line 315: Do we need to free this pointer or is it handled by the lib?

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+		int linSize[4] = { frameWidth * 4, 0, 0, 0 };
+		sws_scale(scalerCtx, avFrame->data, avFrame->linesize, 0, frameHeight, dest, linSize);
+	}
+}
+
+void ComponentVideo::SetVideoFrameSize(int width, int height)
+{
+	
+	ComponentTransform2D* transform = GetOwner()->GetComponentInternal<ComponentTransform2D>();
```
gerard-queralt, line 324: Again, I'd check the transform, log if not found.

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+	transform->SetSize(float2((float) width, (float) height));
+}
+
+void ComponentVideo::RestartVideo()
+{
+	avio_seek(formatCtx->pb, 0, SEEK_SET);
+	if (av_seek_frame(formatCtx, videoStreamIndex, -1, 0) >= 0)
```
gerard-queralt, line 331: Empty if block.

Source/DataModels/Components/UI/ComponentVideo.h
```diff
@@ -0,0 +1,102 @@
...
+public:
+	ComponentVideo(bool active, GameObject* owner);
+	~ComponentVideo() override;
+	void Init();
+	void SetLoop(bool loop);
+	bool GetLoop();
+	void SetCanBeRotate(bool canRotate);
+	bool GetCanBeRotate();
+	void SetRotateVertical(bool rotateVertical);
+	bool GetRotateVertical();
```
gerard-queralt, line 23: In general, getters for booleans use verbs like "is" or "can" instead of "get"
```suggestion
	bool IsLooping();
	void SetCanBeRotate(bool canRotate);
	bool CanBeRotate();
	void SetRotateVertical(bool rotateVertical);
	bool CanRotateVertical();
```

Source/DataModels/Components/UI/ComponentVideo.h
```diff
@@ -0,0 +1,102 @@
...
+	void InternalLoad(const Json& meta) override;
+	void OpenVideo(const char* filePath);
+	void RestartVideo();
+	std::shared_ptr<ResourceVideo> video;
+
+	AVFormatContext* formatCtx = nullptr;
+	AVCodecContext* videoCodecCtx = nullptr;
+	AVPacket* avPacket = nullptr;
+	AVFrame* avFrame = nullptr;
+	SwsContext* scalerCtx = nullptr;
```
gerard-queralt, line 41: None of these pointers are freed. I assume this is all handled by the lib, but pointing it out just in case.

Source/DataModels/Components/UI/ComponentVideo.h
```diff
@@ -0,0 +1,102 @@
...
+inline bool ComponentVideo::GetRotateVertical()
+{
+	return rotateVertical;
+}
+
+inline bool ComponentVideo::GetLoop()
+{
+	return this->loop;
+}
+
```
gerard-queralt, line 86: Extra line break :P

Source/DataModels/Components/UI/ComponentVideo.h
```diff
@@ -0,0 +1,102 @@
...
+
+inline void ComponentVideo::SetLoop(bool loop)
+{
+	this->loop = loop;
+}
+
+inline void ComponentVideo::SetVideo(const std::shared_ptr<ResourceVideo>& video)
+{
+	this->video = video;
```
gerard-queralt, line 95: Make sure reference count is increased. We're creating a copy anyway when assign it, so it makes no difference in performance.
```suggestion
inline void ComponentVideo::SetVideo(std::shared_ptr<ResourceVideo> video)
{
	this->video = std::move(video);
```

Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -358,6 +359,12 @@ void GameObject::CopyComponent(Component* component)
 			break;
 		}
 
+		case ComponentType::VIDEO:
+		{
+			newComponent = std::make_unique<ComponentVideo>(*static_cast<ComponentVideo*>(component));
```
gerard-queralt, line 364: Have you checked that a shallow copy is enought? I worry that two components can't share the same `frameData`, especially if we properly free the memory.

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+	rotateVertical = meta["rotateVertical"];
+	canRotate = meta["canBeRotate"];
+	bool resourceExists = !path.empty() && App->GetModule<ModuleFileSystem>()->Exists(path.c_str());
+	if (resourceExists)
+	{
+		std::shared_ptr<ResourceVideo> resourceImage =
+			App->GetModule<ModuleResources>()->RequestResource<ResourceVideo>(path);
+		if (resourceImage)
+		{
+			video = resourceImage;
```
gerard-queralt, line 136: Use move instead of copy, since it's more efficient.
```suggestion
			video = std::move(resourceImage);
```

Source/DataModels/Resources/ResourceVideo.cpp
gerard-queralt, line 1: Since I assume this PR is intended to be a proper first version of this feature, I don't think the Resource should be empty.

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentVideo.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+
+WindowComponentVideo::~WindowComponentVideo()
+{
+}
+
+void WindowComponentVideo::DrawWindowContents()
+{
+	ComponentVideo* videoComponent = static_cast<ComponentVideo*>(component);
+	//DrawEnableAndDeleteComponent();
```
gerard-queralt, line 24:
```suggestion
	DrawEnableAndDeleteComponent();
```

Source/FileSystem/ModuleResources.cpp
```diff
@@ -24,9 +25,12 @@
 #include "Resources/ResourceMaterial.h"
 #include "Resources/ResourceSkyBox.h"
 #include "Resources/ResourceTexture.h"
+#include "Resources/ResourceVideo.h"
 
 #include "Auxiliar/CollectionAwareDeleter.h"
 
+#include "AxoLog.h"
```
gerard-queralt, line 32: Not needed, since it's in the pch.

Source/Modules/ModuleUI.cpp
```diff
@@ -217,6 +218,17 @@ void ModuleUI::Draw2DGameObject(const GameObject* gameObject)
 			}
 		}
 
+		for (ComponentVideo* video : gameObject->GetComponents<ComponentVideo>())
+		{
+		
+			video->ReadVideoFrame();
```
gerard-queralt, line 224: Won't this make it so the video plays even if the game is not running in the engine?

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+			Init();
```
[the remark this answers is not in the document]
gerard-queralt: Nevermind, forgot that Components don't have an Init so the engine won't call it. However, it is still called twice from within the code (and can be called endlessly for all we know, since the method is public), so the question stands.

gerard-queralt (approved):

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+	// Set GL texture buffer
+	glGenTextures(1, &frameTexture);
+	glBindTexture(GL_TEXTURE_2D, frameTexture);
+	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
+	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
+	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
+	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
+	std::filesystem::path fs_path(video->GetAssetsPath());
+	std::string extension = fs_path.extension().string();
+	if (extension != ".avi")
```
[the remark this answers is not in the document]
ericdacal: Done

Source/DataModels/Components/UI/ComponentVideo.cpp
ericdacal: Done

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentVideo.cpp
ericdacal: Done

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+}
+
+ComponentVideo::~ComponentVideo()
+{
+}
+
+void ComponentVideo::Init()
+{
+	// Set GL texture buffer
+	glGenTextures(1, &frameTexture);
```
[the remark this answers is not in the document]
ericdacal: It seems that can I have two videos in a scene at same time, then I suposse that this works. It is based in ImageComponent.
image

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+{
+	Program* program = App->GetModule<ModuleProgram>()->GetProgram(ProgramType::SPRITE);
+	if (program)
+	{
+		glEnable(GL_BLEND);
+		glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
+
+		program->Activate();
+
+		ComponentTransform2D* transform = GetOwner()->GetComponentInternal<ComponentTransform2D>();
```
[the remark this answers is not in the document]
ericdacal: I've included this check:
`c++
if (!GetOwner()->HasComponent<ComponentTransform2D>())
{
	throw ComponentNotFoundException("Owner does not have ComponentTrasnform2D");
}`

Source/DataModels/Resources/ResourceVideo.cpp
ericdacal: I change this, now I copy the video to Lib, to can be play in Game also. Videos is an special case, because we don't convert the video in to Binary Extension because we make stream directly from the packets of video formats(avi, mp4 and mov).

Source/FileSystem/ModuleResources.cpp
ericdacal: Done

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+	{
+		canRotate = true;
+	}
+	OpenVideo(video->GetAssetsPath().c_str());
+}
+
+void ComponentVideo::Draw() const
+{
+	Program* program = App->GetModule<ModuleProgram>()->GetProgram(ProgramType::SPRITE);
+	if (program)
```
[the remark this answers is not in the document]
ericdacal: Done

Source/Modules/ModuleUI.cpp
ericdacal: Now there are options of Play and Pause video.

ericdacal:
> Aside from the comments, one general question. Is the video able to be paused? Mainly asking because I intend to finish the pause PR this week (#499), and it'd be a great help if this was already doable.

Now there are options of Play and Pause video.

Source/DataModels/Scene/Scene.cpp
```diff
@@ -363,35 +366,21 @@ GameObject* Scene::CreateUIGameObject(const std::string& name, GameObject* paren
...
-			imageFill->SetColor(float4(0.0f, 1.0f, 0.0f, 1.0f));
-			imageFill->SetRenderPercentage(slider->CalculateNormalizedValue());
-			slider->SetFill(fill);
-
-			GameObject* handle = CreateUIGameObject("Handle", gameObject, ComponentType::BUTTON);
-			ComponentTransform2D* handleTransform = handle->GetComponentInternal<ComponentTransform2D>();
-			handleTransform->SetSize(float2(25, 60));
-			handleTransform->CalculateMatrices();
-			slider->SetHandle(handle);
+			//... [Your existing slider code]
```
AuxPort66, line 383: ???

ericdacal: Slide came to life again!

Source/DataModels/Components/UI/ComponentVideo.cpp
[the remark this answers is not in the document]
AuxPort66: I didn't see this question until now. ComponentImage doesn't do the Set of the texture buffer in the component. The base of all the components that render is that they share the texture buffer if they have the same texture. That's why resource Load exists.

AuxPort66 (approved):
👍

Source/DataModels/Components/UI/ComponentVideo.cpp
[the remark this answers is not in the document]
gerard-queralt: Not sure if this is the right approach. Someone would need to catch the exception, which is a weird thing in this context, otherwise the app would crash. Do you think it would be an issue to log an error and abort the method with a return statement? If you can do that, you'd need to either deactivate the program or activate after the check.

Source/DataModels/Resources/ResourceVideo.cpp
gerard-queralt: Please add this to GameBuilder, the whole release should be doable from there. You can check how the Scenes folder are added to the Lib.

Source/Modules/ModuleUI.cpp
gerard-queralt: Thanks, but that wasn't the question. Can you check that the video does not play while in the Engine and not in Play mode? Or that if it does it is the intended behavior.

ericdacal: Actually we include a new option PlayWhenStart, that Play the video when is Play mode. In other hand I think that is better to can play in Engine mode, to previsualize that all is correct and can for example use the Vertital Rotation option.

AuxPort66 (approved):

Vlukkee (approved):
Thanks!

Source/Modules/ModuleUI.cpp
gerard-queralt: Perfect, just to confirm it was intended to work this way 👍

Source/DataModels/Components/UI/ComponentVideo.h
```diff
@@ -0,0 +1,97 @@
...
+struct SwsContext;
+
+class ComponentVideo : public Component, public Drawable
+{
+public:
+	ComponentVideo(bool active, GameObject* owner);
+	~ComponentVideo() override;
+	ComponentVideo(const ComponentVideo& toCopy);
+	void SetLoop(bool loop);
+	bool GetLoop();
```
gerard-queralt, line 15: This and the other getters can be `const`.

Source/DataModels/Components/UI/ComponentVideo.h
```diff
@@ -0,0 +1,97 @@
...
+	std::shared_ptr<ResourceVideo> GetVideo() const;
+	void SetNativeVideoFrameSize();
+	void UpdateVideoFrame();
+	void Draw() const override;
+	void Play();
+	void Pause();
+	void Stop();
+	void RestartVideo();
+	bool isPlayed();
+	bool isPlayAtStart();
```
gerard-queralt, line 31:
```suggestion
	bool IsPlayed();
	bool IsPlayAtStart();
```
Also, I have no idea what "is played" means. Whether it's done playing? I think we can find a better name.

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,337 @@
...
+	return video ? video->GetCanRotate() : false;
+}
+
+void ComponentVideo::SetVideo(const std::shared_ptr<ResourceVideo>& newVideo)
+{
+	this->video = std::move(newVideo);
+	if (video)
+	{
+		video->Load();
+		frameData = new uint8_t[video->GetFrameWidth() * video->GetFrameHeight() * 4];
```
gerard-queralt, line 326: Unless I'm missing it, we're not freeing this memory -> memory leak. I'd suggest using a unique pointer, but whatever you prefer.

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,337 @@
...
+		return;
+	}
+	glEnable(GL_BLEND);
+	glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
+
+	program->Activate();
+
+	if (!GetOwner()->HasComponent<ComponentTransform2D>())
+	{
+		throw ComponentNotFoundException("Owner does not have ComponentTrasnform2D");
```
gerard-queralt, line 72: As I said in the other comment, this is almost certainly going to cause a crash because no one is catching the exception. Just check if the transform is null and return if it is. Make sure to do so before all the other calls that require a closing one, like `program->Activate()` that requires you to call `program->Deactivate()`.

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,337 @@
...
+	}
+	ComponentTransform2D* transform = GetOwner()->GetComponentInternal<ComponentTransform2D>();
+
+	const float4x4& proj = App->GetModule<ModuleCamera>()->GetOrthoProjectionMatrix();
+	const float4x4& model = transform->GetGlobalScaledMatrix();
+	float4x4 view = float4x4::identity;
+
+	ComponentCanvas* canvas = transform->WhichCanvasContainsMe();
+	if (canvas)
```
gerard-queralt, line 81: Should we log a warning/error in the else? Don't think this is expected.

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,337 @@
...
+		
+	if (video)
+	{
+		video->Load();
+		program->BindUniformInt("hasDiffuse", 1);
+		glBindTexture(GL_TEXTURE_2D, video->GetGlFrameBuffer());
+		glTexImage2D(
+			GL_TEXTURE_2D, 0, GL_RGBA8, video->GetFrameWidth(), video->GetFrameHeight(), 0, GL_RGBA, GL_UNSIGNED_BYTE, frameData);
+	}
+	else program->BindUniformInt("hasDiffuse", 0);
```
gerard-queralt, line 107:
```suggestion
	else
	{
		program->BindUniformInt("hasDiffuse", 0);
	}
```

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+	glDisable(GL_BLEND);
+	program->Deactivate();
+	
+}
+
+void ComponentVideo::InternalSave(Json& meta)
+{
+	UID uidVideo = 0;
+	std::string assetPath = "";
```
gerard-queralt, line 123: This looks unused.

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,337 @@
...
+	RestartVideo();
+	ReadVideoFrame();
+	played = false;
+}
+
+void ComponentVideo::RestartVideo()
+{
+	avio_seek(video->GetFormat()->pb, 0, SEEK_SET);
+	if (av_seek_frame(video->GetFormat(), video->GetVideoStream(), -1, 0) >= 0)
+	{
```
gerard-queralt, line 301: Empty if, probably forgot something.

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+bool ComponentVideo::isPlayed()
+{
+	return played;
+}
+
+bool ComponentVideo::isPlayAtStart()
+{
+	return playAtStart;
+}
```
gerard-queralt, line 313: Why not inline these?

Source/DataModels/GameObject/GameObject.cpp
gerard-queralt: I see you've implemented a copy constructor.

Source/DataModels/GameObject/GameObject.cpp
```diff
@@ -624,12 +631,20 @@ Component* GameObject::CreateComponent(ComponentType type)
...
+		}
+
 		case ComponentType::BUTTON:
 		{
 			newComponent = std::make_unique<ComponentButton>(true, this);
 			break;
 		}
 
+
+
```
gerard-queralt, line 647: Don't think we need this.

Source/DataModels/Resources/ResourceVideo.h
```diff
@@ -0,0 +1,118 @@
...
+	unsigned int glFrameBuffer = 0;
+	bool canRotate;
+	std::string extension;
+	unsigned int frameRate;
+	unsigned int numFrames;
+
+	AVFormatContext* formatCtx = nullptr;
+	AVCodecContext* videoCodecCtx = nullptr;
+	AVPacket* avPacket = nullptr;
+	AVFrame* avFrame = nullptr;
```
gerard-queralt, line 53: This looks like is managed by the lib, but can you verify that it is? Just to make sure we're not leaking anything.

Source/DataModels/Scene/Scene.h
```diff
@@ -169,6 +170,7 @@ class Scene
 	std::vector<GameObject*> sceneGameObjects;
 	std::vector<ComponentCamera*> sceneCameras;
 	std::vector<ComponentCanvas*> sceneCanvas;
+	std::vector<ComponentVideo*> sceneVideos;
```
gerard-queralt, line 173: Is this used? Doesn't look like UI is reading it. Haven't checked all the code so might have missed it. If it's not used please remove it, last thing we need is more vectors of components going around :P

Source/FileSystem/ModuleResources.h
```diff
@@ -156,12 +158,17 @@ const std::shared_ptr<R> ModuleResources::RequestResource(const std::string path
 		std::string libraryPath = CreateLibraryPath(uid, type);
 
 		long long assetTime = fileSystem->GetModificationDate(assetPath.c_str());
-		long long libTime = fileSystem->GetModificationDate((libraryPath + GENERAL_BINARY_EXTENSION).c_str());
+		std::string libExtension = GENERAL_BINARY_EXTENSION;
+		if (type == ResourceType::Video)
+		{
+			libExtension = extension;
+		}
```
gerard-queralt, line 165:
```suggestion
		std::string libExtension = type == ResourceType::Video ? extension : GENERAL_BINARY_EXTENSION;
```

Source/FileSystem/ModuleResources.h
```diff
@@ -196,12 +203,18 @@ const std::shared_ptr<R> ModuleResources::RequestResource(const std::string path
...
```
gerard-queralt, line 210: Same here, if you agree with the previous suggestion.

gerard-queralt (changes requested):
Requesting changes because I'm almost sure there are memory leaks. Please also take a look at the comments from my review in August, some of them still stand and I consider them important, like [this one](https://github.com/Horizons-Games/Axolotl-Engine/pull/655#discussion_r1341266872).

Source/DataModels/Components/UI/ComponentVideo.cpp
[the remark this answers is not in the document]
gerard-queralt: Pretty sure this is no longer relevant, we don't seem to initialize explicitly anymore. Can you confirm that's the case?

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+	loop(false),
+	finished(false),
+	rotateVertical(false),
+	canRotate(false)
+{
+
+}
+
+ComponentVideo::~ComponentVideo()
+{
```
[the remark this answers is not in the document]
ericdacal: I think no because this is manage in ResourceVideo. But I include in this destructor the manage of the member frameData and scalerCtx.
`c++
ComponentVideo::~ComponentVideo()
{
	if (frameData)
	{
		delete[] frameData;
		frameData = nullptr;
	}

	if (scalerCtx)
	{
		sws_freeContext(scalerCtx);
		scalerCtx = nullptr;
	}
}
`

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+		program->Activate();
+
+		ComponentTransform2D* transform = GetOwner()->GetComponentInternal<ComponentTransform2D>();
+
+		const float4x4& proj = App->GetModule<ModuleCamera>()->GetOrthoProjectionMatrix();
+		const float4x4& model = transform->GetGlobalScaledMatrix();
+		float4x4 view = float4x4::identity;
+
+		ComponentCanvas* canvas = transform->WhichCanvasContainsMe();
+		if (canvas)
```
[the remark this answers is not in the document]
ericdacal: Video cannot be create outside of a Canvas, but can be move outside. I include a ERROR_LOG for this case.

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+		program->BindUniformFloat("renderPercentage", 1.0f);
+		program->BindUniformInt("direction", 1);
+		
+		if (initialized)
+		{
+			program->BindUniformInt("hasDiffuse", 1);
+			glBindTexture(GL_TEXTURE_2D, frameTexture);
+			glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA8, frameWidth, frameHeight, 0, GL_RGBA, GL_UNSIGNED_BYTE, frameData);
+		}
+		else program->BindUniformInt("hasDiffuse", 0);
```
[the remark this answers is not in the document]
ericdacal: Done

Source/DataModels/Components/UI/ComponentVideo.cpp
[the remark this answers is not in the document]
ericdacal: Now that change, the Init is done inside the VideorResource not in the Component.

Source/DataModels/Components/UI/ComponentVideo.h
ericdacal: Removed line

Source/FileSystem/ModuleResources.h
ericdacal: Done

Source/FileSystem/ModuleResources.h
ericdacal: Done

Source/DataModels/Scene/Scene.h
ericdacal: Yes, is use to initialize the VideoComponent in ModuleScene.

```c++
void ModuleScene::InitVideoComponents()
{
	for (ComponentVideo* componentVideo : loadedScene->GetSceneVideos())
	{
		if (componentVideo->GetOwner()->IsActive() && componentVideo->GetPlayAtStart())
		{
			componentVideo->Play();
		}
	}
}
```

Source/DataModels/Resources/ResourceVideo.h
ericdacal: I check and that variable has to be manage, I do this in Destructor:
```c++
void ResourceVideo::InternalUnload()
{
	glDeleteTextures(1, &glFrameBuffer);
	glFrameBuffer = 0;
	if (formatCtx)
	{
		avformat_close_input(&formatCtx);
	}
	if (videoCodecCtx)
	{
		avcodec_close(videoCodecCtx);
		av_free(videoCodecCtx);
	}
	if (avPacket)
	{
		av_packet_unref(avPacket);
		av_packet_free(&avPacket);
	}
	if (avFrame)
	{
		av_frame_free(&avFrame);
	}
}
```

Source/DataModels/Resources/ResourceVideo.cpp
ericdacal: Done

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+	if (avcodec_open2(videoCodecCtx, videoDecoder, nullptr) < 0)
+	{
+		LOG_ERROR("Couldn't open video codec.");
+		return;
+	}
+
+	// Set video parameters and Allocate frame buffer
+	frameWidth = videoCodecParams->width;
+	frameHeight = videoCodecParams->height;
+	frameData = new uint8_t[frameWidth * frameHeight * 4];
```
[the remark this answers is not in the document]
ericdacal: frameData now is delete in the Destructor

Source/DataModels/Components/UI/ComponentVideo.h
ericdacal: Done

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+		LOG_ERROR("Couldn't allocate AVPacket.");
+		return;
+	}
+	avFrame = av_frame_alloc();
+	if (!avFrame)
+	{
+		LOG_ERROR("Couldn't allocate AVFrame.");
+		return;
+	}
+	initialized = true;
```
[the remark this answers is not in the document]
ericdacal: Now is manage for VideoResource and I delete this variable

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+}
+
+void ComponentVideo::ReadVideoFrame()
+{
+	if (initialized)
+	{
+		int response = -1;
```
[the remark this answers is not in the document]
ericdacal: It doesn't exist now

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+		int error = 0;
```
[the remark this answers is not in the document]
ericdacal: I change the name for result

Source/DataModels/Components/UI/ComponentVideo.cpp
ericdacal: Done

Source/DataModels/Components/UI/ComponentVideo.cpp
ericdacal: It is a deprecated function libav_err2str.

Source/DataModels/Components/UI/ComponentVideo.h
ericdacal: Now is manage and delete in VideoResource

Source/DataModels/Components/UI/ComponentVideo.h
ericdacal: Done

Source/DataModels/Components/UI/ComponentVideo.cpp
ericdacal: Removed

Source/DataModels/Components/UI/ComponentVideo.cpp
```diff
@@ -0,0 +1,334 @@
...
+	meta["assetPathVideo"] = video->GetAssetsPath().c_str();
+}
+
+void ComponentVideo::InternalLoad(const Json& meta)
+{
+	loop = meta["loop"];
+	std::string path = meta["assetPathVideo"];
+	rotateVertical = meta["rotateVertical"];
+	canRotate = meta["canBeRotate"];
+	bool resourceExists = !path.empty() && App->GetModule<ModuleFileSystem>()->Exists(path.c_str());
```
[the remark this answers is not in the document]
ericdacal: Maybe it is not necessary, that is done because also is done in ComponentImage

Source/DataModels/Components/UI/ComponentVideo.cpp
ericdacal: This line restart the video if fails go inside the if. I include now a LOG_ERROR that tell the user that there is a fail in the restart.

Source/DataModels/Components/UI/ComponentVideo.cpp
ericdacal: Is the same but one for packets and the other to avlib, I change the name of error to be more clarify.

Source/DataModels/Components/UI/ComponentVideo.cpp
ericdacal: I include a log if not found the transform

Source/DataModels/Components/UI/ComponentVideo.cpp
ericdacal: dest is define in stack and do not need to be remove, and frameData now is delete in the Destructor

Source/DataModels/Components/UI/ComponentVideo.cpp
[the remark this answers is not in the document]
gerard-queralt: It wasn't when I added the comment. Now everything seems to be getting freed, thanks!

Source/DataModels/Components/UI/ComponentVideo.cpp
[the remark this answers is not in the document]
gerard-queralt: Would like to know why but not that important I guess.

Source/DataModels/Components/UI/ComponentVideo.cpp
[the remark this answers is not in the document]
gerard-queralt: It still does. Not a big deal, just wondering if there's something I'm no reading correctly.

Source/DataModels/Components/UI/ComponentVideo.cpp
gerard-queralt: Just a tip: the formatter automatically calls GetName on GameObjects, so no need to do it yourself 😉

Source/DataModels/Components/UI/ComponentVideo.cpp
gerard-queralt: Now it's deleted in the destructor, but if we had allocated memory before reaching this line you'll leak it. Make sure to delete before allocating new memory.

Source/DataModels/Resources/ResourceVideo.h
gerard-queralt: Awesome, thank you!

Source/DataModels/Components/UI/ComponentVideo.cpp
ericdacal: Finally, I change this for that:
```c++
if (!GetOwner()->HasComponent<ComponentTransform2D>())
{
  LOG_ERROR("Owner does not have ComponentTrasnform2D");
  return;
}
```

Source/DataModels/Components/UI/ComponentVideo.cpp
ericdacal: I add a delete of frameData in each Draw:

```c++
if (video)
{
video->Load();

if (frameData)
{
	delete[] frameData;
	frameData = nullptr;
}


  frameData = new uint8_t[video->GetFrameWidth() * video->GetFrameHeight() * 4];
  memset(frameData, 0, video->GetFrameWidth() * video->GetFrameHeight() * 4);
  ReadVideoFrame();
}
```

Source/DataModels/Components/UI/ComponentVideo.h
ericdacal: Done

gerard-queralt (approved):

Vlukkee (approved):

MacXxs:
Looking great Eric! Just one thing.
When playing a video is constantly printing this in the console. Shouldn't it be only printed once?

image

ericdacal:
> Looking great Eric! Just one thing. When playing a video is constantly printing this in the console. Shouldn't it be only printed once?
> 
> image
> 
> A part from that, why isn't the video_test.avi showing the option to be rotated? Is it because of the format?

This is a debug log that I can remove, about opening_previw.avi, yes I the avi format can not be rotate, but .mp4 and .mov yes.

PR #657, opened 2023-08-21
feat: add categories in WindowInspector context menu

gerard-queralt:
image

Also cleaned up the code (which was the original idea, as the branch name suggests, but got carried away :P)


gerard-queralt:
Also taking the opportunity to merge a writable enum macro that I wrote a few weeks back just for fun. Figured I could use it here, and that it might be helpful to have. Migrating existing enums is most likely impossible, however, because the lack of versioning in our serialization would mean we'd need to regenerate basically everything. Which, if someone wants to do, be my guess :P

markmeliaa (approved):

markmeliaa (approved):

MarcGonzalezMoratona:
There are conflicting files 👀

Piterga12 (commented):
Works great but there's conflicts on the filters

MarcGonzalezMoratona (dismissed):

MarcGonzalezMoratona (approved):

Piterga12 (approved):
Good job!

PR #658, opened 2023-08-23
fix: various fixes to copy constructors

gerard-queralt:
Will also try to add fixes to other reported issues once I manage to reproduce them.

### DIFF
--- a/Source/DataModels/Components/ComponentAnimation.cpp
+++ b/Source/DataModels/Components/ComponentAnimation.cpp
@@ -1,7 +1,7 @@
 #include "StdAfx.h"
 
-#include "ComponentAnimation.h"
 #include "Application.h"
+#include "ComponentAnimation.h"
 #include "ComponentTransform.h"
 
 #include "Animation/AnimationController.h"
@@ -21,18 +21,30 @@ ComponentAnimation::ComponentAnimation(const bool active, GameObject* owner) :
 	Component(ComponentType::ANIMATION, active, owner, true),
 	drawBones(false)
 {
-	controller = new AnimationController();
+	controller = std::make_unique<AnimationController>();
 	lastState = NON_STATE;
+}
+
+ComponentAnimation::ComponentAnimation(const ComponentAnimation& other) :
+	Component(other),
+	controller(std::make_unique<AnimationController>()),
+	stateMachine(other.stateMachine),
+	parameters(other.parameters),
+	defaultPosition(other.defaultPosition),
+	actualState(other.actualState),
+	nextState(other.nextState),
+	lastState(other.lastState),
+	drawBones(other.drawBones)
+{
 }
 
 ComponentAnimation::~ComponentAnimation()
 {
-	delete controller;
 }
 
-AnimationController* ComponentAnimation::GetController()
+AnimationController* ComponentAnimation::GetController() const
 {
-	return controller;
+	return controller.get();
 }
 
 const std::shared_ptr<ResourceStateMachine>& ComponentAnimation::GetStateMachine() const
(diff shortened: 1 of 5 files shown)


markmeliaa (approved):
Works well! Now the engine does not break after duplicating a `ComponentAnimation` :)

markmeliaa (dismissed):

Source/DataModels/Components/ComponentAnimation.cpp
```diff
@@ -22,36 +22,44 @@ ComponentAnimation::ComponentAnimation(const bool active, GameObject* owner) :
...
 	drawBones(false),
 	firstEntry(true),
-	controller(new AnimationController()),
-	stateMachineInstance(new StateMachine())
+	controller(std::make_unique<AnimationController>()),
+	stateMachineInstance(std::make_unique<StateMachine>())
+{
+}
+
+ComponentAnimation::ComponentAnimation(const ComponentAnimation& other) :
```
MarcGonzalezMoratona, line 30: here the parameter is called other, in other components is called toCopy. I would appreciate if you could keep consistency between files 😄

Source/DataModels/Components/ComponentAnimation.h
```diff
@@ -20,13 +20,14 @@ class ComponentAnimation : public Component, public Drawable, public Updatable
...
+	ComponentAnimation(const ComponentAnimation& other);
 	~ComponentAnimation() override;
 
-	AnimationController* GetController();
+	AnimationController* GetController() const;
 
 	StateMachine* GetStateMachineInstance() const;
-	const std::shared_ptr<ResourceStateMachine>& GetStateMachine() const;
-	void SetStateMachine(const std::shared_ptr<ResourceStateMachine>& stateMachine);
+	std::shared_ptr<ResourceStateMachine> GetStateMachine() const;
```
MarcGonzalezMoratona, line 29: why did we remove the const reference of the shared_ptr? was it unnecessary before?

Source/DataModels/Components/ComponentAnimation.cpp
gerard-queralt: My bad :P

Source/DataModels/Components/ComponentAnimation.h
gerard-queralt: Not unnecessary, but in my opinion it's incorrect. If we pass/return it as a const reference, we avoid creating a copy of the object, which in general is a good thing. With shared pointers, however, avoiding to create a copy also means that we won't increase the internal reference counter that decides when it can be destroyed, which to me defeats the whole purpose. I talked a bit about this in this section of our wiki: https://github.com/Horizons-Games/Axolotl-Engine/wiki/Smart-Pointers-Tips-&-Tricks#but-why-cant-we-return-the-unique-pointer-by-reference.

In practice, we were always creating copies, so this won't have any impact whatsoever in performance. Assuming we're using the old signatures which receive and return a const reference to a shared pointer:
```c++
// setter code of StateMachine
this->stateMachine = stateMachine;
```
We are invoking the copy constructor here anyway using this assignment.

```c++
std::shared_ptr<StateMachine> receivedStateMachine = component->GetStateMachine();
```
Even though we are returning a const reference, the caller of the method is capturing by value, which means we make a copy too. However, we are at the mercy of the caller of the API, and they could easily break it:
```c++
const std::shared_ptr<StateMachine>& receivedStateMachine = component->GetStateMachine();
```
With the above code, we have no guarantee that the pointer won't be destroyed while the user operates on it, leading to a crash.

All in all, my thought process is: if we want the user to make a copy, don't leave it up to them. In my opinion, this makes our engine more robust.

Having said all of that, there are people that disagree with this view and think it's too defensive, like Carlos himself, so take it for what it is: an opinion.

markmeliaa (approved):

Dedado3000 (approved):

PR #662, opened 2023-08-24
feat: Final Boss behaviour, battle and attacks

markmeliaa:
Okay, here we go, I hope you brought something to drink!

What does this bring to the table you may ask? Well, this pr brings the foundations for what will be the final boss fight.
A lot of files were changed through the development of this pr, but the most important ones are the following:

- `FinalBossScript` -> this script manages the final boss completely, when to trigger their attacks, when to change from one phase from another... 
This is the main file that you'd have to tweak around during your testing, as in the `Update` of this script there's a function called `TryAttacksIndividually` which shows the attacks alone when the player approaches Bix, and another one called `ManageBossPhases`, which manages a completely funcional behaviour with four phases depending on how much life the boss has (Right now you may prefer testing the attacks with the first function, as the light and heavy attacks are currently broken after the merge of #580).

This Final Boss has four different attacks, just as it was stated in our [GDD](https://docs.google.com/document/d/1t81LIV04Y-2US6tr9BdEuSPTXgEg4NBq0aMZx_zQzoM/edit), and they are managed by the following scripts:
- **CHAOTIC PLASMA HAMMER**
Managed by `ShockWaveAttackScript` and `ShockWaveAttackAreaScript`. There are two versions of these attack, one more defensive (which is triggered when the player casually approaches the boss) and another one more aggresive (which makes the boss run towards the player and attack them). This is managed by two areas (two spherical `RigidBodies`) and the objective is to damage the player if it's on the floor and they are hit by the shockwave range between the two areas.

- **BRUTAL CHARGE**
Managed by `BossChargeAttackScript` and `BossChargeRockScript`. When triggered, the boss will step back a little and then charge forward with all their might until they hit a wall. Upon hitting a wall, rock will fall from the sky through the area where the boss charged and they will damage the player if they hit him while falling. These rocks will stay in the battle area for a while, difficulting the movement of the player.

- **ENERGY SHIELD**
Managed by `BossShieldAttackScript` and `BossShieldScript`. This attack will activate a shield around the boss for a certain time, making them invulnerable to attacks while healing himself while the shield is up. While this attack takes place, several enemies will spawn around to keep the player from stoping this attack, so the final boss can heal as much as possible. (Right now the enemies cannot be duplicated correctly from scripts, so until that works, I programmed a work-around with a "pool" of enemies that enable/disable). The only way to stop this shield attack is by hitting a rock or whatever there will be in range, making it fall over the final boss' head, stunning them and making the attack stop. (This is not yet programmed but I left it as prepared as I could).

- **LAST WISH MISILES**
Managed by `BossMissilesAttackScript` and `BossMissilesMissileScript`. When really low on health (during the **LAST_RESORT** phase), the final boss will disappear the battle area and trigger a missile attack that will spawn several missiles through the battle area, trying to inflict as much damage as possible to the player.

Aside from that, all the logic for the final boss was programmed, to change between phases, what happens in each phase... And the attacks could be modified widely through all the values visible on the inspector for each script.
As you could see, a lot of these attacks relay in having the tag `Wall` for the walls where the **BRUTAL CHARGE** will collide and stop and `Floor` for the floor itself, to trigger the missiles, know when the rocks hit the floor... Idk if that may be an issue but I'm hoping not because tags are easy enough to manage.
You can check all these in the `Gameplay_FinalBoss.axolotl` scene.

----------------------------------------

Additional work carried out during this pr to improve its funcionality:
- Built a basic blockout of the boss area, nothing too shabby.
- Fix sound for the steps when finishing jumping or dashing.
- Don't jump or move while dashing (now it looks like a more steady dash).
- Add vectors for the waypoints of the patrol and the powerups of the enemy death.
- Little rework of the powerup system, when a powerup spawns, instead of being set as a child of the dead enemy, just tp it to the position of the dead enemy, so it can be reused again. Also made that the player could attract and get powerups even when they already have one active and one saved up (just like in `Super Mario Bros.`, if you get a powerup when you have no more space, it will simply disappear).
- Only 2 decimals shown by the float fields of the `ComponentScript` in the `Inspector Window` (4 was way too much IMO).
- Small rework to the patrol, now when an entity gets out of its patrol route, when getting back to the patrol route they will do so trough its neares waypoint at the moment, instead as for the nearest waypoint stored when they left the patrol route.
Also added a random patrol for the bosses (looks really weird, but idk how to perform a realistic movement for the bosses with what we have now).
- Fix particle timer for the light finisher (it was set way to much time).
- Made that both the **BRUTAL CHARGE** and the **ENERGY SHIELD** can be reused by any other mini-boss that we may have in the future.

I hope you like it! (I really hope it, I've spent practically every minute of this last week doing this and IMO, it had a really great outcome, altough yet unfinished).


OriolSors (approved):
Astounding job @markmeliaa, really good approaches and the feeling is close to be a real final boss battle (leaving art out, of course). Few things to take into account:

- Bix gets stucked with her attack. I think this was commented before but now I realized that. It has nothing to do with this PR but it gets the experience more tough to play. 
- The final boss gets stucked with his brutal charge attack (I mean, with the rocks fallen). Maybe this won't be a problem because 2 things: the cooldown and probabilities of the attacks in the ManagePhases and if we treat the rocks as obstacles for the NavMesh. But I saw it necessary to comment it just in case.

For the rest, all functionalities seem be working and the code structure is clear and readable.

markmeliaa:
First of all, thank you both for you kind words and for taking the time to check and test this pull request!

Regarding your comments:
> Bix gets stucked with her attack. I think this was commented before but now I realized that. It has nothing to do with this PR but it gets the experience more tough to play.

Yes, sadly after merging #580, the simple attacks get stuck (Idk why, I tried to check that pr and it touched nothing related to how the attacks work, so it's a mystery, so I told it to Sara hoping that, as she was the owner of the pr, knows why this is happening). But yea, as you say, that sadly hinders the testability of the boss behaviour itself, so I hope it gets fixed asap.

> The final boss gets stucked with his brutal charge attack (I mean, with the rocks fallen). Maybe this won't be a problem because 2 things: the cooldown and probabilities of the attacks in the ManagePhases and if we treat the rocks as obstacles for the NavMesh. But I saw it necessary to comment it just in case.

The rocks make the whole attack area kind of messy for the continued mobility of the boss, I tried to make it as playable as possible but I'm afraid that until we have the `NavMesh` the rocks will make the final boss get stucked at certain times (thankfull the rocks disappear after a certain time, so the boss doesn't get stuck forever, but ideally it should never get stuck).

DADSILVER (approved):
Excellent job, I've tried it and liked it. I only saw one thing I didn't like. When you perform the jumping downward attack, you launch the boss into the air along with its areas. The launching into the air part should be removed.

markmeliaa:
> Excellent job, I've tried it and liked it. 

Thanks!

> When you perform the jumping downward attack, you launch the boss into the air along with its areas. The launching into the air part should be removed.

Wdym? That the jump attack shouldn't push the boss back?

Game/Scripts/BossChargeAttackScript.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+void BossChargeAttackScript::WallHitAfterCharge() const
+{
+	btRigidBody* enemybtRigidbody = rigidBody->GetRigidBody();
+	rigidBody->DisablePositionController();
+	rigidBody->DisableRotationController();
+	enemybtRigidbody->setAngularFactor(btVector3(0.0f, 0.0f, 0.0f));
+	enemybtRigidbody->setLinearVelocity(btVector3(0.0f, 0.0f, 0.0f));
+
+	float3 nextPosition = transform->GetGlobalPosition() - transform->GetGlobalPosition();
```
PaablooCH, line 182: this is not the same as doing `= float3::zero`? if this result is always 0 you don't need to normalize the vector

Game/Scripts/BossChargeRockScript.cpp
```diff
@@ -0,0 +1,93 @@
...
+			DestroyRock();
+		}
+	}
+}
+
+void BossChargeRockScript::OnCollisionEnter(ComponentRigidBody* other)
+{
+	if (rockState == RockStates::SKY && other->GetOwner()->CompareTag("Rock"))
+	{
+		DeactivateRock();
```
PaablooCH, line 46: why you deactivat it and not destroy the go?

Game/Scripts/BossMissilesAttackScript.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+void BossMissilesAttackScript::RotateToTarget(const float3& targetPosition) const
+{
+	Quat errorRotation =
+		Quat::RotateFromTo(transform->GetGlobalForward().Normalized(),
+			(targetPosition - transform->GetGlobalPosition()).Normalized());
+
+	rigidBody->SetRotationTarget(errorRotation);
+}
```
PaablooCH, line 163: i see this function a lot, maybe we can make it static and avoid replicate it each time we need to rotate (probably the same with other actions like move or look)

PaablooCH (approved):
well done, a side of the problem mentioned with the attack I don't see any other big problem. I don't see having the camera as you presented, I would put a unique sampler where we can view the whole scenario, but I know is a pre state.

Game/Scripts/BossMissilesAttackScript.cpp
markmeliaa: This function was done by @OriolSors for the seek movement months ago, because the `SetRotationTarget` of the transform didn't work as expected. IMO, the best option would be that the `SetRotationTarget` performs this calculus by itself, but idk how the rigidbody insides work so it was not for me to decide.

I also tried setting it in the `FinalBossScript` but the rotation is only handled in certain parts of certain attacks, not always, so that is why it was set in each script, so each attack can correctly call it when needed to rotate towards the target.

Game/Scripts/BossChargeRockScript.cpp
markmeliaa: Currently, when a gameobject is deleted with a `ParticleSystem` it throws an error. I was told that this is fixed in another branch, but as idk if that fix will be up when the particles are added to the rocks, before destroying a rock, first the `ParticleSystem` would have to be deleted properly, that's why I setted it up this way.
Also, deactivating it makes that no resources are spend besides running its scripts (which, for the rocks, is only one script), so it's not really an issue rn, the timer of the rock will manage its destruction eventually.

Game/Scripts/BossChargeAttackScript.cpp
```diff
@@ -0,0 +1,251 @@
...
+{
+	return chargeState != ChargeState::NONE;
+}
+
+void BossChargeAttackScript::SpawnRock(const float3& spawnPosition)
+{
+	LOG_DEBUG("Spawn rock at {}, {}, {}", spawnPosition.x, spawnPosition.y, spawnPosition.z);
+
+	GameObject* newRock = App->GetModule<ModuleScene>()->GetLoadedScene()->
+		DuplicateGameObject("Rock Copy", rockPrefab, rockPrefab->GetParent());
```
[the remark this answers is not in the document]
markmeliaa: Not really, because that would somehow limit the potenital rocks that could spawn in the boss area, and although you may say that having 500 rocks (for example) would not hinder performance as they will be deactivated, I see it unnecessary to have an amout of rocks already created when you can just spawn then through the code.

With the powerups is different, because only few of them would be active at a time and their functionality is a bit different, but with the rocks, as they are simple and easy to duplicate, why not do it?

Game/Scripts/BossChargeAttackScript.cpp
markmeliaa: TBH, I picked this functionality from the jumping finisher attack and I haven't even realized about this line. Theorically, yes, changing this to `= float3::zero` would mean no changes in the functionality and we'll avoid having to normalize the vector, I'll make a few tries to see if everything works correctly and get back to you with an answer.

markmeliaa:
> well done, a side of the problem mentioned with the attack I don't see any other big problem. 

Thank you, Pablo!

> I don't see having the camera as you presented, I would put a unique sampler where we can view the whole scenario, but I know is a pre state.

To be franc, me neither, but that is not really for this pr to decide, as it more a level programming thing and I guess they'll decide about that when the final arena is fully built.

Game/Scripts/BossChargeAttackScript.cpp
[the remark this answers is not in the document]
PaablooCH: i comment it because each time you create and destroy a mesh, the batches need to recalculate all the data they have, just to have it in mind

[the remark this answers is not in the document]
markmeliaa: I'll keep that in mind, don't worry. Thanks for pointing it out!

Game/Scripts/BossMissilesAttackScript.cpp
OriolSors: It would be a great idea to put this code snippet inside the rotation controller of the rigidbody, but certainly two things:

- the idea of the rotation controller for rigidbodies was a temporary solution, but knowing that is currently working we can decide to adapt it more
- rotation controller takes the rotation error (what is computed here) and uses it to compute the final rotation value in the current frame. So, depending on the scenario, we would need one type of error computation or another (I mean, using "RotationFromTo(), LookAt()...). If we add this piece of code to the inner rigidbody behaviour we will need to take this calls into account...

Game/Scripts/BossChargeAttackScript.cpp
```diff
@@ -0,0 +1,251 @@
...
+	newRockRigidBody->Enable();
+
+	rocksSpawned.push_back(newRock);
+}
+
+void BossChargeAttackScript::MakeRocksFall() const
+{
+	for (GameObject* spawnedRock : rocksSpawned)
+	{
+		if (!spawnedRock->IsEnabled())
```
gerard-queralt, line 227: Should we not trim the vector in such cases? I mean, from what I understand once a rock is disabled it won't be enabled again, and if that's correct, we can just remove those. I'd suggest either an iterator loop and remove them on the spot, or a pre-filter; for the latter I'd use [`std::erase_if`](https://en.cppreference.com/w/cpp/container/vector/erase2), but we can also use [`copy_if`](https://en.cppreference.com/w/cpp/algorithm/copy) or [`views::filter`](https://en.cppreference.com/w/cpp/ranges/filter_view) and construct a new vector from the view.

Game/Scripts/BossMissilesAttackScript.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+void BossMissilesAttackScript::SwapBetweenAttackStates(float deltaTime)
+{
+	// Manage the rotation and timers
+	if (missilesAttackState == AttackState::STARTING_SAFE_JUMP)
+	{
+		RotateToTarget(midJumpPosition);
+
+		if (transform->GetGlobalPosition().Equals(midJumpPosition, 0.5f))
```
gerard-queralt, line 78: Are we sure we will never pass through this mid-point and skip this check? Don't know other alternatives off the top of my head, but it might be worth looking into some.

Game/Scripts/BossShieldAttackScript.cpp
```diff
@@ -0,0 +1,252 @@
...
+	REGISTER_FIELD(battleArenaAreaSize, ComponentRigidBody*);
+}
+
+void BossShieldAttackScript::Start()
+{
+	shieldingTime = shieldingMaxTime;
+
+	if (enemiesToSpawnParent == nullptr)
+	{
+		return;
```
gerard-queralt, line 43:
```suggestion
		LOG_ERROR("enemiesToSpawnParent not set in BossShieldAttackScript owned by {}", GetOwner());
		throw ComponentNotFoundException();
```
This is just a temporary fix until we merge #654. If the script can run without the code below this check (empty vectors and no health system), no need to throw an exception, we can just like return like we do now.

Game/Scripts/BossShieldAttackScript.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+	isShielding = true;
+	shieldAttackCooldown = shieldAttackMaxCooldown;
+
+	triggerEnemySpawning = true;
+}
+
+bool BossShieldAttackScript::CanPerformShieldAttack() const
+{
+	return shieldAttackCooldown <= 0.0f && !isShielding;
```
gerard-queralt, line 86:
```suggestion
	return shieldAttackCooldown <= 0.0f && !IsAttacking();
```
In the unlikely case we add more attacks

Game/Scripts/BossShieldAttackScript.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+GameObject* BossShieldAttackScript::SelectEnemyToSpawn()
+{
+	if (enemiesReadyToSpawn.empty())
+	{
+		return nullptr;
+	}
+
+	int enemyRange = static_cast<int>(enemiesReadyToSpawn.size());
+	int randomEnemyIndex = rand() % enemyRange;
```
gerard-queralt, line 184: This many instances of `rand() % USER_PROBABILITY` suggests we might be lacking a common place to do this, either a script or a module. We could also do that framerate correction I suggested there.

Game/Scripts/EnemyVenomiteScript.cpp
```diff
@@ -90,73 +105,108 @@ void EnemyVenomiteScript::Update(float deltaTime)
...
 		componentAnimation->SetParameter("IsRunning", false);
 	}
+}
+
+void EnemyVenomiteScript::ResetValues()
+{
+	//componentAudioSource->PostEvent(/* Stop Venomite Sounds */);
+	std::unordered_map<std::string, TypeFieldPairParameter> componentAnimationParameters =
+		componentAnimation->GetStateMachine()->GetParameters();
+	for (std::pair<std::string, TypeFieldPairParameter> parameter : componentAnimationParameters)
```
gerard-queralt, line 198:
```suggestion
	for (const std::pair<std::string, TypeFieldPairParameter>& parameter : componentAnimationParameters)
```

Game/Scripts/PatrolBehaviourScript.cpp
```diff
@@ -4,63 +4,84 @@
...
-	{
-		wayPointTwoTransform = wayPointTwo->GetComponent<ComponentTransform>();
+		transformWaypoints.push_back(waypoint->GetComponent<ComponentTransform>());
 	}
 
 	ownerRigidBody = owner->GetComponent<ComponentRigidBody>();
 	ownerTransform = owner->GetComponent<ComponentTransform>();
 
-	currentWayPointTransform = wayPointOneTransform;
+	currentWayPointTransform = transformWaypoints[0];
```
gerard-queralt, line 27: Paranoic, but should we check that there's at least one element in the vector?

gerard-queralt (commented):
Congratulations and a big thank you to the gameplay team for the amazing work done!!

As always, haven't tested, but left a bunch of picky comments, as is also tradition :P

markmeliaa:
> As always, haven't tested, but left a bunch of picky comments, as is also tradition :P

Appreciate it! As I'm out this week I can't perform those required changes now, but as soon as I get back, they will get addressed in another pr, because we need to merge this as soon as possible so the project can carry on.

Game/Scripts/FinalBossScript.h
```diff
@@ -0,0 +1,81 @@
...
+	ComponentTransform* targetTransform;
+
+	PatrolBehaviourScript* patrolScript;
+	HealthSystem* bossHealthSystem;
+	BossChargeAttackScript* chargeAttackScript;
+	ShockWaveAttackScript* shockWaveAttackScript;
+	BossShieldAttackScript* shieldAttackScript;
+	BossMissilesAttackScript* missilesAttackScript;
+
+	bool isLookingAtTarget;
```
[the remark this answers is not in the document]
markmeliaa: True thx!

Game/Scripts/FinalBossScript.cpp
```diff
@@ -0,0 +1,321 @@
...
+	// Uncomment this line to check the attacks individually (you have to activate each one of them below)
+	//TryAttacksIndividually();
+	
+	// Comment this line if you uncomment the one above and vice versa
+	ManageBossPhases();
+}
+
+void FinalBossScript::ManageBossPhases()
+{
+	if ((bossHealthSystem->GetCurrentHealth() < bossHealthSystem->GetMaxHealth() * 0.2f &&
```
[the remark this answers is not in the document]
markmeliaa: I don't see this viable at all. Although I see why giving the ability to the user to change it through the UI would be useful, it would mean that, as you say, check that neither of the three percentages are incorrect, which would be a bit of a chaos.

Imagine that we have three different percentages: `triggerLastResortPhase`, `triggerDefensivePhase` and `triggerAggresivePhase`, and the way to check that either of them are in its correct value range would be that `triggerLastResortPhase` is between 0 and `triggerDefensivePhase`, `triggerDefensivePhase` is between `triggerLastResortPhase` and `triggerAggressivePhase`, and finally, that `triggerAggressivePhase` is between `triggerDefensivePhase` and 100. 
In paper this is not difficult to apply but that would open the door to possibly skip completely a phase if you set their values to, for example, 0.49, 0.50 and 0.80, which would skip completely the defensive phase.
The solution that ocurs to me is to set a threshold between states (for example, `triggerDefensivePhase` would need to be at least 0.3 greater than `triggerLastResortPhase`, but that would make that the percentages can't vary much (just instead of them being 0.2 and 0.5, they could be 0.1 and 0.4) and I don't see it worth enough to change it and give the ability for the player to set the values through the UI.

tl;dr: We would not get enough value from setting it through the UI because those values need to have some margin between them, so changing the percentages to variables modifiables from the UI is not worth it.

Game/Scripts/FinalBossScript.cpp
```diff
@@ -0,0 +1,321 @@
...
+		bossPhase == FinalBossPhases::DEFENSIVE) || bossPhase == FinalBossPhases::LAST_RESORT)
+	{
+		if (bossPhase != FinalBossPhases::LAST_RESORT)
+		{
+			LOG_INFO("Final Boss is in the LAST RESORT PHASE");
+			bossPhase = FinalBossPhases::LAST_RESORT;
+		}
+
+		ManageLastResortPhase();
```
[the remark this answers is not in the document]
markmeliaa: > First of all: do we want the states to depend on both the health and the current state, or only the health? My main worry is that, if the boss loses a big chunk of health at once, enought to skip an entire phase, we will miss a frame of the more advanced phase.

The current phase should be taken into account because boss phases are not regressive, if the boss enters the `DEFENSIVE` phase, he should not go back to the `AGGRESSIVE` phase, changing back the phase feels counter-intuitive and not user friendly, because IMO, that would confuse the player.

> This is not inherently wrong, but I feel the boss should be able to skip a phase. This is probably wildly paranoic and can't happen, but I feel it's worth discussing just in case.

Skipping a boss' phase with just one attack would not be possible because the health required to change a phase is wildy superior to the damage that the player would do with the strongest of their attacks, but I agree that, given the case, a phase should be skipped accordingly with the damage recieved.

> Second thing I see is that we're changing the phase and handling it on the same method. This makes the code more complex: we want to handle the phase, but we want to only update it once. So we have this inner if. I'd suggest doing this in two steps

The thing is, I did that approach during the early stages of this pr but changed it and I don't remember why, but you are right that is a far better approach, I'm changing it now.

Game/Scripts/FinalBossScript.cpp
```diff
@@ -0,0 +1,321 @@
...
+void FinalBossScript::ReactivateMovement() const
+{
+	rigidBody->SetKpForce(0.3f);
+	rigidBody->SetIsKinematic(false);
+	rigidBody->SetUpMobility();
+}
+
+void FinalBossScript::ManageNeutralPhase()
+{
+	int chargeChance = rand() % 1500; // Trust me, 1 in 1500 chance is enough
```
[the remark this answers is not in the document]
markmeliaa: > With these change, the above computations give
> 
> * 1/(1500 * 60/60) at 60fps -> 60/1500 chances each second -> 0.04 chances each second
> * 1/(1500 * 30/60) at 30fps -> 30/750 chances each second -> 0.04 chances each second

Well thought, but do we have a `FRAMERATE` global in our engine? I'm only seeing `MAX_FRAMERATE` and I think that's not what we need because I'm guessing that it defines the maximum framerate of the engine.

Game/Scripts/FinalBossScript.cpp
```diff
@@ -0,0 +1,321 @@
...
+	bool isAnAttackHappening = shockWaveAttackScript->IsAttacking() || chargeAttackScript->IsAttacking();
+
+	if (isAnAttackHappening)
+	{
+		return;
```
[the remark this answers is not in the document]
markmeliaa: Well thought, although not a big deal, any improvement in performance is a good improvement I guess.

Game/Scripts/FinalBossScript.cpp
```diff
@@ -0,0 +1,321 @@
...
+	{
+		ReactivateMovement();
+		patrolScript->RandomPatrolling(bossState != FinalBossStates::WALKING);
+		bossState = FinalBossStates::WALKING;
+	}
+}
+
+void FinalBossScript::ManageAggressivePhase()
+{
+	int chargeChance = rand() % 500; // Triple the chance of charges
```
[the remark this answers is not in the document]
markmeliaa: I could be if we ever need it, yes, but setting each random value and their increase in each phase would clutter too much the script's UI IMO, and hopefully we would never touch this value once properly balanced.

Game/Scripts/FinalBossScript.cpp
```diff
@@ -0,0 +1,321 @@
...
+	bool isAnAttackHappening = shockWaveAttackScript->IsAttacking() || chargeAttackScript->IsAttacking() ||
+		shieldAttackScript->IsAttacking();
+
+	if (isAnAttackHappening)
+	{
+		return;
+	}
+
+	// The boss is on the defensive now, if the shield attack is available, they will most likely trigger it
+	if (shieldChance < 5 && shieldAttackScript->CanPerformShieldAttack())
```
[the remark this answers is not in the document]
markmeliaa: True, it feels weird that every chance is checked when 0 and suddenly this is checked when less than 5, I swapped it to be a 1 in 200 chance.

Game/Scripts/BossChargeAttackScript.h
```diff
@@ -0,0 +1,74 @@
...
+	bool IsAttacking() const;
+
+private:
+	void ManageChargeAttackStates(float deltaTime);
+
+	void PrepareCharge() const;
+	void PerformChargeAttack();
+	void WallHitAfterCharge() const;
+
+	// These two functions will not be used in the first miniboss
```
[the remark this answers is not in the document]
markmeliaa: Good point, but IMO not worth the struggle because, as you say, it would only be used for another boss so scalability is not an issue here.

Game/Scripts/BossChargeAttackScript.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+void BossChargeAttackScript::Update(float deltaTime)
+{
+	ManageChargeAttackStates(deltaTime);
+}
+
+void BossChargeAttackScript::OnCollisionEnter(ComponentRigidBody* other)
+{
+	if ((other->GetOwner()->CompareTag("Wall") || other->GetOwner()->CompareTag("Rock"))
+		&& chargeState == ChargeState::CHARGING)
```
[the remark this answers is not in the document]
markmeliaa: I did not see any other case in which the first condition was more expensive than the second one, lmk if I missed any and I will change them too.

Game/Scripts/BossChargeAttackScript.cpp
markmeliaa: `rocksSpawned` is emptied when a new charge is performed, in this block of code the only thing that is being taken into account is that only the rocks that have spawned (still in the air) are enabled (if they are not enabled, that means that they got generated far too close to another rock, so they are still in the vector but deactivated). I don't really see the benefit on cleaning those rocks when they will get deleted anyway sooner or later, but maybe I am missing on something.

Game/Scripts/BossMissilesAttackScript.cpp
markmeliaa: Afaik, `SetTargetPosition` moves a `rigidBody` to a given position. If you call `SetTargetPosition` with `midJumpPosition`, you are guaranteed that it will always be reached (unless another `rigidBody` is on the way, which is improbable given that this will happen in the air and the arena is a flat surface so no obstacles should appear). Idk if that is what you really meant but if your concern is that maybe sometimes the `midJumpPosition` will not be reached because it will be skipped, that should never happen (again afaik).

Game/Scripts/BossShieldAttackScript.cpp
markmeliaa: Unlikely but eh, its a quick change, good catch!

Game/Scripts/BossShieldAttackScript.cpp
markmeliaa: That would be a nice addition! It wasn't really used until now (only in the PowerUps if I'm not mistaken) because there was nothing that needed to be random, but with addition of the final boss (and other bosses too) randomness will be more common (idk if it's already added to the project or not).

Game/Scripts/PatrolBehaviourScript.cpp
markmeliaa: Oh, yes! I forgot to check that! Luckily, @UriMurcia fixed it and now, when there is no elements in the vector, the transform of the parent will be the only waypoint that is taken into account for the patrol. Sooo, thanks Uri!

Game/Scripts/FinalBossScript.cpp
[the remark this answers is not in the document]
gerard-queralt: I see your point, thank you for the thorough justification! 😄

Game/Scripts/FinalBossScript.cpp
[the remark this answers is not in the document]
gerard-queralt: We don't have a framerate, but we do provide you a delta time, which is seconds per frame. So, if you do 1/deltaTime, you'll have frames per second.
Both questions are addressed by the recent ModuleRandom. You have access to a function that receives a probability and normalizes it based on the framerate and the max framerate (it assumes max framerate is your target). It also logs the random seed it's using, and provides a constructor that receives a seed, in case we need to test something. Let me know if you can use it and it works properly, there's a big difference from me checking that values look ok to you checking that they _feel_ ok.

Game/Scripts/BossShieldAttackScript.cpp
UriMurcia: Now we already have a module random, Gerard created it is very easy to use :)

Game/Scripts/BossChargeAttackScript.cpp
gerard-queralt: It's not a big deal, but it'd use less memory, since we don't care about the rocks anyway. Probably wouldn't even be noticed tbh, and we'd also need to consider if we care more about the memory usage or the cost of cleaning up the vector.
But, if I understood you properly, the process is:

1. We spawn the rocks
2. We check proximity, and disable accordingly
3. Rocks drop
4. Rocks despawn, back to step 1

If that's indeed how it works, wouldn't it be better to delete the rocks instead of disabling them?

Game/Scripts/BossShieldAttackScript.cpp
markmeliaa: Using it now!

Game/Scripts/FinalBossScript.cpp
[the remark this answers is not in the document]
markmeliaa: K! I'm gonna use it now!

Game/Scripts/BossChargeAttackScript.cpp
markmeliaa: That would be the way to go, indeed, but as I still have to add particle systems, idk if their deletion is fixed already, so I was waiting until I had to add particle systems to see if they get destroyed correctly. In case they get destroyed correctly, then I'll destroy them here without problem, but in case the particle systems are not yet destroyed correctly (which will be weird because I got told that it was fixed a couple weeks ago), then the only thing I'd have to do is destroy each particle system in the `DestroyRock` function of the `BossChargeRockScript.cpp` before destroying the rock itself.

I hope you understood the situation and the change will be performed eventually once the particles are tested and they are deleted correclty.

Game/Scripts/FinalBossScript.cpp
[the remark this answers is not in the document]
markmeliaa: Done! If you could take a look to see if I applied it correctly here's the commit https://github.com/Horizons-Games/Axolotl-Engine/commit/44e968e94afb83b8bd3c1e2f9f152b229d7001c3. I only changed the `rand()` related to this pr, as idk how the others work and may accidentaly break something (also changed the percentage to drop powerups that was set to always drop powerups but nvm that).

Game/Scripts/BossChargeAttackScript.cpp
gerard-queralt: Perfect, sorry on behalf of the engine side 🙇‍♂️

Game/Scripts/BossMissilesAttackScript.cpp
gerard-queralt: Yes, I was worried we were asking it to go to the landing position and checking that we hit this mid-jump one, which could have tunneling issues (a frame begins before the point, the next frame after, so the collision is lost). In hindsight, it makes sense that we first go to the midway position so we also increment/decrement the height, but at this point you already know I'd rather just ask away :P

PR #663, opened 2023-08-26
feat[graphics]: Variance Shadow Mapping

MacXxs:
Implemented variance shadow mapping using gaussian blur and Chebyshev inequality in order to  calculate shadow penumbra for non lit fragments.

**USE F4 TO ACTIVATE OR DEACTIVATE VSM**

|NO VSM|VSM|
|---|---|
|image|image|


AlexisDieu (approved):
Working in game also and find nothing wrong !

PR #667, opened 2023-08-29
feat: Debug Camera Frustum utilities

OriolSors:
- Fixed the Frustum position and orientation. Now you can move scene Camera objects and Frustum gets updated.
- You can select the frustum to use and cull the objects outside it.
- The frustum selected to use is also drawn if you check it.
- Only one frustum can be selected. If none of them are checked, the Engine Camera frustum will be the one selected.

### DIFF
--- a/Source/DataModels/Camera/CameraGameObject.cpp
+++ b/Source/DataModels/Camera/CameraGameObject.cpp
@@ -24,8 +24,6 @@ bool CameraGameObject::Init()
 	Camera::Init();
 	SetKpPosition(5.0f);
 	SetKpRotation(5.0f);
-	KpPosition = 5.0f;
-	KpRotation = 5.0f;
 	SetViewPlaneDistance(DEFAULT_GAMEOBJECT_FRUSTUM_DISTANCE);
 
 	return true;
--- a/Source/DataModels/Camera/CameraGameObject.cpp
+++ b/Source/DataModels/Camera/CameraGameObject.cpp
@@ -19,6 +19,18 @@ CameraGameObject::~CameraGameObject()
 {
 }
 
+bool CameraGameObject::Init()
+{
+	Camera::Init();
+	SetKpPosition(5.0f);
+	SetKpRotation(5.0f);
+	KpPosition = 5.0f;
+	KpRotation = 5.0f;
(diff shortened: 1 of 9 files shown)


Dedado3000 (dismissed):

MacXxs:
Hi Oriol, thanks for the changes!! 
I've found a bug though, if you have the frustum window opened while loading a scene the engine crashes:
image

Gonna keep looking for any possible bugs

MacXxs (changes requested):
The bug i commented just now 👀

MacXxs:
I've found another issue, the transform for the camera isn't updated with its parent's until the parent transform is modified.
20b7b671b6850768efc485395101826d

MacXxs:
Also, maybe we should have the capacity to modify the near and far planes for the added camera. By default is using the ones from the engine camera?

MacXxs:
> > Hi Oriol, thanks for the changes!! I've found a bug though, if you have the frustum window opened while loading a scene the engine crashes: image
> > Gonna keep looking for any possible bugs
> 
> Oh, does this happen before the changes? Can you check it?

Tried it in another branch and it does not happen!

OriolSors:
> I've found another issue, the transform for the camera isn't updated with its parent's until the parent transform is modified. 20b7b671b6850768efc485395101826d

I'm not sure but I think it's not an issue with the parent, but the frustum doesn't get updated until ontransformchanged is called. I can check it if it's an inconvenience

MacXxs:
> > > > Hi Oriol, thanks for the changes!! I've found a bug though, if you have the frustum window opened while loading a scene the engine crashes: image
> > > > Gonna keep looking for any possible bugs
> > > 
> > > 
> > > Oh, does this happen before the changes? Can you check it?
> > 
> > 
> > Tried it in another branch and it does not happen!
> 
> I have replicated it in develop... first time you load is not the problem. The problem appears when loading a scene once you have a scene loaded and the frustum window opened. So; I'm opening an issue

To me it happened loading a scene directly with the window opened, without having loaded one previously

OriolSors:
> Also, maybe we should have the capacity to modify the near and far planes for the added camera. By default is using the ones from the engine camera?

Done!

AuxPort66 (approved):

Dedado3000 (approved):

PR #670, opened 2023-08-29
Feat: Improved enemies

UriMurcia:
_To test this use the scene: **GameplayEnemiesTesting** (rememeber to **bake navigation** before testing; this won't be necessary to do everytime in the future)_

Features of this pull request:

- Refactored all the code from Venomite and Drone
- Patrol behaviour now has unlimited points
- Patrol behaviour now has stopped time on these points
- Enemy movement centralized in one script: AIMovement
- Now enemies have an alert state with an exclamation mark on their head when detecting the player
- Venomite now will only use the melee attack when the player is nearby the enemy
- Rangedattack now has a time before the shot to play a particle system to make it easier for the player to know when they are going to shoot
- Ranged attack now has a system to shoot in random periods of time to make it less repetitive
- VFX for Venomites and Drones
- Several fixes
- Use of the navigation agent by @totisama 

(Already changed every drone and venomite in Boss scene)


totisama (approved):

Game/Scripts/RangedFastAttackBehaviourScript.cpp
```diff
@@ -73,36 +141,82 @@ void RangedFastAttackBehaviourScript::PerformAttack()
...
 	script->Start();
 
-	lastAttackTime = SDL_GetTicks() / 1000.0f;
 	audioSource->PostEvent(AUDIO::SFX::NPC::DRON::SHOT_01);
 
+	currentConsecutiveShots += 1.0f;
+	if (currentConsecutiveShots < numConsecutiveShots)
+	{
+		isConsecutiveShooting = true;
+		nextShotDuration = minTimeConsecutiveShot + static_cast<float>(std::rand()) / (static_cast <float>(RAND_MAX /(maxTimeConsecutiveShot - minTimeConsecutiveShot)));
```
gerard-queralt, line 155: Do we need the bigger cast to float?
```suggestion
		nextShotDuration = minTimeConsecutiveShot + std::rand() / (static_cast <float>(RAND_MAX /(maxTimeConsecutiveShot - minTimeConsecutiveShot)));
```
Also, maybe it's just me, but I'm not sure what exactly we're doing here. Looks like we want a random number in the range [minTimeConsecutiveShot, maxTimeConsecutiveShot]? If that's the case, this looks excessively complex, why not just
```suggestion
		nextShotDuration = minTimeConsecutiveShot + std::rand() % (maxTimeConsecutiveShot - minTimeConsecutiveShot + 1);
```

Game/Scripts/RangedFastAttackBehaviourScript.cpp
```diff
@@ -73,36 +141,82 @@ void RangedFastAttackBehaviourScript::PerformAttack()
...
+		isWaitingForConsecutiveShot = true;
+	}
+	else
+	{
+		isConsecutiveShooting = false;
+		currentConsecutiveShots = 0.0f;
+		lastAttackTime = SDL_GetTicks() / 1000.0f;
```
gerard-queralt, line 162: Use delta time :P

Source/DataModels/Components/ComponentAgent.cpp
```diff
@@ -16,6 +16,7 @@
 ComponentAgent::ComponentAgent(bool active, GameObject* owner) :
 	Component(ComponentType::AGENT, active, owner, true)
 {
+	transform = GetOwner()->GetComponentInternal<ComponentTransform>();
```
gerard-queralt, line 19:
```suggestion
	transform = GetOwner()->GetComponent<ComponentTransform>();

```

gerard-queralt (commented):
Code looks great overall, but I'd appreciate if you could stick to using `GetComponent` instead of `GetComponentInternal`. Not only to centralize how the scripts access a GameObject's components (only `GetComponent` needs to be updated in case something is needed, and we don't break engine code), but also because, should we manage to get hot reloading to work (haven't given up yet), that method won't be available in the gameplay project.
Less important, I encourage to use the delta time provided by the engine instead of SDL methods.

UriMurcia:
> Code looks great overall, but I'd appreciate if you could stick to using `GetComponent` instead of `GetComponentInternal`. Not only to centralize how the scripts access a GameObject's components (only `GetComponent` needs to be updated in case something is needed, and we don't break engine code), but also because, should we manage to get hot reloading to work (haven't given up yet), that method won't be available in the gameplay project. Less important, I encourage to use the delta time provided by the engine instead of SDL methods.

Didn't see any GetComponentInternal, probably copied and pasted xdd, I'm going to change it for sure, thanks!

AuxPort66 (approved):

PR #672, opened 2023-08-30
feat : Graphics/script&perf

AlexisDieu:
- Add a flag to turn on / off debugDraw of RigidBody to gain some performance when playing in engine mode
- Move the ComponentLine script to animate the line in a new script "AnimatedLine"


PaablooCH (approved):
👍

PR #681, opened 2023-08-31
feat: accessing a script that is not running raises an exception

gerard-queralt:
Currently, if a script stops running, you can still get it using `GetComponent`, which might cause an error since the script has no reason to account for that. With this change such cases will raise a separate exception.


Piterga12 (dismissed):

markmeliaa (dismissed):

Piterga12 (approved):

MarcGonzalezMoratona (approved):

Source/Exceptions/AxoDefaultException.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include <exception>
+
+class AxoDefaultException : public std::exception
+{
+public:
+	AxoDefaultException(const std::string& message);
+	~AxoDefaultException() override = default;
+
+	const char* what() const noexcept override;
```
MarcGonzalezMoratona, line 11:
```suggestion
	const char* What() const noexcept override;
```

should not this be What()?

Source/DataModels/Components/ComponentScript.cpp
```diff
@@ -52,6 +53,14 @@ void RunScriptMethodAndHandleException(bool& scriptFailedState, ComponentScript*
 				  exception.what());
 		scriptFailedState = true;
 	}
+	catch (const AccessingFailedScriptException& exception)
+	{
+		LOG_ERROR("Script {}, owned by {} attempted to access script that has stopped running. Error message: {}",
```
MarcGonzalezMoratona, line 58:
```suggestion
		LOG_ERROR("Script {}, owned by {} attempted to access a script that has stopped running. Error message: {}",
```

MarcGonzalezMoratona: you can omit this suggestion if it's too picky!

MarcGonzalezMoratona (approved):

MarcGonzalezMoratona:
I only checked the code, I suppose you tested it in order to make this functionality

Source/Exceptions/AxoDefaultException.h
gerard-queralt: Notice that this is an override, we're extending std::exception: https://en.cppreference.com/w/cpp/error/exception/what

Source/DataModels/Components/ComponentScript.cpp
gerard-queralt: Picky comments are my speciality :P

Piterga12 (approved):

Source/Exceptions/AxoDefaultException.h
MarcGonzalezMoratona: okay, now I understand it. I didn't see that override. thank you!

markmeliaa (approved):
LGTM!

Piterga12 (approved):

PR #683, opened 2023-09-02
Feat: Nav mesh load&save

AuxPort66:
Note that you need to **save the scene** so that NavMesh is saved in the json.

- Load/Save navMesh updated with the result of the bake so we don't bake all the time
- Updated Scene to load the navMesh
- Fix some bugs related to WindowNavigation
- Moved the CreateNavMesh Button to WindowNavigation
- Updated the Premade


Source/DataModels/Resources/ResourceNavMesh.cpp
```diff
@@ -529,7 +529,94 @@ void DrawObstacles(duDebugDraw* dd, const dtTileCache* tc)
 	}
 }
 
- bool ResourceNavMesh::Build(Scene* scene)
+ bool ResourceNavMesh::InitNavMesh(const float* bmin)
+{
+	CleanUp();
+
+	talloc = new LinearAllocator(32000);
```
gerard-queralt, line 536: Could be missing something, but I think we're not freeing these three pointers. We'd need to do so even in case of error, before we return.

AuxPort66: It's in draft because I need to check the cleanUp

totisama (approved):

diegusmejia117 (approved):

OriolSors (approved):

PR #688, opened 2023-09-04
feat: controller input for UI buttons

MarcGonzalezMoratona:
Now, you can easily navigate the UI using a gamepad.
We've also introduced the option to play in fullscreen mode while in engine play mode.
You can now toggle the editor controls on or off in engine play mode.
Some missing windows are now automatically docked.


Piterga12 (approved):
Works great, good job!

Source/Modules/ModuleEditor.cpp
```diff
@@ -251,7 +253,20 @@ UpdateStatus ModuleEditor::Update()
...
+		if (fullscreenScene)
+		{
+			if (editorControls)
+			{
+				bool isEditorControl = windows[i]->GetName() == "Editor Control";
+				windows[i]->Draw(isEditorControl);
+			}
+			bool isSceneWindow = windows[i]->GetName() == "Scene";
+			windows[i]->Draw(isSceneWindow);
+		}
```
gerard-queralt, line 265: Just to make sure I understand. If the scene is full screen, we only draw the scene, and if we want to see the editor controlls we draw them as well. Is that right?
Few picky things:
- Instead of using the hardcoded strings here, I prefer to declare them at the top of the source file in an unnamed namespace, makes it easier to change them when needed. You can check what I did in the [loading screen script](https://github.com/Horizons-Games/Axolotl-Engine/blob/GamePlay/loading-screen/Game/Scripts/SceneLoadingScript.cpp#L17)
- If `editorControls` is `true`, you're making two calls to `Draw`. Harmless, but unnecessary. I'd just compute a single condition and make a single call
```c++
if (fullscreenScene)
{
    bool canBeDrawn = windows[i]->GetName() == windowSceneName;
    if (editorControls)
    {
        canBeDrawn = canBeDrawn && windows[i]->GetName() == windowEditorControlsName;
    }
    windows[i]->Draw(canBeDrawn);
}
```
Or something along those lines.
- You can read/write the `windowEnabled` value only in the `else` block. In fact, you could get rid of the else block if you added a `continue` inside the first `if`, but that's up to you.

Source/Modules/ModuleEditor.h
```diff
@@ -40,6 +40,12 @@ class ModuleEditor : public Module
 	void RefreshInspector() const;
 	std::pair<float, float> GetAvailableRegion();
 
+	bool GetFullscreenScene() const;
```
gerard-queralt, line 43: I don't like using "get" for boolean getters, I think "is" or "has" feel more natural. You're better with naming than I am, so up to you to rename it. Same for the editor controls one. Both feel like you're returning the actual window.

Source/Modules/ModuleEditor.h
```diff
@@ -79,3 +88,23 @@ inline const WindowDebug* ModuleEditor::GetDebugOptions() const
...
+
+inline bool ModuleEditor::GetFullscreenScene() const
+{
+	return fullscreenScene;
+}
+
+inline bool ModuleEditor::GetEditorControls() const
+{
+	return editorControls;
+}
```
gerard-queralt, line 110: Newline at EOF please :P

Source/Modules/ModuleUI.cpp
```diff
@@ -25,9 +25,22 @@
 	#include "DataModels/Windows/EditorWindows/WindowScene.h"
 #endif // ENGINE
 
-ModuleUI::ModuleUI(){};
+namespace
+{
+bool compareButtonPositions(const ComponentButton* a, const ComponentButton* b)
```
gerard-queralt, line 30:
```suggestion
bool CompareButtonPositions(const ComponentButton* a, const ComponentButton* b)
```

Source/Modules/ModuleUI.cpp
```diff
@@ -49,9 +62,27 @@ UpdateStatus ModuleUI::Update()
 
 	bool leftClickDown = input->GetMouseButton(SDL_BUTTON_LEFT) == KeyState::DOWN;
 
+	if (!sortedButtonsIds.empty() && SDL_GetTicks() / 1000.0f > lastButtonChange + 0.2f)
```
gerard-queralt, line 65: Don't use `SDL_GetTicks()`, prefer using `App->GetDeltaTime()`.

gerard-queralt: Also, is that 0.2f supposed to be the cooldown? If so, I'd also add it to an unnamed namespace, same as my previous comment.

Source/Modules/ModuleUI.h
```diff
@@ -21,18 +23,32 @@ class ModuleUI : public Module
...
 
 inline unsigned int ModuleUI::GetQuadVAO() const
 {
 	return quadVAO;
 }
+
+inline void ModuleUI::ClearButtons()
+{
+	sortedButtonsIds.clear();
+}
```
gerard-queralt, line 54: Newline please 😇

gerard-queralt (commented):
Great job Marc, just a few picky comments but nothing important 😉

Source/Modules/ModuleUI.cpp
MarcGonzalezMoratona: `App->GetDeltaTime()` returns de time between two frames, but here I need the amount of time since the last button change. How would you do this? I tried to replace `SDL_GetTicks() / 1000.0f` with `App->GetDeltaTime()` but then the functionality does not work

Vlukkee:
I was testing the branch in the main menu and its options, these are my conclusions:

- In the main menu it works as such, although sometimes going up and down requires you to do the movement twice. I couldn't find out why.

- When I enter a menu option it is not possible to return and it does not recognize the return button.

- The mouse is disabled, I don't know how complicated it is but would it be possible to focus the controller or the mouse depending on which one is moving?

- The in game menu does not work, I have tested it in gameplayfulltrial.

MarcGonzalezMoratona:
> I was testing the branch in the main menu and its options, these are my conclusions:
> 
> * In the main menu it works as such, although sometimes going up and down requires you to do the movement twice. I couldn't find out why.
> * When I enter a menu option it is not possible to return and it does not recognize the return button.
> * The mouse is disabled, I don't know how complicated it is but would it be possible to focus the controller or the mouse depending on which one is moving?
> * The in game menu does not work, I have tested it in gameplayfulltrial.

- There is a cooldown of 0.2 seconds. We can tweak these values to make it feel more acessible.
- I'm working on the back buttons of the menus, I'll let you know when this part is finished before we merge it.
- I made it on purpose to avoid mixing the controls. The mouse works only when the controller is disconnected. We can have both at the same time but then we can have a button hovered by the mouse and another hovered by the gamepad at the same time (I don't know which one would be triggered if we click on the mouse left button when two buttons are hovered at the same time). So let me know what do you prefer, using the mouse and controller independently or using them at the same time knowing this possible issue. Changing between a solution or another by code is really easy and fast to do. We can meet together to discuss this if you prefer doing so.
- I haven't worked yet on the in-game menu. Can we do this part in another PR?

Vlukkee:
>     * There is a cooldown of 0.2 seconds. We can tweak these values to make it more accessible.
For me it is fine but I think it can be improved, if it is a value leave it and we will change it in the future. Right now I couldn't tell you exactly how to improve it and we will have more controller updates.


>     * I'm working on the back buttons of the menus, I'll let you know when this part is finished before we merge it.
Ok, perfect!

>     * I made it on purpose to avoid mixing the controls. The mouse works only when the controller is disconnected. We can have both at the same time but then we can have a button hovered by the mouse and another hovered by the gamepad at the same time (I don't know which one would be triggered if we click on the mouse left button when two buttons are hovered at the same time). So let me know what do you prefer, using the mouse and controller independently or using them at the same time knowing this possible issue. Changing between a solution or another by code is really easy and fast to do. We can meet together to discuss this if you prefer doing so.
Maybe what I say sounds crazy, but you tell me if it's possible or not. In Diablo 2 or 4, when you have a controller and a mouse, both work but only 1 at a time, they become focused when you press a key or move the mouse and in turn when you touch something on the controller it becomes focused.
I know I'm talking about the Diablo but maybe it is possible and not so hard.

>     * I haven't worked yet on the in-game menu. Can we do this part in another PR?
Oh i didnt know it, ok.

MarcGonzalezMoratona:
> > ```
> > * There is a cooldown of 0.2 seconds. We can tweak these values to make it more accessible.
> > ```
> 
> For me it is fine but I think it can be improved, if it is a value leave it and we will change it in the future. Right now I couldn't tell you exactly how to improve it and we will have more controller updates.
> 
I'm trying with 0.15 seconds and it looks better. 0.1 is too fast, so the optimal value should be near. I will update it soon with the new value and let you test it.
> > ```
> > * I'm working on the back buttons of the menus, I'll let you know when this part is finished before we merge it.
> > ```
> 
> Ok, perfect!
> 
> > ```
> > * I made it on purpose to avoid mixing the controls. The mouse works only when the controller is disconnected. We can have both at the same time but then we can have a button hovered by the mouse and another hovered by the gamepad at the same time (I don't know which one would be triggered if we click on the mouse left button when two buttons are hovered at the same time). So let me know what do you prefer, using the mouse and controller independently or using them at the same time knowing this possible issue. Changing between a solution or another by code is really easy and fast to do. We can meet together to discuss this if you prefer doing so.
> > ```
> 
> Maybe what I say sounds crazy, but you tell me if it's possible or not. In Diablo 2 or 4, when you have a controller and a mouse, both work but only 1 at a time, they become focused when you press a key or move the mouse and in turn when you touch something on the controller it becomes focused. I know I'm talking about the Diablo but maybe it is possible and not so hard.
> 
I have also seen this in other games but I think this is a bit more difficult, so it will require more time. It's achievable but I would do it later. Probably the teachers will test it only with a controller for this milestone
> > ```
> > * I haven't worked yet on the in-game menu. Can we do this part in another PR?
> > ```
> 
> Oh i didnt know it, ok.

Vlukkee (dismissed):
Works good for me! Nice job!

Vlukkee (dismissed):

Source/Modules/ModuleUI.cpp
gerard-queralt: You need a bit of refactoring. You'd have to accumulate the delta time in a member of the class, and then check it each frame.

Source/Modules/ModuleEditor.cpp
gerard-queralt: You're right, the code I provided is wrong, but you only need to change the AND for an OR and it should work.

MarcGonzalezMoratona: Solved, thank you!

Source/Modules/ModuleUI.cpp
MarcGonzalezMoratona: Done!

gerard-queralt (approved):
Looks and feels awesome, great job!

Vlukkee (dismissed):

Vlukkee (approved):

Piterga12 (approved):

PR #693, opened 2023-09-06
fix : SpotLight was not using Light Culling correctly

AlexisDieu:
As the title said, after verification in Nsight it was not done correctly


MacXxs:
Hi Alexis, this isn't working correctly.  The reason is that the points that are not directly in the aim of the direction of the spotlight are at a higher distance, which in the case of points near the end of the radius of the spot that are not exactly at the center of the aim won't show up ('cause they are at a higher distance than the radius of the spot, but are inside the spot), which is wrong.
image

MacXxs:
Actually reading the pdf this isn't doing light culling as inteded. We should do it in a separated light pass for local lights (after deferred) where we use light volumes to get the pixels in screen that need to be taken into account for being treated during this pass.

image
image

PR #702, opened 2023-09-09
fix: rigidbody duplicates and possibility to remove from simulation

OriolSors:
- Some calls to dynamics world cleared
- Functions added to Quadtree for adding/removing the RigidBody from simulation
- Commented code as example for culling RigidBodies in the FillRenderList (worse performance; need to be treated wisely)


totisama (dismissed):

totisama (approved):

Piterga12 (approved):
Works well!

PR #707, opened 2023-09-11
Fix: Bix and Allura Normal Combo Attack Animation State Machine

UriMurcia:
Modified the normal attack system (Light attack and heavy attack) from Bix and Allura to make 3 consecutive animations to make the attack combo system less repetitive.

I also did:
- Added a getter in the animation controller due to the state machine not instantly triggering new animations, this way we can detect the name of the new state on the next frame, which not always happened with the animation->GetActualStateName().
- Changed attack input for single click instead of constantly checking if mouse is pressed
- Fixed Boost Of Energy Target error


Game/Scripts/PlayerAttackScript.cpp
```diff
@@ -128,41 +132,105 @@ void PlayerAttackScript::UpdateEnemyDetection()
...
+		switch (currentAttack)
+		{
+			case AttackType::LIGHTNORMAL:
+			case AttackType::HEAVYNORMAL:
+				if (!isNextAttackTriggered) //Calling only when is not currently attacking
+				{
+					LOG_VERBOSE("Normal Normal Attack Soft");
+					numAttackComboAnimation = 0.0f;
+					animation->SetParameter("NumAttackCombo", numAttackComboAnimation);
+					LightNormalAttack();
```
AuxPort66, line 196: At first when I made the combo system the heavy and light attacks were together in the same function with a couple of checks and they were later separated in a code review (Not for me)

It doesn't seem bad to me that you put them together but if you call the LightAttackFunction all the time without checking the real input or diference anything you are breaking the UI and the comboCount, basically now if you hit left click or right click always appears X in the UI and it always counts the same, this is a bug. You can separate in all the switches this two attacks or put the old checks in the NormalAttack Function

AuxPort66 (changes requested):

Game/Scripts/PlayerAttackScript.cpp
UriMurcia: True, didn't realise of that

markmeliaa (approved):
GJ Uri, the animations looks better now!

AuxPort66 (approved):

PR #709, opened 2023-09-11
feat: First version Hacking

OriolSors:
Here we go with hacking ability! Main features:

### Hacking Zone
When the player reaches a certain influence area (GO with Rb + HackZone script), there is the possibility to start the hacking by pressing 'H'. 
- Auto-generated combination of buttons each time you start the hacking.
- Time limit to complete.
- Config options: influence radius, max time to complete and hacking sequence size.
  - Due to UI limitations, now there is a max sequence of 3 (the number of placeholders in the Hud, but can be easily extendable).

### Hacking Combination
- The list of hacking inputs is in HackingCommand script along with the proper mappings to user input type. 
- The UIHackingManager is inspired by the combat one, having the GOs of the commands there, linked to the hacking command mentioned. 
  - Now only two options can be linked (X,Y). UI should incorporate the rest of the controller button images.
- So, if any key/button needs to be readjusted the only thing to do is add the mapping and the GOs to the UI script.

### Steps to test the hacking

1. Copy/Paste the CombatSystem UI element in the hierarchy, delete the Bar object and leave the same enabled/disabled states of the children as the original CombatSystem. Replace the script of UICombo for the UIHackingManager and drag the placeholders (X, Y).
2. Add PlayerHackingUseScript to the character and attach the duplicated UI element above.
3. Create a GameObject in the scene with a RigidBody and the HackZone script.
4. Play with Keyboard or Controller, but for enabling the hacking use the "H" key. Then the combination will appear as buttons, but "translated" in keys in the engine console (if you want to test it with keyboard).
 
### Missing functionalities
- UI should be improved, with more images of the buttons and a proper canvas to display a sequence as long as we want.
- Enabling/Disabling the environment. ComponentAgents should be all disabled as well as Player's RigidBody.


AuxPort66 (changes requested):
You say the input is limited to 3, but I try it a couple of times and I have 4 inputs and when I put it all then UI explodes. So I can't reach a succesfull hacking right know.

Apart of this, we need the limits on the player, like don't moving or attacking when I'm hacking. And probably when we fail restart automatically the hacking.

OriolSors:
> You say the input is limited to 3, but I try it a couple of times and I have 4 inputs and when I put it all then UI explodes. So I can't reach a succesfull hacking right know.
> Apart of this, we need the limits on the player, like don't moving or attacking when I'm hacking. And probably when we fail restart automatically the hacking.



> You say the input is limited to 3, but I try it a couple of times and I have 4 inputs and when I put it all then UI explodes. So I can't reach a succesfull hacking right know.
> 
> Apart of this, we need the limits on the player, like don't moving or attacking when I'm hacking. And probably when we fail restart automatically the hacking.

The sequence size can be modified in each HackZoneScript, so you can put it there as 3.

Okey, I will block the player for now and try to auto-restart, thankss

Game/Scripts/UIHackingManager.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+	inputPositions.push_back(owner->GetChildren()[0]->GetChildren()[0]);
+	inputPositions.push_back(owner->GetChildren()[0]->GetChildren()[1]);
+	inputPositions.push_back(owner->GetChildren()[0]->GetChildren()[2]);
+}
+
+void UIHackingManager::Update(float deltaTime)
+{
+	
+}
```
markmeliaa, line 31: You can delete this function and its declaration in the header if it's not being used.

markmeliaa (approved):
GJ Oriol, the hacking works great for its first approach (and we'll work on it looking well too). I pushed a commit setting the hacking with the steps you described in the description, so other people don't have to reproduce them as well, I hope you don't mind.

The only thing that I'm missing is the visuals, idk if I did not adapt the combo buttons correctly or if it is not yet implemented because we need new buttons, but in any case, the X/Y buttons are not being shown correctly to me in the UI.

Game/Scripts/UIHackingManager.cpp
OriolSors: yes, forgot thankss

OriolSors:
> GJ Oriol, the hacking works great for its first approach (and we'll work on it looking well too). I pushed a commit setting the hacking with the steps you described in the description, so other people don't have to reproduce them as well, I hope you don't mind.
> 
> The only thing that I'm missing is the visuals, idk if I did not adapt the combo buttons correctly or if it is not yet implemented because we need new buttons, but in any case, the X/Y buttons are not being shown correctly to me in the UI.

I tried in the level scene, copying the ComboSystem UI element. Maybe in other scene some elements are missing and that causes not showing the buttons properly... I will try to the gameplay scene to know better what it happens!

markmeliaa (approved):
Works perfect now Oriol, thanks for your job!

AuxPort66 (approved):

PR #722, opened 2023-09-15
refactor: add lateral dash and fix infinite dash error

MarcGonzalezMoratona:
Refactored playerMoveScript to improve code readability, removed unnecessary booleans and checks, simplified logic.
Dash stops after a certain number of ms. Configurable via scripts.
DashForce can be also configured with scripts.
Ability to dash laterally using the right joystick.
Cooldown decreased and can be configured.


UriMurcia (commented):
Marc teacher said it would be interesting to have no cooldown in the dash as it would maybe feel funnier, as well he said that maybe being inmune while dashing was a good idea. Should we implement that?

UriMurcia (changes requested):
Right now the dash can be performed in the air, which I think is a good idea, but it's only working while falling down (not with the going up animation, double jump), as well that it's not doing the dash animation when this happens.
Dashing while perform the Jump attack is making the player fly to the moon, I think we should disable the dash while performing this attack.
Also, I think that teachers said that they wanted to dash lateraly while attacking, but right now if you, for example, are attacking an enemy with the joystick aiming to another direction, the dash will be performed towards the enemy.

MarcGonzalezMoratona:
> -Right now the dash can be performed in the air, which I think is a good idea, but it's only working while falling down (not with the going up animation, double jump), as well that it's not doing the dash animation when this happens. -Dashing while perform the Jump attack is making the player fly to the moon, I think we should disable the dash while performing this attack. -Also, I think that teachers said that they wanted to dash lateraly while attacking, but right now, for example, if you are attacking an enemy with the joystick aiming to another direction, the dash will be performed towards the enemy.

I'll check this, thank you!

MarcGonzalezMoratona:
> Marc teacher said it would be interesting to have no cooldown in the dash as it would maybe feel funnier, as well he said that maybe being inmune while dashing was a good idea. Should we implement that?

Removing the cooldown is easy, the other part I don't know it yet

UriMurcia:
> > Marc teacher said it would be interesting to have no cooldown in the dash as it would maybe feel funnier, as well he said that maybe being inmune while dashing was a good idea. Should we implement that?
> 
> Removing the cooldown is easy, the other part I don't know it yet

Just control in the health that the player takes damage if it's not dashing, it should be pretty easy.

diegusmejia117:
Building on Uri's comment, you can actually patch into the player's health system immortal setting, and just flip it on, that way the player shoudn't take any damage while dashing

UriMurcia:
> > Marc teacher said it would be interesting to have no cooldown in the dash as it would maybe feel funnier, as well he said that maybe being inmune while dashing was a good idea. Should we implement that?
> 
> Removing the cooldown is easy, the other part I don't know it yet

Bu the way, removing the cooldown is not only removing the cooldown, you also need to check that you are not dashing, to do that check the current animation is different from the dash

markmeliaa (changes requested):
Hi Marc! First of all, thanks for your work in this pr, code looks good, let's talk about the funcionality.

The first thing that I noticed is that the dash changed from how it was when you and Peter developed it, idk if something changed during this past weeks, but it used to be instant and now it seems like you are just sliding on the floor (do you remember the reference about Jett? this is much different to how it was) what happened?

About Uri's comments, I mostly agree with all of them, dash should only be performed when we are not attacking (while attacking disable dash and vice versa), and a lateral dash would look really cool indeed!
And regarding the teacher's feedback, having the dash without cooldown would make the game more dynamic, I would very much like that, but if we accomplish to make the dash to how it was before, because now it looks really weird imo. Idk about the immortality/not take damage while dashing, idk if the combination of dashing without cooldown and the dash making you invincible may make the game "weird".

P.S. If you don't remember correcly what I mean with the "old dash", I've rescued a branch (called it `GamePlay/old-dash`) and there, in the `Gameplay_FinalBoss` scene you could check how it should ideally work (and all the animations working accordingly and in order), any doubt you may have feel free to contact me!

MarcGonzalezMoratona:
> Hi Marc! First of all, thanks for your work in this pr, code looks good, let's talk about the funcionality.
> 
> The first thing that I noticed is that the dash changed from how it was when you and Peter developed it, idk if something changed during this past weeks, but it used to be instant and now it seems like you are just sliding on the floor (do you remember the reference about Jett? this is much different to how it was) what happened?
> 
> About Uri's comments, I mostly agree with all of them, dash should only be performed when we are not attacking (while attacking disable dash and vice versa), and a lateral dash would look really cool indeed! And regarding the teacher's feedback, having the dash without cooldown would make the game more dynamic, I would very much like that, but if we accomplish to make the dash to how it was before, because now it looks really weird imo. Idk about the immortality/not take damage while dashing, idk if the combination of dashing without cooldown and the dash making you invincible may make the game "weird".
> 
> P.S. If you don't remember correcly what I mean with the "old dash", I've rescued a branch (called it `GamePlay/old-dash`) and there, in the `Gameplay_FinalBoss` scene you could check how it should ideally work (and all the animations working accordingly and in order), any doubt you may have feel free to contact me!

Hi Mark! In order to solve the infinite dash issue, I had to refactor the file, and to remove a linear velocity that probably is the cause of the behaviour you experienced. I'll try my best to have a dash with the same feeling using an impulse.

I'm working on the lateral dash. I decided to keep the PR in draft until I make the lateral dash, although I usually prefer to develop features in independent branches. The cooldown has been changed to 0.1s, which is the minimum I achieved to avoid spamming the key, because not having any cooldown was breaking Allura's animations. The teachers only asked us to make the cooldown shorter, not to remove it completely (at least this is what I have been told).

About the combination of dash with low cooldown and immortality during dash, this is something we should test and polish by playing.

Thank you for the branch reference, I'll take it to check what has happened.

markmeliaa (changes requested):
Thanks again for your work, Marc, this looks better than before.
However, there are some aspects that I feel that must be addressed before merging this PR:

- First of all, the animation of the dash is completely out of sync. IIRC, there are three animations involving the dash, `DashingInit`, `DashingKeep` and `DashingEng`, the expected behaviour when the dash key is pressed should be: 
    - Stop input from the player because a dash is going to take place
    - Play the `DashingInit` animation
    - Once the `DashingInit` animation is played, play the `DashingKeep` animation at the same time that the dash functionality is triggered (the actual dash happens)
    - After the dash has happened and the player has moved the desired distance, play the `DashingEnd` animation
    - When the `DashingEnd` animation has stopped, accept movement/attack input from the player again.

   That is how it was done in the past, and how is shown in the branch that I told you, so we should aim for a behaviour similar to that.

- The dash got completely disabled when jumping, why was that? Was anything causing an error that you could not fix? Although when you dash you should not be able to jump, when you jump you should be able to dash, giving the player more mobility.

- I like the feeling of having less cooldown on the dash, makes the game more dynamic and fun, gj with that!

markmeliaa:
Also I tried playing the level 1 with a controller and Bix never stops moving after I move the joystick of the controller towards the right unless I make her go left strongly, so idk if this may be a "me" problem but it's weird enough to bring it up.

markmeliaa:
And if you could direct this branch to `game-programming` instead of to `develop`, as this is related to gameplay and that way we keep this kind of stuff in that branch.

MarcGonzalezMoratona:
> Thanks again for your work, Marc, this looks better than before. However, there are some aspects that I feel that must be addressed before merging this PR:
> 
> * First of all, the animation of the dash is completely out of sync. IIRC, there are three animations involving the dash, `DashingInit`, `DashingKeep` and `DashingEng`, the expected behaviour when the dash key is pressed should be:
>   
>   * Stop input from the player because a dash is going to take place
>   * Play the `DashingInit` animation
>   * Once the `DashingInit` animation is played, play the `DashingKeep` animation at the same time that the dash functionality is triggered (the actual dash happens)
>   * After the dash has happened and the player has moved the desired distance, play the `DashingEnd` animation
>   * When the `DashingEnd` animation has stopped, accept movement/attack input from the player again.
>   
>   That is how it was done in the past, and how is shown in the branch that I prepared for you (`GamePlay/old-dash`), so we should aim for a behaviour similar to that.

I checked the state machine with Diego and we made some changes in order to sync the animations with the dash, but probably the result is not what you expected... I'll check it again

> * The dash got completely disabled when jumping, why was that? Was anything causing an error that you could not fix? Although when you dash you should not be able to jump, when you jump you should be able to dash, giving the player more mobility.

I was told to disable the dash when jumping because it would break the playability of the parkour zones. 

> * I like the feeling of having less cooldown on the dash, makes the game more dynamic and fun, gj with that!

Thank you Mark!

MarcGonzalezMoratona:
> Also I tried playing the level 1 with a controller and Bix never stops moving after I move the joystick of the controller towards the right unless I make her go left strongly, so idk if this may be a "me" problem but it's weird enough to bring it up.

This can be a drift problem, so I suggest you to modify the values of 3200 that are inside moduleInput.cpp in order to see if this is solved by increasing the value.

These values go between 0 and 32767, 10% is what SDL recommended, but it can vary from a controller to another, so try 20% or 30% and let's see what happens. Uri explained me that something similar was happening to him, so maybe we can find a value that fits better all the controllers. This is something I can not test

markmeliaa:
> I checked the state machine with Diego and we made some changes in order to sync the animations with the dash, but probably the result is not what you expected... I'll check it again

Thanks!  Right now it just feels like the animations are happening without concordance with the functionality, and that bothers me.

> I was told to disable the dash when jumping because it would break the playability of the parkour zones.

Okay then, idk how it would break the playability having a dash you could use or not use, but if you were told that then np!

> This can be a drift problem, so I suggest you to modify the values of 3200 that are inside moduleInput.cpp in order to see if this is solved by increasing the value.

Hopefully, it's not a controller problem because in other games I use it, the movement works well and it stays on point with the inputs I give, so not a drift from that part. I'll try to tweak the values whenever I try the game again, but knowing its just a me problem then I won't worry about it!

markmeliaa (dismissed):
GJ Marc, congrats!

Piterga12 (approved):

AuxPort66:
> Okay then, idk how it would break the playability having a dash you could use or not use, but if you were told that then np!

I told him we can't make any changes to mobility at this point. If you have a dash in the air, all the jumping puzzles with the force box are useless, It's a lost design. These types of changes in mobility when you have parkour areas must be had from the beginning, so we can't have it.

AuxPort66:
The stateMachine explodes when I receive damage in Allura, I have checked and in develop it does not happen.

markmeliaa:
> I told him we can't make any changes to mobility at this point. If you have a dash in the air, all the jumping puzzles with the force box are useless, It's a lost design. These types of changes in mobility when you have parkour areas must be had from the beginning, so we can't have it right now even if it's cool or better sensation, breaks other things. Also in Allura doesn't have any sense, the animation is a roll. We can't change that animations and the parkour at this point

Yea, Marc explained to me the situation after posting that message, it has sense why it's removed now. The only thing that "bothered me" (not a lot) is why it wasn't thought by who designed the parkout that the dash could be used while jumping, because we knew that the player would have a dash since day 1, but yes, as you said, there is nothing we could change now, and that's alright.

markmeliaa (approved):

Piterga12 (approved):

AuxPort66:
> Yea, Marc explained to me the situation after posting that message, it has sense why it's removed now. The only thing that "bothered me" (not a lot) is why it wasn't thought by who designed the parkout that the dash could be used while jumping, because we knew that the player would have a dash since day 1, but yes, as you said, there is nothing we could change now, and that's alright.

Well, if the dash was intended to be used in the air it clearly wasn't specified properly because even the Allura animation is a roll...

markmeliaa:
> Well, if the dash was intended to be used in the air it clearly wasn't specified properly because even the Allura animation is a roll...

Afaik we only have parkour in the first level, and the first level is Bix only, so no roll ;)

AuxPort66 (approved):

markmeliaa (approved):

PR #724, opened 2023-09-15
fix: controller input

MarcGonzalezMoratona:
Removed componentPlayerInput, now everything is checked inside moduleInput (this fixes the issue with unconsistent controller states)
Added controller input methods for right joystick (but we don't use in scripts yet)
@DADSILVER implemented 360º movement for walking 
360º dash implemented too


Source/Modules/ModuleInput.cpp
```diff
@@ -62,16 +89,12 @@ bool ModuleInput::Init()
...
 {
-#ifdef DEBUG
-	OPTICK_CATEGORY("UpdateInput", Optick::Category::Input);
-#endif // DEBUG
-
 	mouseMotion = float2::zero;
 	mouseWheelScrolled = false;
 
-	for (int i = 0; i < SDL_NUM_SCANCODES; ++i)
+for (int i = 0; i < SDL_NUM_SCANCODES; ++i)
```
gerard-queralt, line 97: Fix indentation.

Source/Modules/ModuleInput.cpp
```diff
@@ -282,6 +304,17 @@ UpdateStatus ModuleInput::Update()
...
 
+	MapInput();
+
+	return UpdateStatus::UPDATE_CONTINUE;
+}
+
+UpdateStatus ModuleInput::Update()
+{
+#ifdef DEBUG
+	OPTICK_CATEGORY("UpdateInput", Optick::Category::Input);
```
gerard-queralt, line 315: Might be worth adding another optick mark at the beggining of the preupdate, just in case.

Source/Modules/ModuleInput.cpp
```diff
@@ -366,3 +399,29 @@ bool ModuleInput::CleanUp()
 
 	return true;
 }
+
+void ModuleInput::MapInput()
+{
+	if (App->IsOnPlayMode())
```
gerard-queralt, line 405: You could, if you want, invert the condition and return, to avoid nesting.

Source/Modules/ModuleInput.h
```diff
@@ -51,6 +51,7 @@ class ModuleInput : public Module
 
 	bool Init() override;
 	UpdateStatus Update() override;
+	UpdateStatus PreUpdate() override;
```
gerard-queralt, line 54: Really dumb thing, but it'd help my OCD if you could declare this method before the Update ^^'

Source/Modules/ModuleInput.h
```diff
@@ -138,6 +134,9 @@ class ModuleInput : public Module
 	std::unique_ptr<SDL_Cursor, SDLCursorDestroyer> moveCursor;
 	std::unique_ptr<SDL_Cursor, SDLCursorDestroyer> zoomCursor;
 	std::unique_ptr<SDL_Cursor, SDLCursorDestroyer> defaultCursor;
+
+	void MapInput();
```
gerard-queralt, line 138: Maybe a more descriptive name, MapControllerInput or something like that.

gerard-queralt (commented):
Code looks great, left some minor comments.

Source/Modules/ModuleInput.cpp
MarcGonzalezMoratona: why is this used for?

Source/Modules/ModuleInput.h
MarcGonzalezMoratona: I was looking for a better name since the first second I created this method. Thank you!

gerard-queralt (approved):
Haven't looked too much at the changes in scripts, but the engine code looks great!

Piterga12 (approved):
The movement and dash feels much better like this! Good job!

diegusmejia117 (approved):

PR #727, opened 2023-09-16
feat: controller rumble on damage taken

MarcGonzalezMoratona:


gerard-queralt:
Couple things:
1) First few commits don't look related to the scope of the MR at all. If you want to merge both things in a single MR, I'd appreciate if you could explain in the description what was done for each "task" so we know where to look.
2) As for the rumble, I don't like the idea of having that code script-side; the fact that we use SDL is an implementation detail that the scripts should not need to know or deal with. I'd encapsulate that code in a method in ModuleInput, which will also avoid repeated code.

PR #731, opened 2023-09-16
feat: reflections using local IBL

PaablooCH:
First approach to reflections.
To create a local IBL you can choose between:

- the normal way, right click inspector -> create ComponentLocalIBL 
image

- Or right click in the hierarchy -> create light -> local IBL 
image

Once created, you have 4 parameters to modify the firsts 2 corresponds to the parallax aabb (the red one) and the others 2 to the influence aabb (green). With the green one you choose where do you want to apply the reflection, two green aabb can collide and it will show a blender between them basic on the weight of each one. With the reds you need to play until find the best proportions for your result, that's what Carlos said: 
image

<h2>Where you can place them?</h2>

This statement 
image
was took from this [link](https://developer.valvesoftware.com/wiki/Cubemaps).

<h2>Some examples</h2>

| Before  | Now |
| ------------- | ------------- |
| ezgif com-video-to-gif  | ezgif com-video-to-gif (1)  |

ezgif com-optimize

Load 00_prueba_reflejo to test.


AlexisDieu:
All the reflection, value parallax and influence + maps seems working correctly but when I try to copy a local IBL it's crashing there
image

AlexisDieu:
Everything related to the reflection and reflection value seems working well, didn't find any issue or something weird actually but when I try to copy a Local IBL, it's stopping the render of all other (It's not in the video but move them fix it) and the copied one is totally black (to fix it I need to generate the maps well it's the only correction I found but maybe copy it again will fix it as for light)
ezgif com-video-to-gif (10)

PaablooCH:
> Everything related to the reflection and reflection values seems working well, didn't find any issue or something weird actually but when I try to copy a Local IBL, it's stopping the render of all other (It's not in the video but move them fix it and make them render again) and the copied one is totally black (to fix it I need to generate the maps well it's the only correction I found but maybe copy it again will fix it as for light)

Is how it works, each time you move, copy or scale you need to click on the button to regenerate the maps.

AlexisDieu:
> > Everything related to the reflection and reflection values seems working well, didn't find any issue or something weird actually but when I try to copy a Local IBL, it's stopping the render of all other (It's not in the video but move them fix it and make them render again) and the copied one is totally black (to fix it I need to generate the maps well it's the only correction I found but maybe copy it again will fix it as for light)
> 
> Is how it works, each time you move, copy or scale you need to click on the button to regenerate the maps.

Okey and for the other IBL that are not rendering anymore after a copy until we move them ?

PaablooCH:
> Okey and for the other IBL that are not rendering anymore after a copy until we move them ?

fixed, thx

AlexisDieu (approved):

PR #734, opened 2023-09-17
feat: Use skybox as cubemap

PaablooCH:
Self explicatory.

<h2>Example</h2>

ezgif com-optimize

**Don't focus on the screen size change**


ericdacal (approved):
It work perfect!

AlexisDieu:
It seems to work correctly but I am just confuse on something, why sometime when a skybox is use also as a cubemap, it's mirroring and sometime it like if it was transparent ? (first picture mirroring second is transparent)
image
image
(Maybe it's just unlucky due to my angle and because it's a cube idk I just wanted to notice it)

PaablooCH:
> It seems to work correctly but I am just confuse on something, why sometime when a skybox is use also as a cubemap, it's mirroring and sometime it like if it was transparent ? 

Hi, now he have a bug in one shader, that is fixed in #731, which switch the x's faces, so probably one of them is wrong.

Source/DataModels/Components/ComponentSkybox.h
```diff
@@ -1,32 +1,41 @@
...
-class ComponentSkybox : public Component, public Drawable
+class ComponentSkybox : public Component
 {
 public:
 	ComponentSkybox(bool active, GameObject* owner);
 	~ComponentSkybox() override;
 	
-	
-	void Draw() const override;
+	void Draw();
```
markmeliaa, line 14: Why did you remove the `cost override`? I see that inside `Draw` you now call `IsEnabled()` istead of just getting the `enable` variable, what's the point in that?

markmeliaa (commented):
I don't really get the render stuff but I have a question about your code:

Source/DataModels/Components/ComponentSkybox.h
PaablooCH: > Why did you remove the `cost override`?

Because the `override` implies that skybox was using the `draw` call inside of Component and this function is called apart during the scene update and has a completely different purpose than the one we use in moduleRender. For the `const`, we can keep it, true (probably I will change it in a future pr, so I don't lose Eric's approve).

> I see that inside `Draw()` you now call `IsEnabled()` istead of just getting the `enable` variable, what's the point in that?

Component has already an `enable` variable that works exactly as the one that I have deleted.

markmeliaa (approved):

PR #735, opened 2023-09-17
feat: rumble

MarcGonzalezMoratona:
Add a rumble method inside moduleInput that allows control for left and right rumble and duration.
Add rumble on damage taken and on hacking mistake

### DIFF
--- a/Game/Scripts/HealthSystem.cpp
+++ b/Game/Scripts/HealthSystem.cpp
@@ -4,6 +4,8 @@
 #include "Components/ComponentAnimation.h"
 #include "Components/ComponentScript.h"
 #include "Components/ComponentParticleSystem.h"
+#include "Application.h"
+#include "ModuleInput.h"
 
 #include "../Scripts/PlayerAttackScript.h"
 #include "../Scripts/EnemyClass.h"
@@ -107,6 +109,9 @@ void HealthSystem::TakeDamage(float damage)
 			float actualDamage = std::max(damage - playerDefense, 0.f);
 
 			currentHealth -= actualDamage;
+
+			ModuleInput* input = App->GetModule<ModuleInput>();
+			input->Rumble(16384, 16384, 250);
 
 			if (currentHealth - damage <= 0)
 			{
--- a/Game/Scripts/HealthSystem.cpp
+++ b/Game/Scripts/HealthSystem.cpp
@@ -111,7 +111,7 @@ void HealthSystem::TakeDamage(float damage)
 			currentHealth -= actualDamage;
 
 			ModuleInput* input = App->GetModule<ModuleInput>();
-			input->Rumble(16384, 16384, 250);
+			input->Rumble();
 
 			if (currentHealth - damage <= 0)
 			{
--- a/Source/Modules/ModuleInput.cpp
+++ b/Source/Modules/ModuleInput.cpp
@@ -367,13 +367,13 @@ bool ModuleInput::CleanUp()
 	return true;
 }
 
-void ModuleInput::Rumble(Uint16 strengthLeft, Uint16 strengthRight, int duration) const
+void ModuleInput::Rumble(Uint16 intensityLeft, Uint16 intensityRight, Uint32 durationMs) const
 {
 	SDL_GameController* controller = FindController();
 
 	if (controller != nullptr)
 	{
-		if (SDL_GameControllerRumble(controller, strengthLeft, strengthRight, duration) != 0)
+		if (SDL_GameControllerRumble(controller, intensityLeft, intensityRight, durationMs) != 0)
 		{
 			LOG_ERROR("Error on controller rumble");
 		}
--- a/Source/Modules/ModuleInput.cpp
+++ b/Source/Modules/ModuleInput.cpp
@@ -22,7 +22,9 @@ ModuleInput::ModuleInput() :
 	mouseMotion(float2::zero),
 	mousePosX(0),
 	mousePosY(0),
-	direction{ JoystickHorizontalDirection::NONE, JoystickVerticalDirection::NONE }
+	direction{ JoystickHorizontalDirection::NONE, JoystickVerticalDirection::NONE },
+	rumbleIntensityMap(defaultRumbleIntensityMap),
+	rumbleDurationMap(defaultRumbleDurationMap)
 {
 }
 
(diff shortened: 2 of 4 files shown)


Piterga12 (dismissed):
Tested it with a ps4 controller, works well

Source/Modules/ModuleInput.cpp
```diff
@@ -366,3 +366,16 @@ bool ModuleInput::CleanUp()
...
+
+void ModuleInput::Rumble(Uint16 strengthLeft, Uint16 strengthRight, int duration) const
+{
+	SDL_GameController* controller = FindController();
+
+	if (controller != nullptr)
+	{
+		if (SDL_GameControllerRumble(controller, strengthLeft, strengthRight, duration) != 0)
+		{
+			LOG_ERROR("Error on controller rumble");
```
gerard-queralt, line 378: Could we get the error? I think we can use [`SDL_GetError`](https://wiki.libsdl.org/SDL2/SDL_GetError)
```suggestion
			LOG_ERROR("Error on controller rumble: {}", SDL_GetError());
```

Source/Modules/ModuleInput.h
```diff
@@ -67,11 +67,13 @@ class ModuleInput : public Module
 	void SetKey(SDL_Scancode scanCode, KeyState newState);
 	void SetMouseButton(Uint8 mouseButtonCode, KeyState newState);
 
-	SDL_GameController* FindController();
+	SDL_GameController* FindController() const;
 	SDL_JoystickID GetControllerInstanceID(SDL_GameController* controller) const;
 
 	JoystickMovement GetDirection() const;
 
+	void Rumble(Uint16 strengthLeft, Uint16 strengthRight, int duration) const;
```
gerard-queralt, line 75: I'd add the units to the `duration` variable. We could even use a [`std::chrono::duration`](https://en.cppreference.com/w/cpp/chrono/duration) but that's overkill.

Game/Scripts/HealthSystem.cpp
```diff
@@ -108,6 +110,9 @@ void HealthSystem::TakeDamage(float damage)
 
 			currentHealth -= actualDamage;
 
+			ModuleInput* input = App->GetModule<ModuleInput>();
+			input->Rumble(16384, 16384, 250);
```
gerard-queralt, line 114: These values all look basically random. I trust them, but whenever someone wants to add another rumble they'd need to go to one of the existing calls and copy-paste them. A couple suggestions:

1. Add an overload without parameters that uses what we think is default (these ones). We could even consider removing the existing one entirely, though it's probably best if we keep it. Thinking about it, we could probably use an overload that takes only one rumble value and a duration and applies that rumble to both sides as well.
2. What I think is best, though admittedly more work, is to dumb this down for the caller and use a couple enum classes that allow them to describe the effect they want. For example, this call could look like this:
```c++
// Using the overload with two parameters I mentioned earlier
input->Rumble(RumbleStrength::Normal, RumbleDuration::Normal);
```
Internally, we'd keep a dictionary that translates these enums into actual values. This is not only easier to use (in my opinion), it has the added benefit of making it easier to adjust the values of this call, since we won't need to change every script that calls it, only the internal dictionary.

gerard-queralt (commented):
Looks better, thanks for the work. Let me know what you think about my suggestions.

Source/Modules/ModuleInput.h
MarcGonzalezMoratona: I thought about adding the units and then I forgot. Thank you! I'll check which is the type used by SDL

Source/Modules/ModuleInput.cpp
MarcGonzalezMoratona: I didn't know we could do this. Nice one!

Game/Scripts/HealthSystem.cpp
MarcGonzalezMoratona: I haven't thought about this. I think this will take a bit of time, but will be good for me to learn more about overloads. I haven't developed this before, let's do it!

MarcGonzalezMoratona: The overload for to keep intensity the same on both sides is an idea I really like. The dictionary for RumbleStrengths is a bit ambiguous for me, because I'm not sure of what's Normal intensity, and we are not sure if this values will be the final ones, at the beginning it was just a POC. The idea looks good to me because of the maintanability benefits

Source/Modules/ModuleInput.h
```diff
@@ -2,13 +2,46 @@
...
+#include <map>
 
 #define NUM_MOUSEBUTTONS 5
 #define BMP_FREELOOKSURFACE "Assets/MouseCursors/freeLook.bmp"
 #define BMP_ORBITSURFACE "Assets/MouseCursors/orbit.bmp"
 #define BMP_MOVESURFACE "Assets/MouseCursors/move.bmp"
 #define BMP_ZOOMSURFACE "Assets/MouseCursors/zoom.bmp"
 
+
+namespace
```
gerard-queralt, line 14: I think we don't need this namespace for the enums. We could move them both to a separate file in the `Enums` folder and forward-declare them, too. Up to you.

Source/Modules/ModuleInput.h
```diff
@@ -2,13 +2,46 @@
...
+};
+enum class RumbleDuration
+{
+	SHORT,
+	NORMAL,
+	LONG,
+	LONGER,
+};
+
+const std::map<RumbleIntensity, Uint16> defaultRumbleIntensityMap({
```
gerard-queralt, line 31: I'd move both of these maps to the source file. Also, they (and by extension the members) can probably be unordered maps.

gerard-queralt: Though we might not even need the members if we don't want them to change, we can just use these two maps as-is.

Source/Modules/ModuleInput.cpp
```diff
@@ -366,3 +368,34 @@ bool ModuleInput::CleanUp()
 
 	return true;
 }
+
+void ModuleInput::Rumble(RumbleIntensity intensityLeft, RumbleIntensity intensityRight, RumbleDuration durationMs) const
+{
+	SDL_GameController* controller = FindController();
+
+	if (controller != nullptr)
```
gerard-queralt, line 376: Should we add a verbose log if there's no controller? Thinking if there can be an instance where we connect the controller but for some reason is not detected. Probably not needed, so up to you.

gerard-queralt (commented):
Code looks much better, great job! Left some picky comments still :P

Source/Modules/ModuleInput.h
gerard-queralt: For further justification, I have two issues with them being in the header:
1) Everyone can access them, even if they can't modify them it clutters the global namespace
2) Most importantly, adjusting these values would require to recompile the header, which in turn requires a recompilation of every source file that includes it and, even worse, headers that include it (which hopefully don't exist, but who knows), which ends up being like a chain reaction

Source/Modules/ModuleInput.cpp
MarcGonzalezMoratona: The only reason I found for the controller not being found is because the controller is not connected. I wouldn't complicate this part, let's suppose the controller always works

Source/Modules/ModuleInput.h
MarcGonzalezMoratona: I'll put the enums out of the namespace but keep it in the moduleInput with the other enums of this file

Source/Modules/ModuleInput.h
MarcGonzalezMoratona: I saw a few days ago that unordered maps have smaller cost for access, so I was already thinking of doing that. I followed the idea from the gamepadMapping, but now I think it makes more sense to use the maps directly. Is it okay if I put the maps as local variables inside the Rumble method?

Also for the gamepadMapping. This will be changed in the [controller input PR](https://github.com/Horizons-Games/Axolotl-Engine/pull/724)

gerard-queralt: It's okay, but is there a reason to do that? They'll need to be instantiated and destroyed whenever we call the method, and they'll be less accessible if we want to change them.

gerard-queralt: To avoid instantiation you could declare them as static inside the method, so they'll only be instantiated the first time the method is called, if you want.

MarcGonzalezMoratona: I think I don't understand what you mean with putting the maps in the source file. That's why I put them in the Rumble method. Which is the place in the source file where you were expecting to have this maps declared?

gerard-queralt: I meant an unnamed namespace at the top of it, sorry for the confusion.

gerard-queralt (approved):
Looks great Marc, thank you for bearing with me!

Source/Modules/ModuleInput.h
MarcGonzalezMoratona: I didn't know unnamed spaces could be created in source files too, but I like it the way it's done right now, so I'll keep it

Piterga12 (approved):

diegusmejia117 (approved):

PR #738, opened 2023-09-19
Fix : ComponentLine copy fixed

AlexisDieu:
- ComponentLine was using the same buffer than the one copied : fixed
- ComponentLine and particle were adding themself to a list in the scene when being copied but it was done during the copy (ctrl+C) and that should be done during the paste (ctrl+V or D) esle the list was incremented without the GameObject existing

### DIFF
--- a/Source/DataModels/Components/ComponentLine.cpp
+++ b/Source/DataModels/Components/ComponentLine.cpp
@@ -35,28 +35,9 @@ ComponentLine::ComponentLine(const bool active, GameObject* owner) : Component(C
 ComponentLine::ComponentLine(const ComponentLine& other, GameObject* owner) :
 	Component(ComponentType::LINE, other.IsEnabled(), owner, true),
 	numTiles(other.numTiles),speed(other.speed),time(other.time),dirtyBuffers(other.dirtyBuffers), 
-	gradient(new ImGradient(false)),offset(other.offset),tiling(other.tiling),sizeFading(other.sizeFading),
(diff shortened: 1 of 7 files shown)


AuxPort66 (approved):
Good job you even catch a memory leak.

Source/DataModels/Components/ComponentLine.cpp
```diff
@@ -32,13 +32,24 @@ ComponentLine::ComponentLine(const bool active, GameObject* owner) : Component(C
 		});
 }
 
+ComponentLine::ComponentLine(const ComponentLine& other, GameObject* owner) :
```
PaablooCH, line 35: why You create the constructor with a second parameter?

AlexisDieu: this si the copy constructor, the first one is for the copy and the second one is to follow the Component(ComponentType::LINE, other.IsEnabled(), owner, true) so the Component constructor. Maybe other.getOwner() was good but I just follow the step I did for the default constructor

PaablooCH: what I mean is those steps are now duplicated and You should use the copy constructor of the `component` class like in the `componentTransform` class for example.
this is what copy constr does in `compoent` class:
`Component::Component(const Component& component) : type(component.type),
	enabled(component.enabled),
	owner(nullptr),
	canBeRemoved(component.canBeRemoved)
{
}`
and before you leave the function `copyComponent` in `gameobject.cpp` (line 484) you set the correct owner, so you are setting it twice:
image

AlexisDieu: I change it, thanks for the feedback

PR #739, opened 2023-09-19
feat: move gameobject shortcut

MarcGonzalezMoratona:


Source/Modules/ModuleEditor.cpp
```diff
@@ -75,7 +75,6 @@ bool ModuleEditor::Init()
 	ImGui::CreateContext();
 
 	ImGuiIO& io = ImGui::GetIO();
-	io.ConfigFlags |= ImGuiConfigFlags_NavEnableKeyboard;	// Enable Keyboard Controls
```
MarcGonzalezMoratona, line 78: I need to remove this because moving the arrows we were selecting more than a gameObject in the hierarchy

PR #742, opened 2023-09-20
feat: copy constructor trail

PaablooCH:
self explanatory

NOTE: json had a grammatical error inside, you will need to regenerate them (save the scene)


MarcGonzalezMoratona (approved):

Source/DataModels/Components/ComponentTrail.cpp
```diff
@@ -42,9 +42,19 @@ onPlay(false), catmunPoints(10)
 	
 }
 
+ComponentTrail::ComponentTrail(const ComponentTrail& trail) : Component(trail), maxSamplers(trail.maxSamplers), 
+duration(trail.duration), minDistance(trail.minDistance), width(trail.width), ratioWidth(trail.ratioWidth), 
+blendingMode(trail.blendingMode), onPlay(trail.onPlay), catmullPoints(trail.catmullPoints)
+{
+	points.reserve(maxSamplers);
+	gradient = new ImGradient();
```
AuxPort66, line 50: If you put new ImGradient(trail.gradient) you will copy the colors too. About colors, I see that we only use the first and the last? Then maybe we should only save the first and the last mark, like in particles. If i'm wrong and you are using all the marks ignore this.

Also we can copy the resource too. Thank you, the rest works well.

MacXxs (approved):

AuxPort66 (approved):

Source/DataModels/Components/ComponentTrail.cpp
PaablooCH: did it

PR #745, opened 2023-09-21
fix: Player movement while hacking

OriolSors:
Fixed issues with player movement while hacking


markmeliaa (commented):
I've tried the level 2 scene and while the movement while hacking seems to be fixed, once I get inside the hacking area, I can trigger the hacking with the E key even after leaving the hacking area and being far far away, is this a known bug?

Game/Scripts/PlayerHackingUseScript.cpp
```diff
@@ -31,7 +31,7 @@ void PlayerHackingUseScript::Start()
 	transform = GetOwner()->GetComponentInternal<ComponentTransform>();
 	rigidBody = GetOwner()->GetComponentInternal<ComponentRigidBody>();
 	hackZone = nullptr;
-
+	playerManager = GetOwner()->GetComponentInternal<PlayerManagerScript>();
```
markmeliaa, line 34: Forgot to add this, we should not use `GetComponentInternal` in the gameplay scripts, use `GetComponent` instead.

OriolSors: Done! Thanks for the advice!

OriolSors:
> I've tried the level 2 scene and while the movement while hacking seems to be fixed, once I get inside the hacking area, I can trigger the hacking with the E key even after leaving the hacking area and being far far away, is this a known bug?

Yes, fixed. I didn't know that issue, first time noticed. Thank you again!!!

markmeliaa (approved):

Piterga12 (approved):
Good job! Works perfectly now!

PR #747, opened 2023-09-21
fix: level 1 cameras

MarcGonzalezMoratona:


AuxPort66:
Art is updating lvl1 scene, so to avoid conflicts we need to import the cameras inside the lvl1 and erase the old ones. @Dedado3000 knows the process.

markmeliaa:
conflicts 👀

Dedado3000 (dismissed):
Good Changes

DADSILVER:
Good job I like all the cameras.

There is a thing no related to your pr, but the jump attack doesn't work.

MarcGonzalezMoratona:
> Good job I like all the cameras.
> 
> There is a thing no related to your pr, but the jump attack doesn't work.

Thank you David! I'm aware about the jumpAttack error. I think it was introduced by the changes in the Paralyze method in the recent hacking PR.  I'm working to solve it in the dash branch

diegusmejia117 (changes requested):
Tested it, cameras are a lot better, gj!

But in the update there are 2 things that must be worked on, 

1- Camera bug Sara already told you about/
2- In the second long passageway, that connects the first combat zone with the cistern parkour, i think the camera angles changed and now is a little lower, this effect is cool but it also creates a problem, as with the new angle, you can see the skylights hole. and you can see outside into the skybox.

I suggest to move this particular camera so that this is not seen.

MarcGonzalezMoratona:
> 2- In the second long passageway, that connects the first combat zone with the cistern parkour, i think the camera angles changed and now is a little lower, this effect is cool but it also creates a problem, as with the new angle, you can see the skylights hole. and you can see outside into the skybox.

This camera angle is lower because the player was reaching the ceiling and the camera went out of the corridor, so we should find a way to solve both problems in the best possible way

MarcGonzalezMoratona:
@diegusmejia117 @AuxPort66 @DADSILVER @Dedado3000 the cameras are ready to be tested again!

Changes since last time:
- Parkour 1 platform is now more visible (Miki requested it)
- Intersection after parkour 1 camera does not cross the walls (Sara requested it)
- Corridor before parkour 2 camera angle is a bit lower to avoid seeing the skybox (Diego requested it)
- Parkour 2 camera does not cross the waterfall when you go to the left side (right side cannot be improved because we need to see the force cube and the next platform to jump) (Sara requested it)
- Miniboss door first entering camera does not cross the wall (Teacher Jordi requested it)

diegusmejia117 (approved):

diegusmejia117 (approved):

Piterga12 (dismissed):

Dedado3000 (approved):

diegusmejia117 (approved):
Tested it before, thanks Marc for channgin the camera i told you!

PR #753, opened 2023-09-25
feat: rumble on jump attack

MarcGonzalezMoratona:
This has to be tested after merging #758


markmeliaa:
Why did you decide to add the rumble in the jump attack as well? For now, iirc the rumble is only added when you "do something wrong" (aka taking damage or missing during the hacking), why should we add it to the jump attack (more so when no other attack makes the controller rumble).

MarcGonzalezMoratona:
> Why did you decide to add the rumble in the jump attack as well? For now, iirc the rumble is only added when you "do something wrong" (aka taking damage or missing during the hacking), why should we add it to the jump attack? (more so when no other attack makes the controller rumble)

We thought that would be cool to have rumble in more places. The idea was not to add rumble when you do something wrong, it was to add it where it makes sense for the player to feel more immersed. We added it to the jump attack because it's an attack that makes a grenade explosion or a hole in the ground.

If we find other places to add this rumble (like a cinematic, picking a poster or when you select an option in the menu), there are different kinds of rumble that we can add (on left/right sides of the controller independently, with higher/lower intensity and longer/shorter. It's an easy to implement feature, that gives good feedback to the player and does not break other features. So probably we should use it as much as possible. To give you an example, yesterday I installed Disney Speedstorm in my computer and they use it a lot (in the menu, during a race, after every race... ), and it doesn't feel too much rumble.

I think it will be better not to limit the rumble to "wrong actions" (we can make a special rumble for this), but of course this is something we can decide together, and see where it fits better and which kind of rumble to use at every place

markmeliaa:
> If we find other places to add this rumble (like a cinematic, picking a poster or when you select an option in the menu), there are different kinds of rumble that we can add (on left/right sides of the controller independently, with higher/lower intensity and longer/shorter. It's an easy to implement feature, that gives good feedback to the player and does not break other features. So probably we should use it as much as possible. To give you an example, yesterday I installed Disney Speedstorm in my computer and they use it a lot (in the menu, during a race, after every race... ), and it doesn't feel too much rumble.
> 
> I think it will be better not to limit the rumble to "wrong actions" (we can make a special rumble for this), but of course this is something we can decide together, and see where it fits better and which kind of rumble to use at every place

Okay, I just wanted to make sure that we have a clear vision on where to implement the rumble and check that we are not just applying it randomly, thanks for your answer!

markmeliaa (approved):

Piterga12 (approved):

PR #754, opened 2023-09-26
perf: Improve collision management

OriolSors:
Now the ManageCollisions() only check RigidBodies that are inside the Player's Quadtree.


totisama (approved):
Nice solution!!

AlexisDieu (approved):

PR #759, opened 2023-09-27
feat: Elevator behaviour

OriolSors:
- Activation Logic has an option for taking the hacking into account.
- Elevator front doors linked with the hacking completion.
- Added a cooldown to the elevator in order to prevent infinite loop up and down.
- Elevator goes up when the player is going to face the miniboss.
- Elevator goes down when the miniboss is defeated.

For testing:
- Attach miniboss GameObject to the ElevatorCore script, and custom the cooldown and the elevator speed.

### DIFF
--- a/Game/Scripts/ActivationLogic.cpp
+++ b/Game/Scripts/ActivationLogic.cpp
@@ -18,6 +18,7 @@
 
 #include "../Scripts/CameraControllerScript.h"
 #include "../Scripts/PlayerManagerScript.h"
+#include "../Scripts/HackZoneScript.h"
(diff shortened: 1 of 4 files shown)


DADSILVER (dismissed):
Good Job

markmeliaa (commented):
GJ! I liked that you encapsulated functionality in different functions so for the elevator to go up/down you just have to call the function.

markmeliaa (approved):
Whoops, commented instead of approving it.

DADSILVER (approved):

PR #760, opened 2023-09-27
feat: loading scripts logic

gerard-queralt:
This merge request adds a `SceneLoadingScript` and a demo loading screen courtesy of @ericdacal.

# How does it work?

`SceneLoadingScript` exposes a single member, which is the real scene we want to load. It will automatically format the path, so setting the string field to "Example" will result in an attempt to load `Lib/Scenes/Example.axolotl`. It has a constant, hard-coded path to what is assumed to be the default loading scene, `LoadingScreen`. This can be changed later if needed (say, by adding more loading screens and showing one randomly). On starting, the script verifies that both the constant loading screen and the dynamic scene to be loaded exist.

The script needs to be called by another one; in this MR I've updated the Main Menu scene so the "Play" button calls this new script. When the method `SceneLoadingScript::StartLoad` is called, the script loads the loading scene, waits 6 frames so that `ModuleRender` can tick (render is blocked during loading) and starts the load of the actual scene, using the callbacks the scene loading method provides.

# How can you test?

You'll need a sort-of functional Game flow to test this, and remember that only Main Menu has been updated.

1. Load `00_MainMenu_VS3.axolotl`, `LoadingScreen.axolotl` and `__LEVEL1_ALFA.axolotl`, in whatever order you want. This will populate the Lib folder.
2. Once the three scenes have been loaded, choose "Build" on the main menu of the engine, select `00_MainMenu_VS3.axolotl` as the starting scene, choose to regenerate the .zip file and start the build (compilation configuration doesn't matter, up to you).
3. Grab a coffee, call your family, do some exercise, whatever you feel like.
4. Once the build is done, start the game and press "Play". You should see a screenshot of the sewers (the temporary loading screen) and after a while the level one should load.

# Engine changes

Engine-side required mostly guards to prevent updates from happening while the scene was loaded. ~Coolest addition, even though it's probably useless outside this use case, is that an optional parameter has been added to the `Scheduler::ScheduleTask` method to set a delay (in frames) for the task to run. This was done, as stated before, so the `ModuleRender` could run for a bit.~ Cherrypicked in #766 instead.

# What to do?

The scenes would need to be updated so all transitions use this logic, eithe by me or someone else in this MR or another.

We also need to see how much stuff we can put on the loading screen. The quickest way to make this whole thing work was to prevent a bunch of modules from updating during load, but we can try to roll those changes back and allow some functionality so we can have some visual feedback.


markmeliaa (approved):
Code looks great and the functionality works as expected, the only issue that I have is that maybe the loading time is a bit to high, but if you decided that it was necessary to set the `frameDelay` this high, then you'd know better!

And I agree that we should add some functionality (at least visual) to the load screen, like maybe a loading wheel or a loading bar if plausible, but for now it looks fine enough!

gerard-queralt:
> the only issue that I have is that maybe the loading time is a bit to high, but if you decided that it was necessary to set the `frameDelay` this high, then you'd know better!

Didn't really test it, just went with my gut and thought a 10th of a second would be a good value. In theory, a single frame of delay should be good enough, I was just being defensive in case there were other pending actions that needed to be fired. I'll try to change it when I get back home, feel free to try it yourself if you want 😄

markmeliaa:
> Didn't really test it, just went with my gut and thought a 10th of a second would be a good value. In theory, a single frame of delay should be good enough, I was just being defensive in case there were other pending actions that needed to be fired. I'll try to change it when I get back home, feel free to try it yourself if you want 😄

From my tries, the music and sounds start at like, three/four seconds in, but the load screen does not disappear until like a lot later, idk how long it would take to render all the scene but which threw me off the most was that the sounds were played long before the loading screen disappeared, rather than the load itself taking so long. So imo, we either stop the sounds from playing so early or we make the loading screen take less time (if possible).

gerard-queralt:
Oh, sorry, I think I didn't explain myself, but I'm pretty sure what you're reporting is unrelated to the delay. This 6 frames I mentioned is the time we wait between finishing the loading screen load and starting the load of the scene we want. It shouldn't cause that much delay, I think it's the same "problem" we have on develop, where music and such start before you can see the scene. Worth checking out, but IMO out of scope for this MR. Let me know if you disagree.

gerard-queralt:
Cherrypicked the Scheduler changes in #765. In theory merging order should not matter, but ideally we merge that one first so I can rebase and clean a bit the history of this branch.

markmeliaa:
> I think it's the same "problem" we have on develop, where music and such start before you can see the scene. 

Oh I see now, then don't mind me, this pr is fine as is.

markmeliaa:
The pr is now erasing a bunch of assets, is that intended?

gerard-queralt:
> The pr is now erasing a bunch of assets, is that intended?

Whoops, it's most definitely not, sorry about that. Looks like it got borked when I merged the branch from #765. Not at home right not, will fix it as soon as I can 🙇‍♂️

markmeliaa (dismissed):

markmeliaa (dismissed):

Vlukkee (dismissed):
Works fine! Nice job

Vlukkee (approved):

markmeliaa (approved):

PR #763, opened 2023-09-28
feat<text>: add text with freety lib

elbuho12Jesus:
I created a UI tool that works with the `Canvas`, this is used to add 2D text from different sources, please tell me if you need any changes.


Source/DataModels/Components/UI/ComponentText2D.cpp
```diff
@@ -0,0 +1,267 @@
...
+{	
+	programShared->BindUniformInt("text", 0);
+	programShared->BindUniformFloat3("textColor", color);	
+	
+	glActiveTexture(GL_TEXTURE0);
+	glBindVertexArray(VAO);
+	/*ENGINE_LOG("VAO: %i", VAO)*/
+	// iterate through all characters
+	std::string::const_iterator c;
+	for (c = text.begin(); c != text.end(); c++)
```
gerard-queralt, line 192: Wouldn't it be easier to use a [range loop](https://en.cppreference.com/w/cpp/language/range-for)? Up to you though.

Source/DataModels/Components/UI/ComponentText2D.cpp
```diff
@@ -0,0 +1,267 @@
...
+		float w = ch.Size.x * scale;
+		float h = ch.Size.y * scale;		
+		// update VBO for each character
+		float vertices[6][4] = { { xpos / 48.0, (ypos + h) / 48.0, 1.0f, 1.0f },
+								 { xpos / 48.0, ypos / 48.0, 1.0f, 0.0f },
+								 { (xpos + w) / 48.0, ypos / 48.0, 0.0f, 0.0f },
+
+			{ xpos / 48.0, (ypos + h) / 48.0, 1.0f, 1.0f },
+								 { (xpos + w) / 48.0, ypos / 48.0, 0.0f, 0.0f },
+								 { (xpos + w) / 48.0, (ypos + h) / 48.0, 0.0f, 1.0f }
```
gerard-queralt, line 208: Make sure to add `f` at the end of these "floats" (which are now doubles) to avoid compiler warnings.

Source/DataModels/Components/UI/ComponentText2D.cpp
```diff
@@ -0,0 +1,267 @@
...
+					break;
+				case GL_OUT_OF_MEMORY:
+					error = "OUT_OF_MEMORY";
+					break;
+				case GL_INVALID_FRAMEBUFFER_OPERATION:
+					error = "INVALID_FRAMEBUFFER_OPERATION";
+					break;
+			}
+			LOG_INFO(error);
+		}
```
gerard-queralt, line 235: 1. Why use a `while` loop? You're never changing the value of `err` as far as I can tell
2. If it's an error, why log it as info?
```suggestion
		GLenum err = glGetError();
		if (err != GL_NO_ERROR)
		{
			std::string error;
			switch (err)
			{
				case GL_INVALID_OPERATION:
					error = "INVALID OPERATION";
					break;
				case GL_INVALID_ENUM:
					error = "INVALID ENUM";
					break;
				case GL_INVALID_VALUE:
					error = "INVALID VALUE";
					break;
				case GL_OUT_OF_MEMORY:
					error = "OUT OF MEMORY";
					break;
				case GL_INVALID_FRAMEBUFFER_OPERATION:
					error = "INVALID FRAMEBUFFER OPERATION";
					break;
				default:
					error = "UNKNOWN ERROR";
					break;
			}
			LOG_ERROR("OpenGL error while binding the texture in ComponentText2D owned by {}; error: {}", GetOwner(), error);
		}
```

Source/DataModels/Components/UI/ComponentText2D.cpp
```diff
@@ -0,0 +1,267 @@
...
+		/*for (int i=0;i<6;i++)
```
gerard-queralt, line 236: Remove commented out code, here and elsewhere, if it's not needed. If it is, add an AXO_TODO to address the reason why it's commented out.

Source/DataModels/Components/UI/ComponentText2D.cpp
```diff
@@ -0,0 +1,267 @@
...
+		// update content of VBO memory
+		glBindBuffer(GL_ARRAY_BUFFER, VBO);
+		/*ENGINE_LOG("VBO: %i", VBO)*/
+		glBufferSubData(GL_ARRAY_BUFFER, 0, sizeof(vertices), vertices); // be sure to use glBufferSubData and not glBufferData
+
+		glBindBuffer(GL_ARRAY_BUFFER, 0);
+		// render quad
+		glDrawArrays(GL_TRIANGLES, 0, 6);
+		// now advance cursors for next glyph (note that advance is number of 1/64 pixels)
+		x += (ch.Advance >> 6) * scale; // bitshift by 6 to get value in pixels (2^6 = 64 (divide amount of 1/64th pixels by 64 to get amount of pixels))
```
gerard-queralt, line 250: Is there not another way to do this? Always weary when seeing bit operations... Also, if you follow my comment to prevent copying Characters by getting them as references, make sure it doesn't screw with this. Can't remember off the top of my head if the bit-shift operator will modify the memeber.

Source/DataModels/Components/UI/ComponentText2D.cpp
```diff
@@ -0,0 +1,267 @@
...
+		currentFont = resourceFont;
+	}
+#endif
+	text = meta["text"];
+	color.x = static_cast<float>(meta["color_x"]);
+	color.y = static_cast<float>(meta["color_y"]);
+	color.z = static_cast<float>(meta["color_z"]);
+}
+
+std::vector<std::shared_ptr<ResourceFont>> ComponentText2D::getFonts()
```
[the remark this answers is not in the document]
gerard-queralt: I see it's used to select a font, but I still think the component is doing too many things at once.

Source/DataModels/Components/UI/ComponentText2D.cpp
```diff
@@ -0,0 +1,267 @@
...
+	glBindTexture(GL_TEXTURE_2D, 0);
+	//programShared->Deactivate();
+}
+
+void ComponentText2D::SetCharacters(unsigned int item)
+{
+	
+	std::vector<std::shared_ptr<ResourceFont>> resourceFonts = getFonts();
+	currentFont = resourceFonts[item];
+	currentFont.get()->LoadCharacter();
```
gerard-queralt, line 262: We're computing all possible fonts when we already know which one we want, and exposing a public member method in the process... I think a better approach would be to create a helper class that exposes a method like:
```c++
std::shared_ptr<ResourceFont> RequestFont(Font requestedFont);
```
Where `Font` is an enum class representing all the available font's in the engine (in your current code a script could call `SetCharacters` with any value and cause an aout-of-bounds read on the vector). You might even be able to memoize the resources vector after the first call, thought that might cause some issues if we want new fonts to be added at runtime. For an example of a helper class, you can take a look at `GameBuilder` or `SceneLoader`.

Source/DataModels/Components/UI/ComponentText2D.cpp
```diff
@@ -0,0 +1,267 @@
...
+}
+
+void ComponentText2D::Update()
```
gerard-queralt, line 265: Remove the empty overload and don't inherit from `Updatable` if your component does not update. It will take up memory in the Scene member vectors.

Source/DataModels/Resources/ResourceFont.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+}
+
+bool ResourceFont::ImportFont(std::string nameFount)
+{
+	// FreeType
+	// --------
+	FT_Library ft;
+	// All functions return a value different than 0 whenever an error occurred
+	if (FT_Init_FreeType(&ft))
```
gerard-queralt, line 34: We are initializing and shutting down the library each time we import a font? That feels like a very unnecessary and costly thing to do. Why not make `ModuleUI` take care of this?

Source/DataModels/Resources/ResourceFont.cpp
```diff
@@ -0,0 +1,161 @@
...
+void ResourceFont::InternalLoad()
+{
+}
+
+void ResourceFont::InternalUnload()
+{
+
+}
+
+bool ResourceFont::ImportFont(std::string nameFount)
```
gerard-queralt, line 28:
```suggestion
bool ResourceFont::ImportFont(const std::string& nameFount)
```

Source/DataModels/Resources/ResourceFont.cpp
```diff
@@ -0,0 +1,161 @@
...
+	// All functions return a value different than 0 whenever an error occurred
+	if (FT_Init_FreeType(&ft))
+	{
+		/*ENGINE_LOG("ERROR::FREETYPE: Could not init FreeType Library");*/
+		return false;
+	}
+
+	// find path to font
+	//std::string font_name = FileSystem::getPath("resources/fonts/Antonio-Bold.ttf");
+	std::string font_name = nameFount;
```
gerard-queralt, line 42: Do we need to create a copy of the string here? If we do, disregard my previous comment: pass the string by copy to the method and get rid of this `font_name` (which should be fontName).

Source/DataModels/Resources/ResourceFont.cpp
```diff
@@ -0,0 +1,161 @@
...
+	if (font_name.empty())
+	{
+		/*ENGINE_LOG("ERROR::FREETYPE: Failed to load font_name");*/
```
gerard-queralt, line 45: Put the logs back in, but make them more accurate instead of copy-pasting the same message. We didn't fail to load a font, the user passed us an empty string.

Source/DataModels/Resources/ResourceFont.cpp
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	// load font as face
+	FT_Face face;
+	if (FT_New_Face(ft, font_name.c_str(), 0, &face)) 
+	{
+		/*ENGINE_LOG("ERROR::FREETYPE: Failed to load font");*/
+		return false;
+	}
+	else
```
gerard-queralt, line 56: Unnecessary else, we're returning in the if block.

Source/DataModels/Resources/ResourceFont.cpp
```diff
@@ -0,0 +1,161 @@
...
+		// disable byte-alignment restriction
+		//glPixelStorei(GL_UNPACK_ALIGNMENT, 1);
+
+		// load first 128 characters of ASCII set
+		for (unsigned char c = 0; c < 128; c++)
+		{
+			// Load character glyph 
+			if (FT_Load_Char(face, c, FT_LOAD_RENDER))
+			{
+				LOG_INFO("ERROR::FREETYTPE: Failed to load Glyph");
```
gerard-queralt, line 70: Not info, either warning or error.

Source/DataModels/Resources/ResourceFont.cpp
```diff
@@ -0,0 +1,161 @@
...
+			/*LOG_INFO("n %c", c);
+			LOG_INFO("width %u", face->glyph->bitmap.width);
+			LOG_INFO("rows %u", face->glyph->bitmap.rows);
+			LOG_INFO("left %u", face->glyph->bitmap_left);
+			LOG_INFO("top %u", face->glyph->bitmap_top);
+			LOG_INFO("advance in x %u", static_cast<unsigned int>(face->glyph->advance.x));
+			LOG_INFO("size %u", face->glyph->bitmap.width * face->glyph->bitmap.rows);
+			LOG_INFO("size 2 %u", (unsigned int) sizeof((uint8_t) face->glyph->bitmap.buffer));*/
+
+			storeFont.push_back(store);
```
gerard-queralt, line 94:
```suggestion
			storeFont.emplace_back(
				c,
				face->glyph->bitmap.width,
				face->glyph->bitmap.rows,
				face->glyph->bitmap_left,
				face->glyph->bitmap_top,
				static_cast<unsigned int>(face->glyph->advance.x),
				sizeBuffer,
				pixel);
```
Or some variation of that.

Source/DataModels/Resources/ResourceFont.cpp
```diff
@@ -0,0 +1,161 @@
...
+			
+	}
+	// destroy FreeType once we're finished
+	FT_Done_Face(face);
+	FT_Done_FreeType(ft);
+	return true;
+}
+
+
+	bool ResourceFont::LoadCharacter()
```
gerard-queralt, line 105:
```suggestion
bool ResourceFont::LoadCharacter()
```

Source/DataModels/Resources/ResourceFont.cpp
```diff
@@ -0,0 +1,161 @@
...
+	{
+		//disable byte-alignment restriction
+		glPixelStorei(GL_UNPACK_ALIGNMENT, 1);
+		/*ENGINE_LOG("load textures")*/
+		for (StoreFontInPixels store : storeFont)
```
gerard-queralt, line 110:
```suggestion
		for (const StoreFontInPixels& store : storeFont)
```

Source/DataModels/Resources/ResourceFont.cpp
```diff
@@ -0,0 +1,161 @@
...
+			glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
+			glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
+			// now store character for later use
+			Character character = {
+				texture,
+				float2(store.width, store.height),
+				float2(store.left, store.top),
+				store.advance_x
+			};
+			characters.insert(std::pair<char, Character>(store.symbol, character));
```
gerard-queralt, line 141:
```suggestion
			characters.insert(std::pair<char, Character>(store.symbol, std::move(character)));
```
Or construct it in-place, up to you.

Source/DataModels/Resources/ResourceFont.cpp
```diff
@@ -0,0 +1,161 @@
...
+			glBindTexture(GL_TEXTURE_2D, 0);
+		}
+		return true;
+	}
+
+	std::map<char, Character> ResourceFont::GetCharacters()
```
gerard-queralt, line 147:
```suggestion
std::map<char, Character> ResourceFont::GetCharacters()
```

Source/DataModels/Resources/ResourceFont.cpp
```diff
@@ -0,0 +1,161 @@
...
+	{
+		if (characters.empty()) 
+		{
+			this->LoadCharacter();
+		}		
+		return characters;
+	}
+
+	bool ResourceFont::IsCharacterEmpty()
```
gerard-queralt, line 156:
```suggestion
bool ResourceFont::IsCharacterEmpty()
```

Source/DataModels/Resources/ResourceFont.h
```diff
@@ -0,0 +1,59 @@
...
+		const std::string& fileName,
+		const std::string& assetsPath,
+		const std::string& libraryPath);
+	virtual ~ResourceFont() override;
+	ResourceType GetType() const override;
+	bool ImportFont(std::string nameFount);
+	bool LoadCharacter();
+	std::map<char, Character> GetCharacters();
+	bool IsCharacterEmpty();
+	std::vector<StoreFontInPixels> storeFont;
```
gerard-queralt, line 41: Better to make the member private and use getters and setters, as per the C++ standard.

Source/DataModels/Resources/ResourceFont.h
```diff
@@ -0,0 +1,59 @@
...
+	ResourceFont(UID resourceUID,
+		const std::string& fileName,
+		const std::string& assetsPath,
+		const std::string& libraryPath);
+	virtual ~ResourceFont() override;
+	ResourceType GetType() const override;
+	bool ImportFont(std::string nameFount);
+	bool LoadCharacter();
+	std::map<char, Character> GetCharacters();
+	bool IsCharacterEmpty();
```
gerard-queralt, line 40:
```suggestion
	bool IsCharacterEmpty() const;
```

Source/DataModels/Scene/Scene.cpp
```diff
@@ -367,6 +368,10 @@ GameObject* Scene::CreateUIGameObject(const std::string& name, GameObject* paren
 			gameObject->CreateComponent<ComponentImage>();
 			sceneInteractableComponents.push_back(gameObject->CreateComponent<ComponentButton>());
 			break;
+		case ComponentType::TEXT2D:
+			gameObject->CreateComponent<ComponentText2D>();
+			// sceneInteractableComponents.push_back(gameObject->CreateComponent<ComponentText2D>());
```
gerard-queralt, line 373:
```suggestion
```

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentText2D.cpp
```diff
@@ -0,0 +1,53 @@
...
+}
+
+void WindowComponentText2D::DrawWindowContents()
+{
+	DrawEnableAndDeleteComponent();	
+		
+	std::string items;
+	std::string texts;
+	
+	for (std::shared_ptr<ResourceFont> i : componentText2D->getFonts())
```
gerard-queralt, line 27:
```suggestion
	for (const std::shared_ptr<ResourceFont>& resource : componentText2D->getFonts())
```

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentText2D.cpp
```diff
@@ -0,0 +1,53 @@
...
+	{
+		items+=i.get()->GetFileName();
```
gerard-queralt, line 29:
```suggestion
		items += i->GetFileName();
```

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentText2D.cpp
```diff
@@ -0,0 +1,53 @@
...
+		items += '\0';
```
gerard-queralt, line 30: Shouldn't this be a line break?
```suggestion
		items += '\n';
```

Source/DataModels/Windows/SubWindows/ComponentWindows/WindowComponentText2D.h
```diff
@@ -0,0 +1,20 @@
...
+{
+public:
+	WindowComponentText2D(ComponentText2D * component);
+	~WindowComponentText2D() override;
+
+protected:
+	void DrawWindowContents() override;	
+	std::unique_ptr<WindowFontInput> inputFont;
+	ComponentText2D* componentText2D;
+	char str1[128];
```
gerard-queralt, line 18: Use better names, please.

Source/FileSystem/Importers/FontImporter.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+FontImporter::~FontImporter()
+{
+}
+
+void FontImporter::Import(const char* filePath, std::shared_ptr<ResourceFont> resource)
+{
+	/*ENGINE_LOG("Import font from %s", filePath);*/
+	char* buffer{};
```
gerard-queralt, line 19: Never deleted, memory leak.

Source/FileSystem/Importers/FontImporter.cpp
```diff
@@ -0,0 +1,115 @@
...
+	unsigned int size;
+	if (resource.get()->ImportFont(filePath)) {
```
gerard-queralt, line 21:
```suggestion
	if (resource->ImportFont(filePath))
	{		
```

Source/FileSystem/Importers/FontImporter.cpp
```diff
@@ -0,0 +1,115 @@
...
+		
+			unsigned int header[5] =
+			{
+				it.symbol,
+				it.width,
+				it.height,				
+				it.advance_x,
+				it.size
+			};
+		int header2[2] = { it.left, it.top };
```
gerard-queralt, line 90: Better name.

Source/FileSystem/Importers/FontImporter.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+			size += sizeof(header) + sizeof(header2) + sizeof(unsigned char) * it.size;
+	}
+	char* cursor = new char[size];
```
gerard-queralt, line 94: This leaks.

Source/FileSystem/Importers/FontImporter.cpp
```diff
@@ -0,0 +1,115 @@
...
+	
+	fileBuffer = cursor;
+	for (StoreFontInPixels it : resource.get()->storeFont)
```
gerard-queralt, line 97:
```suggestion
	for (const StoreFontInPixels& it : resource->storeFont)
```

Source/Shaders/text_fragment.glsl
gerard-queralt, line 1: Talk to the graphics team to make sure both shaders are needed, maybe we can reuse some.

gerard-queralt (changes requested):
Main things that I think can be improved:
- Coding style: make sure your code follows our programming guidelines. I shared a link in one of my comments on how you can reformat the contents of your files.
- Pass by reference: most of your methods create unnecessary copies, be it in the parameters or in loops. Tried to keep an eye out for them, but please take another look yourself to see that I didn't miss any.
- Unnecessary access to raw pointers: whenever you want to access the memory managed by a unique or shared pointer using the member access operator `->`, you're first accessing the raw pointer by calling `get`. This is not needed, which you can see in the[ documentation](https://en.cppreference.com/w/cpp/memory/shared_ptr/operator*). I didn't point out every single instance of this, please check on your own and correct it.
- Commented-out code: there's a lot of commented out code. Either remove it or add an AXO_TODO so it gets addressed (hopefully).
- Memory leaks: there are a couple of leaks in the importer.

Source/DataModels/Components/UI/ComponentText2D.cpp
```diff
@@ -0,0 +1,267 @@
...
+	meta["color_y"] = static_cast<float>(color.y);
+	meta["color_z"] = static_cast<float>(color.z);	
+}
+
+void ComponentText2D::InternalLoad(const Json& meta)
+{	
+#ifdef ENGINE
+	std::string path = meta["assetPathImage"];
+	bool resourceExists = !path.empty() && App->GetModule<ModuleFileSystem>()->Exists(path.c_str());
+	if (resourceExists)
```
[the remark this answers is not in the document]
elbuho12Jesus: I check that the source exist in the folder Assets, but maybe I should do it another way.

elbuho12Jesus:
I finished with the requirements, could you check if this is ready to use?

PR #765, opened 2023-09-28
fix[Scheduler]: prevent empty tasks from running

gerard-queralt:
Not really sure why this happens or if it's an actual issue (ran into it while attempting to force a crash elsewhere), but more prottection never hurt.


gerard-queralt:
Closing this, will do a new MR also integrating the changes done in #760

PR #766, opened 2023-09-28
feat: proper Schedulable class

gerard-queralt:
This cherrypicks the changes in #760 and merges the fix from #765. Felt like the Scheduler was carrying too much responsability, so moved most of it to a proper Schedulable class instead of having a `using` declaration.

### DIFF
--- a/Source/Application.cpp
+++ b/Source/Application.cpp
@@ -189,11 +189,11 @@ void Application::OnPause()
 {
 }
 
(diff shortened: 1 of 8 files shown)


Dedado3000 (dismissed):
I Don't See nothing bad in code and i loaded some times (more than spected to crash) and 0 crashes

Dedado3000 (approved):

markmeliaa (approved):

MarcGonzalezMoratona (approved):
Code looks great

PR #767, opened 2023-09-30
fix[ModuleScene]: schedule script calls to load scene

gerard-queralt:
Hidden in a private function so users can't request their scene load to run in-place, which was the cause of the error and would never work.


Dedado3000 (commented):
I will test it when i get home

uuv3duh85i741

markmeliaa (approved):

Dedado3000 (approved):

PR #768, opened 2023-09-30
fix: doublejump afterattack animation

MarcGonzalezMoratona:
This fixes state machines of bix and allura for jump after attack and double jump after attack animations, as well as the attack+jump+jumpattack and attack+doublejump+jumpattack combo animations

To be merged after #758


markmeliaa (commented):
Idk if this is related, but with Allura, if I do a single jump and then jump attack, the animation of the jump attack is correctly displayed, but if I do a double jump and then a jump attack, the animation of the jump attack is not displayed at all (just with Allura, Bix does not display this behaviour, their jump attack animations are shown correctly at all times).

MarcGonzalezMoratona:
> Idk if this is related, but with Allura, if I do a single jump and then jump attack, the animation of the jump attack is correctly displayed, but if I do a double jump and then a jump attack, the animation of the jump attack is not displayed at all (just with Allura, Bix does not display this behaviour, their jump attack animations are shown correctly at all times).

I'll take a look and check if the same thing happens in develop. Thank you Mark!

markmeliaa:
> I'll take a look and check if the same thing happens in develop. Thank you Mark!

Thanks! Even if it does happen, do you think you can fix it? As it is an animation fix it kind of gets into the context of this pr.

MarcGonzalezMoratona:
> > I'll take a look and check if the same thing happens in develop. Thank you Mark!
> 
> Thanks! Even if it does happen, do you think you can fix it? As it is an animation fix it kind of gets into the context of this pr.

Yes, it happens in develop, so I'll try to fix it in this PR and it will be solved everywhere

MarcGonzalezMoratona:
@markmeliaa state machine updated!

markmeliaa (approved):
Works fine now, thanks!

Piterga12 (approved):

PaablooCH (approved):

PR #769, opened 2023-09-30
fix: Allura jump attack & other small changes

markmeliaa:
Mainly, this pr's objective was to add the correct logic to Allura's jump attack, but in the process a few more things got fixed:
- The grenade of the jump attack is thrown with more force to the ground.
- The StateMachine Window is not opened each time the stop button is pressed.
- Vfx for Allura's grenade being thrown to the ground.
- Fix the way to differentiate if a jump attack is finisher or not.
- Stop Bix's and Allura's movement during jump attack.
- Add a function to filter the list of enemies by distance, to limit Allura's jump attack range.
- Deal damage at the right time to the enemies with Allura jump attack. (still works, but somehow the jump attack does not trigger the mesh effect for the `MiniBoss2` in Allura's trial scene, although it takes damage correctly. Any other enemy in that scene works fine).
- Add nullity checks to the effects of the jump attack, so if any of them are not set, the script can continue working.
- Allura does not get stuck in the attack animations if attacking while walking.
- Removed the need to create a `forceArea` for each one of Allura's jump attacks, as the bullet now falls straight to the floor.
- A small jump was added when Allura throws the grenade to match better its animation.
- Updated scenes accordingly with the new functionality.
- You can't switch characters while an attack is taking place (thanks David 😄).

You can test this new functionality in either of the levels or in the trial scene for Allura.


Game/Scripts/JumpFinisherArea.cpp
```diff
@@ -47,62 +47,55 @@ void JumpFinisherArea::Update(float deltaTime)
...
+	}
+
+	if (actualLandingParticleSystem)
 	{
 		App->GetModule<ModuleScene>()->GetLoadedScene()->DestroyGameObject(actualLandingParticleSystem);
 	}
 
-	actualLandingParticleSystem = loadScene->DuplicateGameObject(landingParticleSystemPrefab->GetName(), landingParticleSystemPrefab, loadScene->GetRoot());
-	actualLandingParticleSystem->GetComponent<ComponentParticleSystem>()->Enable();
+	if (landingParticleSystemPrefab)
```
gerard-queralt, line 78: As a reminder, you can use `Script::Asert` to check for things like these beforehand and avoid nullity checks. Of course it depends on the logic of the script, if it's acceptable that such objects are null then ignore me.

Game/Scripts/PlayerAttackScript.cpp
```diff
@@ -403,50 +403,70 @@ void PlayerAttackScript::InitJumpAttack()
...
 	jumpFinisherScript->VisualLandingEffect();
-	if (enemyDetection->AreAnyEnemiesInTheArea())
+
+	std::vector<ComponentRigidBody*> enemiesToHit = enemyDetection->GetEnemiesInTheArea();
+	if (!isMelee)
+	{
+		enemiesToHit.clear();
+		enemiesToHit.reserve(enemyDetection->GetEnemiesInTheArea().size());
+		enemyDetection->FilterEnemiesByDistance(6.5f, enemiesToHit); // 6.5f like the size of Bix jump attack
+	}
```
gerard-queralt, line 446: Minor thing, but we could save on a function call (not sure how expensive `GetEnemiesInTheArea` is so not sure of the impact either).
```suggestion
	std::vector<ComponentRigidBody*> enemiesToHit;
	if (isMelee)
	{
			enemiesToHit = enemyDetection->GetEnemiesInTheArea();
	}
	else
	{
		enemiesToHit.reserve(enemyDetection->GetEnemiesInTheArea().size());
		enemyDetection->FilterEnemiesByDistance(6.5f, enemiesToHit); // 6.5f like the size of Bix jump attack
	}
```

Game/Scripts/PlayerAttackScript.cpp
```diff
@@ -403,50 +403,70 @@ void PlayerAttackScript::InitJumpAttack()
...
 		comboSystem->SuccessfulAttack(comboCountJump, currentAttack);
 	}
 }
 
 void PlayerAttackScript::EndJumpFinisherAttack()
 {
 	jumpFinisherScript->VisualLandingEffect();
-	if (enemyDetection->AreAnyEnemiesInTheArea())
+
+	std::vector<ComponentRigidBody*> enemiesToHit = enemyDetection->GetEnemiesInTheArea();
```
gerard-queralt, line 459: Same thing as my previous comment.

Game/Scripts/PlayerAttackScript.cpp
```diff
@@ -566,7 +586,7 @@ void PlayerAttackScript::ResetAttackAnimations(float deltaTime)
 			{
 				actualName = "JumpAttackRecovery";
 			}
-			if (animation->GetController()->GetStateName() != currentAttackAnimation)
+			if (animation->GetController()->GetStateName() != actualName)
```
gerard-queralt, line 589: Hate to be that guy, but you probably meant `currentName` :P

gerard-queralt (commented):
Code looks great, just a few really, really minor things, feel free to ignore them.

Game/Scripts/JumpFinisherArea.cpp
markmeliaa: Yea, here I didn't use the assert because, for example, Allura has no landingParticleSystem for the jump finisher, so it threw an error when the jump finisher was triggered for Allura, as only Bix uses it.

Game/Scripts/PlayerAttackScript.cpp
markmeliaa: Well seen!

Game/Scripts/PlayerAttackScript.cpp
markmeliaa: Fair, although I didn't create that variable, it just happened that I had to use it, but I'll change it!

gerard-queralt: I imagine, appreciate you changing it :D

DADSILVER (approved):
Good Job. All the bugs seem fixed, and Allura's new jump attack is astonishing.

OriolSors (approved):
All is working properly!! Good Job!

PR #771, opened 2023-10-03
Camera boss

Dedado3000:
New Boss Level with Art.
- Camera boss that follows you and camera


diegusmejia117 (approved):
Very good, very bueno

ericdacal (approved):

PR #789, opened 2023-10-07
fix[SceneLoad]: fixed memory leak in load scene async

MacXxs:
Loading a scene in a separated thread caused memory leaks in textures, 'cause textures can only be created, deleted and made resident and non resident in the same thread of the opengl context.

As Carlos already said: https://discord.com/channels/1029122111832338563/1030064402684194897/1130104514372960296

**This  was the cause for black textures in the scene, due to running out of vram so it coulnd't load any more textures**


PaablooCH (approved):
👍

gerard-queralt (approved):

PR #798, opened 2023-10-09
fix: Tutorial crashing

Dedado3000:


Vlukkee (approved):

Piterga12 (approved):

gerard-queralt (approved):
reminder that you can use Assert to avoid these checks :P

PR #807, opened 2023-10-10
feat: Spaceship Escape

UriMurcia:
To test this pr you can go the _SpaceshipLevel.axolotl_ scene. This should be the final level, so please, all the feedback is welcomed.


gerard-queralt (approved):

PaablooCH (dismissed):
hi, the level is awesome but it feels like you loose to much hp per hit. Besides that, I found a few more things:

- After this point the camera starts to do somethings that is not normal like some kind of shaking.
image
 
- this particles stop too suddenly, it not feels good.
image

- try to make the bounding box of the particles bigger, because they dissapear suddenly when they are near the ship.
- I would put some random value to determine when the flames start, because those one always are off when you arrive.
image

- when you die there is not any death scene or something similar.
- when I finished the level, it crashes 
image

UriMurcia:
> hi, the level is awesome but it feels like you loose to much hp per hit. Besides that, I found a few more things:
> 
> * After this point the camera starts to do somethings that is not normal like some kind of shaking.
>   image
> * this particles stop to suddenly, it not feels good.
>   image
> * try to make the bounding box of the particles bigger, because they dissapear suddenly when they are near the ship.
> * I would put some random value to determine when the flames start, because those one always are off when you arrive.
>   image
> * when you die there is not any death scene or something similar.
> * when I finished the level, it crashes
>   image

UriMurcia:
> hi, the level is awesome but it feels like you loose to much hp per hit. Besides that, I found a few more things:
> 
> * After this point the camera starts to do somethings that is not normal like some kind of shaking.
>   image
> * this particles stop to suddenly, it not feels good.
>   image
> * try to make the bounding box of the particles bigger, because they dissapear suddenly when they are near the ship.
> * I would put some random value to determine when the flames start, because those one always are off when you arrive.
>   image
> * when you die there is not any death scene or something similar.
> * when I finished the level, it crashes
>   image

Thanks for taking a look Pablo! I'll start working on these issues asap!

Vlukkee (approved):

AuxPort66 (approved):

PR #822, opened 2023-10-13
fix: proper use of mutexes + scheduler fix

gerard-queralt:
Couple of things fixed in this MR:
- Mutexes are now properly locked. Didn't realize that I forgot to give a name to a lock a few months back, and have been copy-pasting that mistake all over the place. Fixed that and the deadlocks that came up because of it.
- Scheduler::RunTasks stores a snapshot of the current state of the Schedulable queue. We retrieve the front of the queue as a reference `Schedulable& schedulable = ...`, but if the queue reallocates somewhere else (because ScheduleTask was called, for example), that memory can become invalid, which I'm almost positive is the cause of the original crash reported.

### DIFF
--- a/Source/Auxiliar/GameBuilder.cpp
+++ b/Source/Auxiliar/GameBuilder.cpp
@@ -103,7 +103,7 @@ void OnFileZipped(const FileZippedData& data)
 					maxMinutesToZip,
 					data.timeTaken.count());
 	}
-	std::scoped_lock(fileZippedDataMutex);
+	std::scoped_lock lock(fileZippedDataMutex);
 	lastFileZippedData = data;
 }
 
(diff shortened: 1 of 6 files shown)


markmeliaa:
Is there any way to check if this is correctly fixed rather than randomly trying for a crash?

markmeliaa (approved):
Code looks great, good use of locks and mutexes, I like it! Learned about `scoped_lock` and `recursive_mutex`, gj!

gerard-queralt:
> Is there any way to check if this is correctly fixed rather than randomly trying for a crash?

Not that I can think of, sorry. In the branch I had for testing I hacked in an infinite loading loop, we could try that, but in my testing it sometimes happened in the first load and sometimes in the sixtieth, so it's not a guarantee either :/

gerard-queralt:
> I trust that Gerard tested the feature to make sure that the engine still works after this change

And if your trust is misplaced I'll just fix it later, as always 😜

PR #825, opened 2023-10-13
feat: Force improvements

MarcGonzalezMoratona:
Starting from @xavi1111 's PR, I made some changes to the force:

- Force is now working in both boxes of level 1
- The script has been refactored and now is similar to the hacking script
- Bix does not slide on the second box (vertical one)
- Player is paused during force
- Bix rotates towards the box position
- The script logic has been simplified to make the force work in an axis at a time
- The box can only move from 3 to 15 units far from the player (3 is the minimum because less can make the box hit the player)
- Now the force works with gamepad
- Force animation is not triggered if the raycast does not hit a forceable box
- Now you don't have to keep B button pressed to do the force, it enables and disables as the hacking does
- PlayerRotationScript has been removed from the codebase (it was only used to rotate Bix during force)
- This PR removes the PlayerRotationScript from Bix and Allura prefabs. 
- Force UI particle added to the first forceable box of level 1. When the player hits the forceable box using button B, the particle stops. (It will not play anymore, this is intended but can be changed if it makes more sense to you)
- The hacking particle (reused for the force) plays on start. (Level 2 keeps working the same way)

I could not add the force particle to the second forceable box because it disappears when the camera approaches it. (Can we suppose that players will understand the feature just with the first forceable box?)
Let me know if you have any idea about why this happens. If you manage to add this particle to the second box, it's a nice to have!


Dedado3000 (changes requested):
There are some things that are faling:
- I can't push the object with the mouse using the scroll
- I can't Test the up and down of the box beause it crashes
- There is a jiggle effect that you can see in the video



https://github.com/Horizons-Games/Axolotl-Engine/assets/71272598/c2b062f3-5a6f-43ec-b919-2e99cd56af71

MarcGonzalezMoratona:
> * up and down of the box beause it crashes

I didn't expect to have this working with the mouse, but I can fix it easily for both mouse wheel and movement. It should work without giggle with the gamepad, but I'll check what happens with the mouse.
The second box does not have a particle system because the particle disappears if I add it, so I'll add a check to avoid the crash if there's no particle system.

Do you expect the first box to move in two axes?

MarcGonzalezMoratona:
I solved requested changes. Let me know if there's something missing yet. When I get the approval, I'll update the scene and solve conflicts, and ask for approval again

Dedado3000 (approved):

Piterga12 (commented):
Everything perfect, bu t i would add a limit to the first box so that we cannot place it on the toxic water (as the toxic water is lower and makes it look like is floating

image
image

Piterga12 (approved):

xavi1111 (approved):
Tested the branch force seems to work nicely now. The refactor of the force script was key to avoid problems found on the previous branch great job

MarcGonzalezMoratona:
> Everything perfect, bu t i would add a limit to the first box so that we cannot place it on the toxic water (as the toxic water is lower and makes it look like is floating
> 
> image image

done! let me know if it fixed

Piterga12 (dismissed):
Works great!

Dedado3000 (approved):

MarcGonzalezMoratona:
There's a particle missing for UI in the second box. Please wait until we add this to merge the PR!

Vlukkee:
works great! Change the influence zone and when show te particle and will be amaizing

Dedado3000 (approved):

AuxPort66 (approved):

PR #826, opened 2023-10-14
fix: gameplay bix & allura death

DADSILVER:
Now the jump attack works.


AuxPort66:
The UI of life still have problems, it's not completely emptied. Furthermore, I died in miniboss 1 and my character, despite not being able to move or attack, kept sliding, and then like 30 seconds after applies the death animation. We probably should apply death animation in the first instance we recieve damage that kill us.

DADSILVER:
> The UI of life still have problems, it's not completely emptied. Furthermore, I died in miniboss 1 and my character, despite not being able to move or attack, kept sliding in idle, and then like 30 seconds after applies the death animation.

I solved UI problems. Now, when any character dies his health bar is put to 0. But I need more info about the error that you mention. Which character stays idle and how die (changing the health system script or by taking dmg?)

markmeliaa (approved):
Code looks great, jump attack now works correctly. I also tried dying a couple times while attacking and it worked fine.

DADSILVER:
> The UI of life still have problems, it's not completely emptied. Furthermore, I died in miniboss 1 and my character, despite not being able to move or attack, kept sliding in idle, and then like 30 seconds after applies the death animation.

Fixed. It was a legacy error related to:

			currentHealth -= actualDamage;

			ModuleInput* input = App->GetModule<ModuleInput>();
			input->Rumble();

			if (currentHealth - damage <= 0)

As you can see currentHealth was updated at the beginning of TakeDamage and after this in the conditional try again to subtract the damage and this conditional set death state.

AuxPort66 (approved):
I will modify some aspects of the UI GameOver on other PR. The bug is solved so I close this

PR #828, opened 2023-10-15
Fix: Now the spawner is able to reactivate all enemies once the pool of enemies is empty

totisama:
After the second activation of the shield, the pool of enemies is empty, so now the spawner is able to reactivate the enemies and add them to the pool.


Game/Scripts/BossShieldEnemiesSpawner.cpp
```diff
@@ -60,7 +60,7 @@ void BossShieldEnemiesSpawner::Update(float deltaTime)
 
 void BossShieldEnemiesSpawner::StartSpawner()
 {
-	if (enemiesReadyToSpawn.size() <= 0)
+	if (enemiesReadyToSpawn.size() < 4)
```
markmeliaa, line 63: Do you mind declaring this variable as a `constexpr` in the header file and set its value to 4 rather than just writting here a 4? It would improve readability and make this look less "hard coded".

markmeliaa (approved):

diegusmejia117:
Just some bugs i noticed:
-right now some enemies get stuck on the platform, this happens i noticed when they are hitted by the sliding fence, when they are in this state they are inmortal and doesnt move but they activate the receive damage animation.

image

image

- when the elevator end its second cycle it drops suddenly to its downward position, i dont know if this is an intended behaviour o a bug but pls check it as it is very noticible.

diegusmejia117:
works as intended, good job!, just coordinate with @Dedado3000 to see why the camera is not pointing up at the boss when she is the balcony.

diegusmejia117 (approved):

markmeliaa (approved):
GJ Rodrigo!

PR #829, opened 2023-10-15
fix: spawn enemies

DADSILVER:
Please test 2 times in the scene _LEVEL02_ALFA. Now only appear venomites


Dedado3000 (dismissed):
I liked the experiences with only venomites, i feel a bit hard but works perfectly

diegusmejia117:
Just some feedback:
-when the venomites take the elevator down they run to the fringe of the arena, and stay there:
1- Please move the final destination of the venomites to the center of the Hand Logo.
2- Please test why the venomites just stay there and dont follow you.

markmeliaa (approved):

diegusmejia117 (approved):

MarcGonzalezMoratona:
Everything looks great, but I would make the elevators move a bit slower because the venomites appear super fast! I don't know if this can be addressed in another PR. Good job David!

MarcGonzalezMoratona (approved):

PR #830, opened 2023-10-15
Fix: gameOver waits until health bar do the animation

AuxPort66:
GameOver screen now only appears when the UI health bar ends the full animation or certain time pass. With this you can proper see the death of the character.

The flow should be Bix death animation -> HealthBar goes down -> In case the HealthBar goes down is too short we have a auxiliar timer to wait a minimum time -> GameOver screen.


markmeliaa (dismissed):
GJ, code looks great and now the death->Game Over flow feels better.

DADSILVER (approved):

Vlukkee (approved):

diegusmejia117 (approved):

PR #834, opened 2023-10-17
Fix: Switch Characters Time and Fixed Particles

Piterga12:
Improved the times for the switch, so it's faster and the player doesn't jump too much.
Also fixed that the particles had the follow transform on (Miki fixed this for the release, but wasn't uploaded to GitHub).

Please let me know if the switch feels better like this.


DADSILVER (dismissed):

markmeliaa (dismissed):

MarcGonzalezMoratona (dismissed):

AuxPort66:
The second character appears too low, she should come out higher so the particle have the effect of the change

AuxPort66 (approved):

PR #839, opened 2023-10-17
feat: improve boss charge attack and add new rocks for it

markmeliaa:
The main goal of this PR was to add the correct fbx of the rocks for the final boss battle and also avoid that the boss gets stuck behind her own rocks.

This were the tasks developed:
- Add Rock FBX and include it correctly in the charge attack
- Improve check when rocks collide
- Don't show spawning rocks before correctly set in their position
- Avoid rocks from spawning near the edges
- Randomize rock Y rotation when spawning
- Add comments for where the vfx should go
- Make the charge attacks EVEN less likely of happening (if you feel they still happen a lot, lmk)

Bugs that idk how to solve:
- The shield sends the player flying if you are near it when it appears.
- If a rock falls over an enemy, its collision and damage are calculated way too early (this is a rigidbody/physics problem, but I don't see it as too serious because it does not happen a lot).


diegusmejia117:
Just tested it, great changes!, now i only saw 2 things:

-the boss stills does a lot of dashing(less now but still)
-sometimes the rocks block the boss movemente too much
-when the boss jumps into the balcony, she now stays idle instead of doing the missible attack animation

diegusmejia117 (approved):

AuxPort66 (approved):

PR #843, opened 2023-10-18
Fix: Changes in obstacles logic

totisama:
Now the Obstacle should be working correctly, so the boss shouldnt get stuck with them


markmeliaa (approved):
Ty Rodrigo!

diegusmejia117 (approved):

PR #849, opened 2023-10-20
feat: Missions on Level 1

Piterga12:
The missions are on the level 1, the mission stays visible for 30secs (this can be easily changed on the inspector) or when we enter on the trigger of the next mission (then the new mission appears and the old one leaves


Dedado3000 (approved):
Nice Work

Vlukkee (changes requested):
Really nice job, need to made some changes

No timer in missions
Add 2 new mission hangar and cascade place.

Vlukkee (approved):
Amazing job!

markmeliaa (approved):
Great job Peter, congrats!

PR #852, opened 2023-10-20
Rework In Combat System

Dedado3000:
Reworked the system that makes you detect if you are in combat, this fix some interactions and add new things:
- now in miniboss level2 only have the bar when you go down the arena
- Now is easy to acces to combat system


Vlukkee (dismissed):
Good, pretty and cheap!

Vlukkee (approved):
May the latin power with you!

Piterga12 (approved):

PR #853, opened 2023-10-20
fix: drone and venomite death

DADSILVER:
- Now drones follow you until they explode. then they stop.
- Now venimotes don't make a weird animation after dead seeking a target.
- Now venimotes stop after dead instead of following you.
- shieldboss only picks enemies disabled instead of enemies dead. Sometimes drones disappear before explode.
- If is inmortal is activated you cannot die.


diegusmejia117 (approved):
Good job, works great, just one observation, i don't know if its something related to this pr in specific, but the second miniboss seems to have crazy spams from time to time(seems it happens after the laser).

Vlukkee:
> Good job, works great, just one observation, i don't know if its something related to this pr in specific, but the second miniboss seems to have crazy spams from time to time(seems it happens after the laser).

Its in develop too so its not this pr

Vlukkee (approved):

PR #854, opened 2023-10-21
refactor: level 2 spawn from enemies

DADSILVER:
- Now enemies have a path to the navmesh and then they patrol.
- Fix the error with the doors. When enemies try to pass the door is closed because the player is in combat mode.
- Now when an enemy is respawned we force stop patrolling in case that the enemy finishes in this state.

### DIFF
--- a/Game/Scripts/ActivationLogic.cpp
+++ b/Game/Scripts/ActivationLogic.cpp
@@ -27,7 +27,8 @@
 REGISTERCLASS(ActivationLogic);
 
 ActivationLogic::ActivationLogic() : Script(),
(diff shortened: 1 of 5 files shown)


Vlukkee:
When you are near the elevators the minions crash with the elevator rigid.
If this can be improve will be great . I dont know if its easy or not.

https://github.com/Horizons-Games/Axolotl-Engine/assets/25518155/831a99d6-2bad-48a2-a255-85f0c419564b

Vlukkee (approved):

PaablooCH (approved):

Dedado3000 (approved):

PR #857, opened 2023-10-22
Feat: Input changed

AuxPort66:
- Different bugs on combo system UI (The fade effect was not initialized correctly, etc.)
- Added Special buttons to the UI and special indicators
- New UI effects
- Special now activates automatically
- Changed PowerUp to B and Dash to RT.
- Added RB and LB as specials.
- Change all the compareTags("Enemy") so we can use the priority Tag on the boss and mini boss

You can test the changes in lvl1


Game/Scripts/CameraControllerScript.cpp
```diff
@@ -12,6 +12,8 @@
 
 #include "Camera/CameraGameObject.h"
 
+#include "ComboManager.h"
```
markmeliaa, line 15: Include not necessary?

Game/Scripts/UIComboManager.cpp
```diff
@@ -38,34 +40,57 @@ void UIComboManager::Init()
...
 		transparency = noFillBar->GetComponent<ComponentImage>()->GetColor().w;
 	}
+
+	GameObject* specialBoxGO = owner->GetChildren()[2];
+	specialBoxGO->Disable();
+	specialBox.text = specialBoxGO->GetChildren()[1];
+	GameObject* buttonsGO = specialBoxGO->GetChildren()[2];
+	specialBox.lb = buttonsGO->GetChildren()[0];
+	specialBox.separator = buttonsGO->GetChildren()[1];
+	specialBox.rb = buttonsGO->GetChildren()[2];
```
markmeliaa, line 50: Instead of hardcoding these values based on their position as children, wouldn't it be better to make them `SERIALIZE_FIELD`s and set them through the UI?

markmeliaa (approved):
Functionality looks great, just a couple comments regarding the code.

DADSILVER (approved):

PR #864, opened 2023-10-24
feat : Boss rock breakable

AlexisDieu:
- Adding component Breakable to the Rock that is falling
- It breaks when hitting the player or the boss and also after a time on the floor
- You can modify the time before it breaks when it's on the ground
- You can modify the time before the pieces disapear when hitting something

Todo : adding a Fade out on all pieces to make it not just disapear instantly


Vlukkee (approved):
¬¬

AuxPort66 (approved):

AuxPort66:
We will change the effect when the rock is on the floor because we want something more chill, but we mantain the breakable when it hits the player/enemy (and we will use the fade out for both effects)

PR #865, opened 2023-10-24
fix: Miniboss2 move with shield and slide over the floor seeking

DADSILVER:
- Now miniboss doesn't move if he enters in shield seeking.
- Now miniboss doesn't slide if he is seeking but in range attack.


Vlukkee (approved):
Nice job! The only thing is the venomites push the mini boss  when they crrash and move him.
I approve it because it something not easy to do.

PaablooCH (approved):

PR #870, opened 2023-10-25
fix&feat: improved boss fight

markmeliaa:
In this pr, numerous fixes were applied to the boss fight in order to improve it:
- Boss dies correctly.
- When a rock falls over a patrol point, that patrol point is set as UNAVAILABLE.
- Changed landing position after the missiles attack.
- Balanced the cooldown for the attacks (yes, the charge included, this should make that, once for all, the charge is not triggered everytime).
- Tally correctly the missiles animations to the triggering and ending of the missiles attack.
- Combo bar is now visible during the boss fight (thanks David).
- Several animation adjustments between attacks.
- Swapping between chrarcters in the boss fight no longer causes a crash.
----
Another changes performed in this pr:
- Removed unused scenes.
- Renamed main scenes.
- Set the grid always to true by default and show rigidbodies always to false by default.
- ImGuizmo is not drawn during play mode.
- Added the fbx of the missile to the missiles attack of the boss and set it working (with vfx it will look much better).
----
Known bugs (Rodrigo is working on them):
- The boss sometimes walks backwards.
- The boss keeps walking when doing the take damage animation (ideally, he should stay in place and then continue walking).


UriMurcia:
At some point the game crashes in my pc, I don't know why because it doesn't give any error and can't do anything. Screen ferezes but music continues. I think it happened after the rock falling on the ground :(

UriMurcia:
Other things I've seen that I think are not the objective of this PR but could be cool to fix:
- The boss still uses the movement animation when is stopped at a point patrol. To fix this you just have to add this parameter at the state machine:
image

- The boss, when doing the charge attack, sometimes doesn't hit the player even if the player is not moving.
- The boss is patrolling a lot, if the player doesn't attack the boss, it can be patrolling without attacking for a long time.
- The walking animation of the boss doesn't match the movement. We could both turn down the animation speed and turn up the velocity of the boss in the AIMovement script.

markmeliaa:
> At some point the game crashes in my pc, I don't know why because it doesn't give any error and can't do anything. Screen ferezes but music continues. I think it happened after the rock falling on the ground :(

It happened to me too! I assumed it was only a me problem, seems it is something more serious... It's as you said, no crash or engine error, the screen just freezes and there is not way to interact with it.

markmeliaa:
> * The boss, when doing the charge attack, sometimes doesn't hit the player even if the player is not moving.

I tried fixing everything else, but I don't know what do you mean with this one.

P.S. Check if the walking animation is now correctly adjusted, if not, let me know at which number you think it would be best to set it.

Vlukkee (approved):

Piterga12 (approved):

PR #872, opened 2023-10-27
Fix: Stuck in Dash and repeating DoubleJumps anims

Piterga12:
This fixes two bugs where the player would perform twice the double jump animation and a bug where the player would get stuck in the dashing animation if you dash right when falling from a very short fall (like in the image)
image


markmeliaa (approved):
Great job Peter, thanks for fixing this!

AuxPort66 (approved):

PR #878, opened 2023-10-29
feat: add planar reflection; fix localIBL

PaablooCH:
I added a number of planar reflections on the level 2, these reflections are for the enemies and heroes and also the blocked door.
Aditionallly, the locals inside the armory are fixed and now display correctly the relfections.


markmeliaa (approved):
Great job Pablo, looks better now!

Vlukkee (approved):

PR #881, opened 2023-10-31
fix[Combat Tutorial]

diegusmejia117:
Added some code to take care of the potential death of the dummy before doing the required attacks for the tutorial.


Piterga12 (changes requested):
Two things only, i have to press twice the d-pad button on the first text (one time paralyzes the player and the second time actually changes to the next text, and second, on the jump attack part, i can finish the tutorial out of the tutorial zone

https://github.com/Horizons-Games/Axolotl-Engine/assets/80328187/001d271e-d94d-4a94-aef2-6e0472675e4b

Piterga12 (approved):
Thanks for the work Diego!

AuxPort66 (approved):

PR #882, opened 2023-11-01
fix&refactor: powerup tutorial

diegusmejia117:
Added new state so the player can take his time learning about power ups, also refactored the update to use switch instead of ifs.

### DIFF
--- a/Game/Scripts/CombatTutorial.cpp
+++ b/Game/Scripts/CombatTutorial.cpp
@@ -36,7 +36,7 @@
 REGISTERCLASS(CombatTutorial);
 
 CombatTutorial::CombatTutorial() : Script(), combatDummy(nullptr), userControllable(false), combatTutorialUI(nullptr),
-debugPowerUp(nullptr), finalWaitTime(2.0f), finalTotalWaitTime(5.0f), tutorialActivable(false), nextStateActive(true),
+debugPowerUp(nullptr), finalWaitTime(1.0f), finalTotalWaitTime(5.0f), tutorialActivable(false), nextStateActive(true),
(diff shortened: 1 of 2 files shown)


Game/Scripts/CombatTutorial.cpp
markmeliaa, line 1: Have you considered, instead of throwing all the functionality in each case of the `switch`, create a function with a proper name to better understan what each case does?

markmeliaa (approved):

Game/Scripts/CombatTutorial.cpp
```diff
@@ -84,22 +84,27 @@ void CombatTutorial::Update(float deltaTime)
 		doorRigidbody->SetIsTrigger(false);
 		componentMoveScript->SetIsParalyzed(true);
 
-		if (tutorialUI->GetTutorialCurrentState() == 1)
+		switch (tutorialUI->GetTutorialCurrentState())
```
MarcGonzalezMoratona, line 87: if you want to continue using the switch approach, I think that you can use only one. You probably don't need two switch with the same condition

MarcGonzalezMoratona: If you join both switch statements, put the cases in order 😄

MarcGonzalezMoratona:
When I finish the tutorial now the door does not open
image

MarcGonzalezMoratona (changes requested):
The functionality is not working as expected

MarcGonzalezMoratona (approved):
Everything is working now, the code looks much better. Good job!

Piterga12 (approved):
Works great!

PR #883, opened 2023-11-02
fix: Boss charge attack

OriolSors:
- Boss charge attack failed when Player was not moving. Inaccurate target point fixed.
- Avoided camera to rotate around X axis when Boss gets closer to Player.


markmeliaa (approved):
Looking great now, many thanks Oriol!

PaablooCH (approved):
looks fine 👍

PR #884, opened 2023-11-02
feat: allow different loading screens

gerard-queralt:
Also fixed warnings, again.


markmeliaa (commented):
Code looks great and it loads correctly, do I have to do something special to see different loading screens? Or they aren't set?

gerard-queralt:
> Code looks great and it loads correctly, do I have to do something special to see different loading screens? Or they aren't set?

Nothing is implemented yet, my original MR got a regression so there are no loading screens yet. This is just to accomodate the possibility, since Sandra designed different loading screens for each level and the intention is to use them.

markmeliaa (approved):

Vlukkee (approved):

PR #885, opened 2023-11-05
fix: more boss issues

markmeliaa:
(I hope this is the last pr fixing things related to the boss)

This pr fixes:
- The boss was not displaying the death animation when dying.
- The boss was taking damage randomly when doing the charge attack or the missiles attack.
- Adapt a bit the cooldowns so the boss does not patrol too much.
- The seeking shockwave wasn't being triggered as expected.


PaablooCH (approved):
Everything is fixed correctly. What I think could be better is this time during the battle, when the boss is patrolling, because she is too far from the player after doing the charge attack, she is there in one coner and I'm not even in the other half. It takes me out of the battle. I don't know if we can delete that state or reduece the times she is patrolling, maybe making bigger her detection.

https://github.com/Horizons-Games/Axolotl-Engine/assets/62140600/83190ffd-23cc-4cc6-8f32-fd4361a2f887

markmeliaa:
> Everything is fixed correctly. What I think could be better is this time during the battle, when the boss is patrolling, because she is too far from the player after doing the charge attack, she is there in one coner and I'm not even in the other half. It takes me out of the battle. I don't know if we can delete that state or reduece the times she is patrolling, maybe making bigger her detection.
> 
> 
> 
> https://github.com/Horizons-Games/Axolotl-Engine/assets/62140600/83190ffd-23cc-4cc6-8f32-fd4361a2f887
> 
> 
> 
> 

Sadly, now is too late to implement those changes as they would change too much the boss battle and we are closing in to the final deadline, but the feedback is much appreciated!

PR #886, opened 2023-11-05
Refactor: create bullet loader and add fixes power ups

DADSILVER:
You can try it on all levels.

- Now, the power-ups are appearing with the new model.
- It has the possibility of appearing 1 out of 5 times when killing an enemy.
- Allura has the correct VFX.
- Loaders have been created for enemies and Allura.


Vlukkee:
Nice job! Now I can get power-ups! 
A few things:
- The size of the power-ups can you put it half of the size? Diego and I think it is better. I believe it's 0.100
- Now in the tutorial LVL 01 the dummy doesn't drop a power-up, maybe you changed something?
- Diego tell me if you can change the Allura bullet to a green one because in level 02 the orange one is not seeing well. 
- The level 3 you change it? I only see 1 and 2.

DADSILVER:
> Nice job! Now I can get power-ups! A few things:
> 
> * The size of the power-ups can you put it half of the size? Diego and I think it is better. I believe it's 0.100
> * Now in the tutorial LVL 01 the dummy doesn't drop a power-up, maybe you changed something?
> * Diego tell me if you can change the Allura bullet to a green one because in level 02 the orange one is not seeing well.
> * The level 3 you change it? I only see 1 and 2.

All the things are fixed. Level Final boss official is changed—the others are not.

markmeliaa (approved):
A really good job David, congrats!

MarcGonzalezMoratona:
The code looks great, no issues found. Good job!

PaablooCH (commented):
Is this how it should rotate? Wouldn't it be better for it to turn in on itself? Also It seems that I take the power-up far from it.

https://github.com/Horizons-Games/Axolotl-Engine/assets/62140600/f8742110-d15d-4a10-866a-9c1f6e9174e8

This is not related but I've found a pacifist drone.

https://github.com/Horizons-Games/Axolotl-Engine/assets/62140600/73bb295e-e77e-4d32-89ab-ce58529c0aad

PaablooCH (changes requested):
> Is this how it should rotate? Wouldn't it be better for it to turn in on itself? Also It seems that I take the power-up far from it.
> 
>  Video.sin.titulo.Hecho.con.Clipchamp.1.mp4 

Okay, I have replayed the level and it seems to work as expected. It rotates around itself, so I will mark this as a bug.

PaablooCH (approved):

PR #887, opened 2023-11-07
chore: delete files that we no longer use

PaablooCH:
delete files that we no longer use or added by mistake


AuxPort66 (commented):
The Sewer Cube is used in Level 1 and the Sunset SkyBox Cube is used in any new scene. Please don't delete the metas. It will regenerate it anyways and in the case of Sewer there will be problems with the UID of the scene.

The rest seems fine. Thank you.

DADSILVER (approved):

markmeliaa (approved):

PR #888, opened 2023-11-07
fix: increased planar performance

PaablooCH:
Fix an issue when loading a scene with planar reflections.
Level 2  have more stable fps.
Removed most of the planar.


DADSILVER (approved):
Thx a lot Pablo you save the enemies' spawn.

markmeliaa (approved):
👏👏👏

PR #889, opened 2023-11-08
Feat: Missions wait for the previous mission to disappear

Piterga12:
Now before the new mission appears, it waits for the previous one to disappear from the screen so that they don't overlap eachother.

Also a couple of grammar error on the switch and combat tutorial texts (just a capital I in both)


Game/Scripts/UIImageDisplacementControl.cpp
```diff
@@ -89,6 +89,34 @@ void UIImageDisplacementControl::SetImageToStartPosition()
 	isMoving = false;
 }
 
+bool UIImageDisplacementControl::IsImageInStartPosition() const
+{
+	if (imageTransform->GetPosition().x == startPosition.x &&
```
MarcGonzalezMoratona, line 94:
```suggestion
	return imageTransform->GetPosition().x == startPosition.x && imageTransform->GetPosition().y == startPosition.y && imageTransform->GetPosition().z == startPosition.z;
```

MarcGonzalezMoratona: you can simplify this if-else block in a single return statement

MarcGonzalezMoratona (commented):
The code looks good! Nice job Peter!

MarcGonzalezMoratona (approved):

Vlukkee (approved):

PR #890, opened 2023-11-08
feat: add an icon to the compiled .exe file

gerard-queralt:
Took the chance to also change the name of the project and solution to `AxolotlEngine`. Since the solution has changed it won't show up in your recent visual studio projects, and since the binary has a new name you might get prompted to give it access to your network as if it were a new program.


MarcGonzalezMoratona (approved):
nice one!

markmeliaa (approved):
Looking great!

PR #895, opened 2023-11-13
fix: in Engine code, use GetComponentInternal

gerard-queralt:
Noticed unhanded ComponentNotFoundExceptions.


MarcGonzalezMoratona (approved):
code looks good to me. Haven't tested because I suppose it works the same way as before

Dedado3000 (approved):
Gerard talked wit me about this change and seems correct

Piterga12 (approved):
Godd job noticing the missing GetComponentInternal s

PR #900, opened 2023-11-16
fix: couple more aspects of the boss fight

markmeliaa:
What got fixed this time?

- The boss correctly performs the death animation when dying (I hope this time nobody has any issue with that, as I can't come up with more ways to fix it).
- The boss does not get stuck in the jump for the missiles attack.
- The boss does not escape from the arena after a charge attack.

And also cleared some warnings.


PaablooCH (approved):
death animation ✔️ 
help me s*******er i'm stuck ✔️ 
boss can not escape from us ✔️

gerard-queralt (approved):

